PR #78, opened 2018-09-14
Refactor Sink.Options to options.SinkOptions

relud:


jklukas:
If we do this for SinkOptions, I think we should do so for DecoderOptions as well in this same PR.

I have a slight preference for SinkOptions and DecoderOptions living at the top level rather than inside `options`. I think that preference is mostly about keeping the symmetry that every class under `options` is an enum type used as an option. But this isn't a super strong conviction.

As a bonus for this PR, I just realized that `--help=Options` wasn't totally doing what we wanted, but this PR should enable us to have `--help=SinkOptions` work correctly. We should update README if we make explicit reference to `--help=Options`.

relud:
we do, it's in the deduplicate PR, because DecoderOptions was really small before that PR

PR #84, opened 2018-09-18
Test ingestion-edge with docker-compose

relud:


relud:
good point, I updated the README to use docker-compose

I noticed that in the switch I had left out ROUTE_TABLE, which had disguised the pubsub emulator being unreachable >_<

Also I added an additional docker-compose.override.yml (which is loaded by default) so that devs get a live-updating local server (but it doesn't use gunicorn like prod and ci) and added a default version.json so that tests pass for that mode.

PR #85, opened 2018-09-18
Pre-load all schemas in ingestion-beam

### LINKED ISSUE #57: Pre-load all schemas in ingestion-beam
walk `resources/schemas` and add each schema file found

jklukas:
Closes #57

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ValidateSchema.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ValidateSchema.java
@@ -4,6 +4,7 @@
 
 package com.mozilla.telemetry.decoder;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.io.Resources;
 import com.mozilla.telemetry.transforms.MapElementsWithErrors;
 import com.mozilla.telemetry.util.Json;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ValidateSchema.java
```diff
@@ -34,7 +66,6 @@ private static void loadSchema(String name) throws SchemaNotFoundException {
       JSONObject rawSchema = Json.readJSONObject(schema);
       schemas.put(name, SchemaLoader.load(rawSchema));
     } catch (IOException e) {
-      // TODO load all existing schemas on startup
```
relud: excellent! we should remove the `if (!schemas.containsKey(name))` section from `getSchema` that calls `loadSchema`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ValidateSchema.java
```diff
@@ -19,13 +22,42 @@
 
 public class ValidateSchema extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
   private static class SchemaNotFoundException extends Exception {
-    SchemaNotFoundException(String msg) {
-      super(msg);
+    SchemaNotFoundException(String name) {
+      super("No schema with name: " + name);
```
jklukas: Refactored SchemaNotFoundException so we don't need to duplicate the message.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ValidateSchema.java
```diff
@@ -34,11 +66,18 @@ private static void loadSchema(String name) throws SchemaNotFoundException {
...
     }
   }
 
+  private static Schema getSchema(String name) throws SchemaNotFoundException {
+    final Schema schema = schemas.get(name);
+    if (schema == null) {
+      throw new SchemaNotFoundException(name);
+    }
+    return schema;
+  }
```
jklukas: Added `getSchema(String)` so that we can still throw SchemaNotFoundException if the given schema is not present in the map.

jklukas:
I subtracted the code as requested, but ended up adding some more to ensure we're still checking errors.

PR #86, opened 2018-09-18
Add metrics to several Decoder transforms

### LINKED ISSUE #38: Add metrics to Sink and Validate
I'm not sure if Beam includes default metrics, but there are a few things we should be at least counting to understand failure modes.

Beam provides an [example pipeline that uses beam Metrics](https://github.com/apache/beam/blob/master/examples/java/src/main/java/org/apache/beam/examples/DebuggingWordCount.java) we can follow as a guide.

Some things we may want to count:

- How far we get in extracting location info in Validate; we could increment one counter on every event, another counter after getting a hostname from IP, another after extracting city, etc.

jklukas:
Closes #38

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
@@ -4,6 +4,7 @@
 
 package com.mozilla.telemetry.decoder;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.maxmind.db.CHMCache;
 import com.maxmind.geoip2.DatabaseReader;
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -92,7 +109,7 @@ public PubsubMessage apply(PubsubMessage value) {
     }
   }
 
-  private final Fn fn = new Fn();
+  private  Fn fn = new Fn();
```
relud: why not final? also nit: `s/(private ) /\1/`

jklukas: This change was unintentional. Will revert.

PR #87, opened 2018-09-18
Bug: Allow dash and underscore in path attributes

jklukas:


ingestion-beam/src/test/java/com/mozilla/telemetry/SinkTest.java
```diff
@@ -59,7 +60,7 @@ public void mainCanWriteToStdout() {
     });
   }
 
-  @Test
+  @Test(expected = PipelineExecutionException.class)
```
jklukas: I originally intended for this test to declare an expected exception, but realized it never did throw what I intended. It wasn't interpreting `${some_attribute}` as a placeholder. Now it does and it throws as expected.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
```diff
@@ -142,7 +142,7 @@ public String replaceDynamicPart(List<String> values) {
       // This regex matches either of these patterns,
       // capturing the name of the placeholder as group 1
       // and the default value (or null if no default is provided) as group 2.
-      "\\$\\{(\\p{Alnum}+)(?::-(\\p{Alnum}+))?}"
+      "\\$\\{([A-Za-z0-9-_]+)(?::-([A-Za-z0-9-_]+))?}"
```
relud: Attribute keys and values can be any string, so I think we should make this a lot broader, like: `"\\$\\{([^${:}]+)(?::-([^${:}]*))?}"`. This allows anything but special characters, and lets the default be empty.

I think we also need to expand this to throw an exception if there are any `${` that don't map to a placeholder, but that can be filed as an issue for later.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
```diff
@@ -142,7 +142,7 @@ public String replaceDynamicPart(List<String> values) {
       // This regex matches either of these patterns,
       // capturing the name of the placeholder as group 1
       // and the default value (or null if no default is provided) as group 2.
-      "\\$\\{(\\p{Alnum}+)(?::-(\\p{Alnum}+))?}"
+      "\\$\\{([^{}]+?(?=:-|}))(?::-([^{}]+))?}"
```
jklukas: Updated the regex and documented the restrictions on naming.

The new regex allows all characters other than curly braces and relies on a reluctant quantifier (`+?`) and positive lookahead (`(?=:-|})`) to determine the end of the attribute name.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
jklukas: > I think we also need to expand this to throw an exception if there are any ${ that don't map to a placeholder, but that can be filed as an issue for later.

I'm not sure I fully understand the nuance you're getting at here. Would you be willing to file an issue with the specifics?

relud: yeah, i'll file an issue

PR #88, opened 2018-09-19
Implement disk queue in ingestion-edge

relud:
still need to add tests that forcibly trigger the queue to make sure it works, but this is ready for early review, if anyone has time.

### DIFF
--- a/ingestion-edge/ingestion_edge/disk_queue.py
+++ b/ingestion-edge/ingestion_edge/disk_queue.py
@@ -1,0 +1,70 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .conf import QUEUE_FILE
+from .conf import QUEUE_POSITION_FILE
+from flask import Blueprint
+from google.cloud.pubsub import PublisherClient
+from os import remove
+from os.path import exists
+from threading import Lock
+from typing import Tuple
+import ujson
+
+disk_queue = Blueprint("disk_queue", __name__)
+client = PublisherClient()
+read_lock = Lock()
+write_lock = Lock()
+
+
+def write(**kwargs):
+    with write_lock:
+        # TODO disable autoscaling
+        with open(QUEUE_FILE, "a") as q:
+            q.write(ujson.dumps(kwargs) + "\n")
+            q.flush()
+
+
+@disk_queue.route("/__flush__")
+def flush_queue() -> Tuple[str, int]:
+    if not exists(QUEUE_FILE):
+        # early return for empty queue
+        return "", 204
+    with read_lock:
+        # initialize queue state
+        done = False
+        start = 0
+        end = 0
+        if exists(QUEUE_POSITION_FILE):
+            with open(QUEUE_POSITION_FILE, "r") as qp:
+                start = int(qp.read())
+        try:
+            # process the queue
+            # TODO release write lock between line reads
+            with write_lock:
+                # TODO rotate queue file and handle rotated files first
+                with open(QUEUE_FILE, "r") as q:
+                    for line in q:
+                        if end < start:
+                            # TODO use q.seek(start)
+                            end += 1
+                            continue
+                        client.publish(**ujson.loads(line)).result()
+                        # update end only on success
+                        end += 1
+            remove(QUEUE_FILE)
+            done = True
+        finally:
+            # update the queue position
+            if end > start and not done:
+                # record queue position if we moved forward and didn't finish
+                with open(QUEUE_POSITION_FILE, "r+") as qp:
+                    qp.write(str(end))
+                    qp.flush()
+            elif done and start > 0:
+                # remove existing queue position because we finished
+                remove(QUEUE_POSITION_FILE)
+    # TODO enable autoscaling
+    # return number of queue requests successfully published
+    return str(end - start) + "\n", 200
--- a/ingestion-edge/ingestion_edge/disk_queue.py
+++ b/ingestion-edge/ingestion_edge/disk_queue.py
@@ -6,65 +6,93 @@ from .conf import QUEUE_FILE
 from .conf import QUEUE_POSITION_FILE
 from flask import Blueprint
 from google.cloud.pubsub import PublisherClient
-from os import remove
-from os.path import exists
-from threading import Lock
 from typing import Tuple
 import ujson
+import threading
+import os
+import os.path
+import sys
 
 disk_queue = Blueprint("disk_queue", __name__)
 client = PublisherClient()
-read_lock = Lock()
-write_lock = Lock()
+read_lock = threading.Lock()
+write_lock = threading.Lock()
 
 
-def write(**kwargs):
+def _remove_if_found(path):
+    try:
+        os.remove(path)
+    except FileNotFoundError:
+        pass
+
+
+def write(publish_kwargs):
+    # write lock allows one writer at a time to avoid splicing writes
     with write_lock:
         # TODO disable autoscaling
+        # TODO keep file pointer open and handle file rotation
         with open(QUEUE_FILE, "a") as q:
-            q.write(ujson.dumps(kwargs) + "\n")
+            ujson.dump(publish_kwargs, q)
+            q.write("\n")
             q.flush()
 
 
 @disk_queue.route("/__flush__")
 def flush_queue() -> Tuple[str, int]:
-    if not exists(QUEUE_FILE):
-        # early return for empty queue
-        return "", 204
+    published = 0
+    # read_lock allows one flush at a time to avoid duplicate publish calls
     with read_lock:
         # initialize queue state
-        done = False
-        start = 0
-        end = 0
-        if exists(QUEUE_POSITION_FILE):
-            with open(QUEUE_POSITION_FILE, "r") as qp:
-                start = int(qp.read())
         try:
-            # process the queue
-            # TODO release write lock between line reads
-            with write_lock:
-                # TODO rotate queue file and handle rotated files first
-                with open(QUEUE_FILE, "r") as q:
-                    for line in q:
-                        if end < start:
-                            # TODO use q.seek(start)
-                            end += 1
-                            continue
-                        client.publish(**ujson.loads(line)).result()
-                        # update end only on success
-                        end += 1
-            remove(QUEUE_FILE)
-            done = True
+            with open(QUEUE_POSITION_FILE, "rb") as qp:
+                start = int.from_bytes(qp.read(), sys.byteorder)
+        except FileNotFoundError:
+            start = 0
+        done = False
+        end = start
+        try:
+            # process queue
+            # TODO rotate queue file and handle rotated files first
+            # NOTE rotated files will not require write_lock
+            with open(QUEUE_FILE, "r") as q:
+                # seek to previous end position
+                q.seek(start)
+                # can't use `for line in q` because it disables `q.tell()`
+                # and doesn't allow for releasing write_lock between line reads
+                # or using formats with framing other than newline
+                while True:
+                    # write_lock ensures we don't read a partial line
+                    # or write to queue immediately before removing it
+                    with write_lock:
+                        line = q.readline()
+                        if not line:
+                            # at EOF, remove queue and break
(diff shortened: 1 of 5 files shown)


ingestion-edge/requirements.txt
```diff
@@ -8,3 +8,4 @@ pytest==3.8.0
 pytest-cov==2.6.0
 pytest-flake8==1.0.2
 python-dateutil==2.7.3
+ujson==1.35
```
relud: please ignore that this still isn't hashed req's yet, filed #89 to fix

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+disk_queue = Blueprint("disk_queue", __name__)
+client = PublisherClient()
+read_lock = Lock()
+write_lock = Lock()
+
+
+def write(**kwargs):
+    with write_lock:
+        # TODO disable autoscaling
+        with open(QUEUE_FILE, "a") as q:
```
jklukas: I can understand that we definitely want to flush every time we write here, but I'm not sure how expensive it is to open the file. If opening the file is more expensive than the flush, we may want to move that outside this method. It's probably not worth investigating until we're ready to do performance testing.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+        try:
+            # process the queue
+            # TODO release write lock between line reads
+            with write_lock:
+                # TODO rotate queue file and handle rotated files first
+                with open(QUEUE_FILE, "r") as q:
+                    for line in q:
+                        if end < start:
+                            # TODO use q.seek(start)
+                            end += 1
```
jklukas: My assumption on first read-through was that `start` and `end` were byte positions in the file, but looks like they're line numbers? Will that need to be refactored in order to use `seek`? We might want to use more specific variable names to make this easier to follow.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, you can obtain one at http://mozilla.org/MPL/2.0/.
+
+from .conf import QUEUE_FILE
+from .conf import QUEUE_POSITION_FILE
+from flask import Blueprint
+from google.cloud.pubsub import PublisherClient
+from os import remove
+from os.path import exists
```
jklukas: I tend to prefer importing `os` and `os.path` as packages and then referencing these functions with fully qualified names (os.remove, os.path.exists) since the function names themselves are so short. In particular, I had assumed the call to `remove` below was some locally defined function.

Similarly, this would read more clearly to me if we used `threading.Lock()` below.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+def write(**kwargs):
+    with write_lock:
+        # TODO disable autoscaling
+        with open(QUEUE_FILE, "a") as q:
+            q.write(ujson.dumps(kwargs) + "\n")
+            q.flush()
+
+
+@disk_queue.route("/__flush__")
+def flush_queue() -> Tuple[str, int]:
```
jklukas: How will this get called? Is this going to be part of the kubernetes configuration for the pod lifecycle? A docstring here would be helpful.

jklukas:
https://github.com/balena/python-pqueue is the most promising option I've come across. In particular, it does file rotation, maintains state files, and provides a transaction-like API where we can get data and then acknowledge that we're done with it. Sounds like it relies on pickling, though, which we probably want to avoid.

Looks like it's one of a few offshoots of https://github.com/scrapy/queuelib

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import Tuple
+import ujson
+
+disk_queue = Blueprint("disk_queue", __name__)
+client = PublisherClient()
+read_lock = Lock()
+write_lock = Lock()
+
+
+def write(**kwargs):
```
mythmon: I think it would be more appropriate to take a single argument, instead of `**kwargs`. Semantically, this is taking a single item to write, not a set of key/value pairs.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+    with write_lock:
+        # TODO disable autoscaling
+        with open(QUEUE_FILE, "a") as q:
+            q.write(ujson.dumps(kwargs) + "\n")
```
mythmon: I'm not sure about `ujson`, but the standard library `json` has both `json.dumps` and `json.dump`, where `dumps` is "**dump** to **s**tring". If `ujson` works the same, I expect this would be better as `ujson.dump(kwargs, q)`. That wouldn't automatically include the newline at the end, but I think an additional `q.write("\n")` would be fine.

Mostly I expect this to be more efficient than generating an intermediate string and then writing it to a file, but I've never profiled it.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+            q.flush()
+
+
+@disk_queue.route("/__flush__")
+def flush_queue() -> Tuple[str, int]:
+    if not exists(QUEUE_FILE):
+        # early return for empty queue
+        return "", 204
```
mythmon: "if exists, open" is generally frowned upon, as it can introduce race conditions. This also doesn't guard again a queue file that is unreadable for other reasons (such as permissions). The standard advice is to open the file and catch the error, reacting appropriately.

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -0,0 +1,70 @@
...
+                        client.publish(**ujson.loads(line)).result()
+                        # update end only on success
+                        end += 1
+            remove(QUEUE_FILE)
+            done = True
+        finally:
+            # update the queue position
+            if end > start and not done:
+                # record queue position if we moved forward and didn't finish
+                with open(QUEUE_POSITION_FILE, "r+") as qp:
```
mythmon: Why do you want mode `r+` here? The docs tell me that this opens the file for reading and writing, putting the file pointer at the beginning of the file. That would mean that if `end` changed from a 4 digit number to a 3 digit number, the file would become corrupt.

I think mode `w` would be more appropriate here, which truncates the file and opens it for writing.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -39,8 +39,12 @@ def handle_request(topic: str, **kwargs) -> Tuple[str, int]:
             if value is not None
         }
     )
-    # returns message_id, which we don't use
-    future.result()
+    try:
+        # publish message
+        client.publish(**publish_kwargs).result()
+    except Exception:  # TODO only catch transient exceptions
```
mythmon: I think your TODO already covers this, but I want to be sure: `except Exception` almost always masks unexpected problems. Catching more specific errors is generally a better idea.

ingestion-edge/ingestion_edge/disk_queue.py
mythmon: I agree that using byte positions would be better.

ingestion-edge/ingestion_edge/publish.py
relud: i am absolutely not going to leave this catching `Exception`, I just haven't figured out what the block throws yet.

ingestion-edge/ingestion_edge/disk_queue.py
relud: I will add a docstring. k8s needs to call this endpoint regularly on each container. i don't know if this is possible, but it's the only solution I've come up with that doesn't cause problems.

ingestion-edge/ingestion_edge/disk_queue.py
relud: yes. the `TODO` comment here is for using byte positions,

ingestion-edge/ingestion_edge/disk_queue.py
relud: opening the file is expensive, checking for a changed inode and reusing a file handle and seeking to EOF is faster but more complicated so I haven't written it yet.

ingestion-edge/ingestion_edge/disk_queue.py
relud: i don't want `r+`, I thought it did something it doesn't (truncate after write on close, which i also don't want now that i consider it)

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -6,65 +6,93 @@
...
+                            # remove position first to prevent corrupted state
+                            _remove_if_found(QUEUE_POSITION_FILE)
+                            _remove_if_found(QUEUE_FILE)
+                            done = True
+                            break
+                    # when this throws an exception, stop flushing
+                    # TODO return 202 for transient exceptions
+                    client.publish(**ujson.loads(line)).result()
+                    # record new file position
+                    end = q.tell()
```
mythmon: I didn't know what `.tell()` did, so I looked it up. Among other things, the docs say

> f.tell() returns an integer giving the file object’s current position in the file represented as number of bytes from the beginning of the file when in binary mode and an opaque number when in text mode.

I'm not sure what they mean by "opaque number". I suspect this is how they guard against text encoding issues in text mode, where one character can be more than one byte. I'm not sure if this is something to worry about or not.

The docs say that you can use the results of `q.tell()` to seek from the beginning of a file in text mode, but that doing so from the middle of the file is undefined behavior. I think that means that you're in the clear here?

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -6,65 +6,93 @@
...
+        except FileNotFoundError:
+            # no queue to flush
+            return "", 204
         finally:
-            # update the queue position
-            if end > start and not done:
+            # update queue position if we didn't finish but made progress
+            if not done and end > start:
+                # use upside down floor division for ceiling division
+                end_length = -(-end.bit_length() // 8)
```
mythmon: I think this would be clearer as

```python
import math
end_length = math.ceil(end.bit_length / 8.0)
```

ingestion-edge/ingestion_edge/disk_queue.py
```diff
@@ -6,65 +6,93 @@
...
+                end_bytes = end.to_bytes(end_length, sys.byteorder)
                 # record queue position if we moved forward and didn't finish
-                with open(QUEUE_POSITION_FILE, "r+") as qp:
-                    qp.write(str(end))
+                with open(QUEUE_POSITION_FILE, "wb") as qp:
+                    qp.write(end_bytes)
```
mythmon: I think it is an improvement to seek based on byte position instead of line number, but I'm not as convinced that storing the queue position as bytes instead of text is a win. It seems more complicated, and would make it a lot harder to read the queue file by hand if it came to that. It saves a few bytes, but the open/read/close loop is surely dominated by opening the file, not reading a few bytes.

relud:
@jklukas I had looked into those but they aren't thread-safe:

https://github.com/scrapy/queuelib#queuelib
> Note: Queuelib isn't thread-safe.

https://github.com/balena/python-pqueue#how-to-use
> Note that pqueue is not intended to be used by multiple processes.

There's also `[persist-queue](https://github.com/peter-wangxu/persist-queue)` which is thread-safe, and would use pickle protocol version 4, which is supposed to be reasonably fast, and fixes a bug where using json isn't safe with gzipped request data.

`persist-queue` was the very first solution i found when considering this, and I am honestly not sure why I didn't go with it. I think I was avoiding pickle? I just read through their file implementation, and I think it has almost all the semantics we need. state is updated on every write, all the right locks in all the right places, rotated files, etc.

One limitation of `persist-queue` is that [the file queue can't tell the difference between tasks in the queue](https://github.com/peter-wangxu/persist-queue/blob/master/persistqueue/queue.py#L215), so we need to process the queue single-threaded in order to ensure we are always acking the message we actually sent.

I'm going to close this and file a new PR to use `persist-queue`

PR #91, opened 2018-09-20
Implement disk queue in ingestion-edge

relud:


ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,13 +3,40 @@
...
+                item = q.get_nowait()
+            except Empty:
+                break
+            try:
+                client.publish(**item).result()
+                q.task_done()
+                tasks_done += 1
+            except Exception:
+                return str(tasks_done), 202
+    return str(tasks_done), 200
```
jklukas: Let's make these 200 and 202 literals named constants to better document intention.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,13 +3,40 @@
...
+            try:
+                item = q.get_nowait()
+            except Empty:
+                break
+            try:
+                client.publish(**item).result()
+                q.task_done()
+                tasks_done += 1
+            except Exception:
+                return str(tasks_done), 202
```
jklukas: Probably worth adding a comment here to explain that by returning without calling `q.task_done()`, we aren't updating the cursor. I assume the intention here is that the next call to __flush__ will try again to send the message.

Do we need a TODO here to investigate whether there are cases where a message could fail infinitely? Something like a particularly large message might cause a non retryable error.

Or would we rely on monitoring at that point to let us know the records aren't getting flushed and that we need to investigate?

mythmon: I'm not sure that 202 is the right status code here. 202 implies that there is a chance the work will be done in the future even if no other requests happen. From [the RFC](https://tools.ietf.org/html/rfc7231#section-6.3.3):

> The representation sent with this response ought to describe the request's current status and point to (or embed) a status monitor that can provide the user with an estimate of when the request will be fulfilled.

Looking through the status codes, I don't see any thing more appropriate than `200 OK`. Maybe if more details are needed, you could return a JSON object that has something like number of items processed/number remaining?

relud: good point, that is the intention, but that's not what would always happen. `except Exception` is meant to be replaced with something that only catches transient exceptions, so that we don't retry indefinitely.

relud: how about 503? we didn't completely flush the queue, and we are temporary unable to do so. From [the RFC](https://tools.ietf.org/html/rfc7231#section-6.6.4):

> The 503 (Service Unavailable) status code indicates that the server is currently unable to handle the request due to a temporary overload or scheduled maintenance, which will likely be alleviated after some delay.

mythmon: 503 seems more appropriate. It is definitely different than 202 though, since 202 implied success and 503 implies failure. Client libraries might care about that, but as long as you're consistent it shouldn't be a big deal.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,13 +3,50 @@
...
+                q.task_done()
+            except Empty:
+                break
+            try:
+                client.publish(**item).result()
+                done += 1
+            except Exception:  # TODO only catch transient exceptions
+                # NOTE: this queue currently must requeue items on failure
+                q.put(item)
+                msg = '{"done":%d, "pending":%d}\n' % (done, q.qsize())
```
mythmon: Why do json interpolation yourself instead of using the stdlib `json`?

relud: *shrug* `json` is slow? i went from a non-json string to a json one and didn't think about it

mythmon: You do you, but I wouldn't hand code json using string formatting until the profiler told me this line was slow, and even then I'd be really reluctant.

relud: i will not do me. i will do json, like you suggest. seems good.

relud:
re-requesting reviews because of significant changes

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,44 +3,135 @@
...
+    ServerError,
+    # A python Future timed out
+    TimeoutError,
+    # The API is throttling requests
+    TooManyRequests,
+)
 
 publish = Blueprint('publish', __name__)
 client = PublisherClient()
+q = SQLiteAckQueue(QUEUE_PATH)
```
jklukas: Did we switch to the SQLite queue because it was more stable? Can we add a comment here to give some context?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,44 +3,135 @@
...
+
+
+def _publish(topic, data, attrs) -> str:
+    """Publish one message to the topic
+
+    Uses PublisherClient to try to publish the message in a batch and handles
+    non-transient batch failure by retrying synchronously with a single message
+    """
+    try:
+        return client.publish(topic, data, **attrs).result()
```
jklukas: I want to make sure I understand this correctly. We have a `client` defined in the module scope. One instance of this Flask app will have many threads running and all those threads will be accessing the same client. When a thread hits this line, it blocks waiting for the client to complete and release a batch. So a batch will contain at most n messages where n is the number of Flask application threads. A single thread is only ever handling one message at a time; it never moves on to the next message until the pubsub client acknowledges delivery.

It may be worth writing up documentation on the semantics here, either in a docstring here or in a spot where we configure batching and threading settings.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,44 +3,135 @@
...
+        except Empty:
+            # queue is empty or sqlite needs a retry
+            continue
+        try:
+            try:
+                _publish(*message)
+            except:  # noqa: E722
+                # nack then re-raise on absolutely any exception
+                q.nack(message)
+                raise
```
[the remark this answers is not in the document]
jklukas: I think this works as written, but the nested `try` is a bit difficult to parse. I think I'd prefer duplication of the `q.nack` call so that this can be flatter:

```
try:
  _publish(*message)
except TRANSIENT_ERRORS:
  q.nack(message)
  response = ...
  return ...
except:
  q.nack(message)
  raise
else:
  q.ack(message)
  done += 1
```

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -3,44 +3,135 @@
...
+from .conf import QUEUE_PATH
 from datetime import datetime
 from flask import Blueprint, request
-from google.cloud.pubsub import PublisherClient
+from google.api_core.exceptions import Aborted
+from google.api_core.exceptions import Cancelled
+from google.api_core.exceptions import Forbidden
+from google.api_core.exceptions import RetryError
+from google.api_core.exceptions import ServerError
+from google.api_core.exceptions import TooManyRequests
```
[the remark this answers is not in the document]
jklukas: Since we only use each of these once to build the `TRANSIENT_ERRORS` tuple, I'd recommend importing the package and using fully qualified names:

```
import google.api_core.exceptions

TRANSIENT_ERRORS = (
    # Comment
    google.api_core.exceptions.Forbidden,
    # Comment
    google.api_core.exceptions.RetryError,
...
```

ingestion-edge/ingestion_edge/publish.py
relud: we switched to `SQLiteAckQueue` because `Queue` does not implement acks, which we need.

relud: that said, we should use a SQLite queue even if we don't need `ack` because it provides multi-process safe (and threadsafe if we had threads) access to the queue and it's [faster than `Queue`](https://github.com/peter-wangxu/persist-queue/issues/68).

ingestion-edge/ingestion_edge/publish.py
relud: TL;DR no it's asynchronous.

Note: my understanding here is based on very recent reading, and may be more detailed than you need, but it's all new information for me.

> We have a client defined in the module scope. 

Yes

> One instance of this Flask app will have many threads running and all those threads will be accessing the same client.

Not necessarily. With gunicorn each asynchronous worker (may be a process or thread) will have one instance of this Flask app. Each worker will have many coroutines. All of those coroutines will be accessing the same client. Coroutines differ from threads in that they are managed by python (or whatever language is using them) so they only get paused when python thinks it's okay, like when waiting on a `concurrent.futures.Future` or IO operation.

Note: our docker container is currently using the `sync` worker class by default. This will be changed to an asynchronous worker class like `gevent` in the future.

> When a thread hits this line, it blocks waiting for the client to complete and release a batch.

When a coroutine hits this line it should yields because `publish(...)` returns a `Future` which the coroutine waits for via `.result()`.

> So a batch will contain at most n messages where n is the number of Flask application threads.

Where n is the number of coroutines being run by this worker, and a worker will run a coroutine for every request it receives. The actual limit is lower because the client library also starts new batches based on limits for timeout, message count, and batch size.

> A single thread is only ever handling one message at a time; it never moves on to the next message until the pubsub client acknowledges delivery.

A single coroutine is only ever handling one message at a time, but a single worker may handle thousands of connections, each of which may contain multiple requests.

relud: > configure batching and threading settings

we should modify the ~~batching~~ client settings eventually because it defaults to a 60 second timeout on api calls, and we may call the api twice per request before queuing to disk, and I suspect that kind of latency is too high.

~~threading~~ worker settings are managed via env vars by ops. We should eventually add a section to each `README.md` indicating what settings exist and how they are set in stage/prod.

I think we should handle both of those in later PRs

ingestion-edge/ingestion_edge/publish.py
relud: also:

> Can we add a comment here to give some context?

absolutely

PR #97, opened 2018-09-24
Add test utilities for reading files and comparing contents

### LINKED ISSUE #54: Add a test helper for running `main` and validating output on the filesystem
As the project grows and number of test cases increases, we're going to want to handle the following issues:

1. Escaped JSON strings in Java source code are difficult to read and modify
2. We will want to be able to test the option parsing and pipeline building logic of the entry classes' `main` methods
3. We will want to test interaction with external datastores like GCS and Pubsub

This issue doesn't address (3), but we can address (1) and (2) by adding a test helper class that can read in newline-delimited JSON documents from Java resource files representing expected input and output, and compare to the results of an invocation of `main`. This helper would only support running pipelines with input and output types of `file`.

We'd likely want to use the [FileAssert utilities of assertj](http://joel-costigliola.github.io/assertj/assertj-core-features-highlight.html#file-content-string-assertions) or [implement a hamcrest matcher](https://stackoverflow.com/a/466978/1260237).

We'd also likely need to add a pipeline option for setting number of shards (so that tests can set nShards=1) or make our matcher capable of handling output lines being spread across many output files.

jklukas:
Closes #54

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/SinkMainTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/SinkMainTest.java
@@ -1,0 +1,109 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry;
+
+import static com.mozilla.telemetry.matchers.Lines.matchesInAnyOrder;
+import static org.junit.Assert.assertThat;
+
+import com.google.common.io.Resources;
+import com.mozilla.telemetry.matchers.Lines;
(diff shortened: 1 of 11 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/SinkMainTest.java
```diff
@@ -0,0 +1,109 @@
...
+        "--inputType=file",
+        "--input=" + input,
+        "--outputFileFormat=json",
+        "--outputType=file",
+        "--output=" + output,
+        "--errorOutputType=file",
+        "--errorOutput=" + errorOutput
+    });
+
+    List<String> validLines = Lines.resources("testdata/basic-messages-normalized.ndjson");
```
relud: for consistent naming this file should be `testdata/basic-messages-valid.ndjson`

Issue #74, opened 2018-09-13
Write tests for local IO in ingestion-beam's Sink pipeline


relud:
specifically we should test

- [x] `--inputType=file`
- [x] `--outputType=`
   - [x] `file`
   - [x] `stdout`
   - [ ] `stderr`
- [x] `--errorOutputType=`
   - [x] `file`
   - [ ] `stdout`
   - [ ] `stderr`

using local file paths, ideally with junit tests.

jklukas:
Is there any reason to do extensive testing of stdout and stderr outputs? Do we expect to ever have use cases for those, or are they simply for testing?

#54 is relevant here. I may have time early next week to start working on a Hamcrest matcher for comparing files without regard to message order.

relud:
we should focus on testing produse cases, so not stdout/stderr. we should add a warning when they are selected to tell people when they use a component that is not considered not prod quality.

jklukas:
See https://github.com/mozilla/gcp-ingestion/pull/97 which introduces utilities for doing this sort of testing on top of junit.

jklukas:
Closing this, as we have tests that exercise stdout and local file output for main output and error output.

PR #98, opened 2018-09-25
Add SinkMain tests for placeholders and text<->Json

jklukas:
Partially addresses #74


ingestion-beam/src/test/resources/testdata/basic-messages-valid-null-attributes.ndjson
```diff
@@ -0,0 +1,4 @@
+{"attributeMap":null,"payload":"dGVzdA=="}
+{"attributeMap":null,"payload":"dGVzdA=="}
+{"attributeMap":null,"payload":"dGVzdA=="}
+{"attributeMap":null,"payload":""}
```
relud: i think we should keep these in the same order as the source, even though the tests don't check order, so that it's more obvious to humans what is responsible for what.

jklukas: That's legit. I just pushed a change that puts the null payload as the third of 4 messages consistently and also removes the `mixed-validity` file in favor of a glob on the valid and invalid files, so we have one less file to keep in sync.

PR #100, opened 2018-09-25
Add util.Time and implement parseDuration using java 8 time API

jklukas:
This parses a wider variety of formats, removes the odd com.google.cloud.teleport package space, and will make the transition off of joda time easier.


ingestion-beam/src/main/java/com/mozilla/telemetry/util/Time.java
```diff
@@ -0,0 +1,95 @@
...
+          .replaceAll("minutes", "M")
+          .replaceAll("minute", "M")
+          .replaceAll("mins", "M")
+          .replaceAll("min", "M")
+          .replaceAll("hours", "H")
+          .replaceAll("hour", "H")
+          .replaceAll("days", "DT")
+          .replaceAll("day", "DT")
+          .replaceAll("\\s+", "")
+          .toUpperCase();
```
relud: I think we should set the case first, so that we don't care about matching case in `replaceAll`

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Time.java
```diff
@@ -0,0 +1,95 @@
...
+      if (!modifiedValue.contains("T")) {
+        modifiedValue = "T" + modifiedValue;
+      }
+      if (modifiedValue.endsWith("T")) {
+        modifiedValue += "0S";
+      }
+      modifiedValue = "P" + modifiedValue;
```
relud: we should only add `"P"` if it's not in the value, so we can handle things like `P1day`

ingestion-beam/src/test/java/com/mozilla/telemetry/util/TimeTest.java
```diff
@@ -0,0 +1,36 @@
...
+package com.mozilla.telemetry.util;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class TimeTest {
+
+  @Test
+  public void testParseDuration() {
```
relud: we should test `ValueProvider` mode too

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Time.java
```diff
@@ -0,0 +1,95 @@
...
+
+    return toJoda(duration);
+  }
+
+  /**
+   * Like {@link #parseDuration(String)}, but using a {@link ValueProvider}.
+   *
+   * <p>The value will be parsed once when first used; the result is cached and reused.
+   */
+  public static ValueProvider<org.joda.time.Duration> parseDuration(ValueProvider<String> value) {
```
relud: optional side-quest: implement `public static ValueProvider<Integer> parseDurationSeconds(ValueProvider<String> value)` so that `DecoderOptions.Parsed` can provide what `Deduplicate.MarkAsSeen` actually needs

jklukas: When you label it an "optional side-quest", how can I refuse? You expect me to finish without 100% items collected?

ingestion-beam/src/test/java/com/mozilla/telemetry/util/TimeTest.java
jklukas: Added

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Time.java
jklukas: We now lowercase first, do the replacements, and uppercase the result.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/TimeTest.java
jklukas: Added

jklukas:
Marking for review again for validation of the side quest such that I might obtain the Sparkling Golden Staff of Unit Specification.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
```diff
@@ -111,23 +110,15 @@ protected PubsubMessage processElement(PubsubMessage element) throws DuplicateId
...
-    private MarkAsSeen(ValueProvider<URI> uri, ValueProvider<Duration> ttl) {
+    private MarkAsSeen(ValueProvider<URI> uri, ValueProvider<Long> ttlSeconds) {
       super(uri);
-      this.ttl = ttl;
+      this.ttlSeconds = ttlSeconds;
     }
 
     private String setex(byte[] id) {
-      return getJedis().setex(id, getTtlSeconds(), new byte[0]);
+      return getJedis().setex(id, ttlSeconds.get().intValue(), new byte[0]);
```
relud: Instead of putting off the `.intValue()` until here, I think we should use `ValueProvider<Integer>`, unless that also requires a method call, in which case this is fine

jklukas: I was using Long because java.time.Duration stores seconds as a long, so it seemed more natural to allow that precision.

Thinking more, though, I agree with you that it makes sense to narrow to `int` as soon as possible, since the whole point of this value is to pass it to the jedis API which expects and int.

I've left the `parseSeconds` method as returning a long, but now use a lambda inside `parseDecoderOptions` to cast to int and throw an exception on first access if the provided value overflows integer.

In general, it seems like good practice to do as much validation within `parseDecoderOptions` as possible, so that the parsed options are as safe and natural to use elsewhere as possible.

PR #102, opened 2018-09-25
Fix DecoderOptions.parse 

### LINKED ISSUE #101: Fix DecoderOptions.parse
An exception occured while executing the Java class. Method [parse] has multiple definitions [public static com.mozilla.telemetry.decoder.DecoderOptions$Parsed com .mozilla.telemetry.decoder.DecoderOptions.parse(com.mozilla.telemetry.decoder.DecoderOptions), public static com.mozilla.telemetry.options.SinkOptions$Parsed com.mozilla.telemetry.options.SinkOptions.parse(com.mozilla.telemetry.options.SinkOptions)] with different return types for [com.mozilla.telemetry.decoder.DecoderOptions]. -> [Help 1]

jklukas:
Fixes #101 

Also adds DecoderMainTest to exercise DecoderOptions.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
@@ -64,7 +64,7 @@ public interface DecoderOptions extends SinkOptions, PipelineOptions {
    * A custom {@link PipelineOptions} that includes derived fields.
    *
    * <p>This class should only be instantiated from an existing {@link DecoderOptions} instance
-   * via the static {@link #parse(DecoderOptions)} method.
+   * via the static {@link #parseDecoderOptions(DecoderOptions)} method.
    * This follows a similar pattern to the Beam Spark runner's {@code SparkContextOptions}
    * which is instantiated from {@code SparkPipelineOptions} and then enriched.
    */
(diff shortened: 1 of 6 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -79,17 +79,17 @@
...
     return parsed;
   }
 
   /**
    * Set all the derived fields of a {@link DecoderOptions.Parsed} instance.
    */
-  static void enrich(Parsed options) {
-    SinkOptions.enrich(options);
+  static void enrichDecoderOptions(Parsed options) {
+    SinkOptions.enrichSinkOptions(options);
```
relud: I think we can drop the `SinkOptions.` here now?

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
```diff
@@ -0,0 +1,54 @@
...
+        "--outputFileFormat=json",
+        "--outputType=file",
+        "--output=" + output,
+        "--errorOutputType=file",
+        "--errorOutput=" + errorOutput,
+        "--geoCityDatabase=GeoLite2-City.mmdb",
+        "--seenMessagesSource=none"
+    });
+
+    List<String> errorOutputLines = Lines.files(outputPath + "/error*.ndjson");
```
relud: I think `outputPath + "/error*.ndjson"` -> `errorOutput + "*.ndjson"` is more clear (I was confused at first why we used `outputPath` for `errorOutputLines`)

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
jklukas: IntelliJ tells me "Static method may be invoked on containing interface class only", so looks like we need to qualify it. Inheritance of static methods seems like a bit of an ugly corner of Java's type system.

PR #109, opened 2018-09-28
Replace Flask with Sanic

relud:
finally got sanic functional


ingestion-edge/requirements.txt
```diff
@@ -1,6 +1,5 @@
-blinker==1.4
-dockerflow==2018.4.0
-Flask==1.0.2
+aiohttp==3.4.4
+git+git://github.com/relud/python-dockerflow@master#egg=dockerflow
```
relud: temporary solution until dockerflow merges Sanic support and makes a release.

this should probably block merging this PR

ingestion-edge/tests/conftest.py
```diff
@@ -2,16 +2,22 @@
...
-def client() -> Generator[FlaskClient, None, None]:
+def client() -> Generator[SanicTestClient, None, None]:
     with TemporaryDirectory() as tmp:
         app = create_app(QUEUE_PATH=tmp, TESTING=True)
-        with app.test_client() as client:
-            yield client
+
+        @app.exception(Exception)
+        def handler(request, exception):
+            print(traceback.format_exc(), file=sys.stderr)
```
relud: because sanic won't print tracebacks by default

ingestion-edge/contract-tests/test_api.py
```diff
@@ -97,7 +101,7 @@ def test_publish(
         rec_time = parse(attrs.pop("submission_timestamp")[:-1])
         assert (req_time - rec_time).total_seconds() < 1
         assert "remote_addr" in attrs
-        attrs.pop("remote_addr")
+        assert attrs.pop("remote_addr") != ""
```
mythmon: I think this would be better as `assert attrs.pop("remote_addr")`, which uses that the empty string is falsey. It would also guard against `None`, if for whatever reason that was passed (though I don't think `None` is likely/possible?)

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -55,20 +56,22 @@ def _publish(topic: str, data: bytes, attrs: Dict[str, str]) -> str:
     worker.
     """
     try:
-        return client.publish(topic, data, **attrs).result()
+        return await async_wrap(client.publish(topic, data, **attrs))
```
mythmon: As a random aside, in JS `return await foo` is at best a no-op, and at worst it spins the event loop an extra time. This is because async functions mostly desugar into functions that return promises. Because of that, some style guides suggest replacing `return await foo` with `return foo` in JS (though personally I leave it to make it clearer that `foo` is an async thing.

Python doesn't work that way, and `return foo` would make a coroutine function that returns another coroutine. So `return await` is required in Python. It is interesting to compare the two languages.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -55,20 +56,22 @@ def _publish(topic: str, data: bytes, attrs: Dict[str, str]) -> str:
...
-        if len(response.message_ids) != 1:
-            raise PublishError("message was not successfully published")
-        return response.message_ids[0]
+        future = batch.publish({'data': data, 'attributes': attrs})
+        batch.commit()  # spawns thread to commit batch
+        return await async_wrap(future)
 
 
-def flush(q: SQLiteAckQueue) -> Tuple[str, int]:
+async def flush(request: Request, q: SQLiteAckQueue) -> Tuple[str, int]:
```
mythmon: Is the return type annotation here still correct? I'd expect the new type to be a Response object, or maybe some sort of coroutine thing?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -131,34 +136,38 @@ def submit(q: SQLiteAckQueue, topic: str, **kwargs) -> Tuple[str, int]:
         if value is not None
     }
     try:
-        _publish(topic, data, attrs)
+        await _publish(topic, data, attrs)
     except TRANSIENT_ERRORS:
         # transient api call failure, write to queue
         q.put((topic, data, attrs))
-    return "", 200
+    return response.raw(b"")
```
mythmon: I probably should have noticed this before, but should this be [204 No Content](https://http.cat/204) instead?

ingestion-edge/ingestion_edge/publish.py
relud: ah, it is not correct. async functions return a coroutine when run, but I would expect `Tuple[str, int]` to be what you provide to `return`. The real problem is that this function now returns a `sanic.response.HTTPResponse` object.

ingestion-edge/ingestion_edge/publish.py
relud: I would like to return 204, but [the spec says return 200](https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#postput-response-codes)

That said, I think the above `try` clause needs an `except: return response.raw(b"", 400)`, or something because a non-transient request failure indicates a bad request that should not be retried.

relud: perhaps [422](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422) is appropriate here?

mythmon: 422 is part of [RFC4918](https://tools.ietf.org/html/rfc4918#section-11.2), which is for WebDAV. I try and avoid WebDAV. I think just 400 would be fine, since there aren't really multiple failure cases here.

relud: eh, spec says 400, can be updated to 422 later if needed

ingestion-edge/constraints.txt
```diff
@@ -0,0 +1,82 @@
+aiofiles==0.4.0 \
```
jklukas: What is the difference between requirements and constraints? When I've seen multiple requirements files before they've been named `requirements-dev.txt`, etc. Could this be named `requirements-<something>.txt` to make it more obvious it's a pip requirements file?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -77,8 +78,8 @@ def flush(q: SQLiteAckQueue) -> Tuple[str, int]:
     """
     if q.size == 0:
         # early return for empty queue
-        return "", FLUSH_EMPTY_STATUS
-    done = 0
+        return response.raw(b"", HTTP_STATUS.NO_CONTENT)
+    result = {"done": 0, "pending": 0}
```
jklukas: Given that we make reference to these `done` and `pending` fields via string accessors several times, I'd like to  see them turned into string constants, or even better we can probably turn this into a class:

```
@dataclass 
Result:
  done: int = 0
  pending: int = 0

  def json(self):
    return json.dumps(self.__dict__)
```

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -131,34 +134,45 @@ def submit(q: SQLiteAckQueue, topic: str, **kwargs) -> Tuple[str, int]:
...
     # Use a SQLiteAckQueue because:
     # * we use acks to ensure messages only removed on success
     # * persist-queue's SQLite*Queue is faster than its Queue
     # * SQLite provides thread-safe and process-safe access
-    q = SQLiteAckQueue(**app.config.get_namespace("QUEUE_"))
+    q = SQLiteAckQueue(**{
+        key[6:].lower(): value
+        for key, value in app.config.items()
+        if key.startswith("QUEUE_")
+    })
```
jklukas: This is maybe a bit too terse. Readability would be improved if we name the dict we're defining inline. Perhaps:

```
queue_config = {key[6:]...}
q = SQLiteAckQueue(**queue_config)
```

ingestion-edge/constraints.txt
relud: `requirements.txt` is the packages to install, `constraints.txt` constrains versions that can be installed for dependencies.

It doesn't make sense to name it `requirements-<something>.txt` imo, because it's not a pip requirements file.

I would be okay with moving both files into a directory, something like `requirements/constraints.txt` and `requirements/defaults.txt`. I don't particularly like `defaults.txt` but I can't think of a better name.

ingestion-edge/ingestion_edge/publish.py
relud: `.json` won't do because we should leave json encoding to `response.json`, making it `response.json(result.__dict__, <status>)` or `def as_dict(self): self.__dict__` or something.

seems like overkill, i'd rather go back to using two variables and a `dict` call.

relud: actually, i take that back, python's type system don't make either of those options notably safer than this.

relud: specifically note that the following works and passes flake8:
```from dataclasses import asdict, dataclass
import json


@dataclass
class Result:
    done: int = 0
    pending: int = 0


r = Result()
r.don = 1
print(json.dumps(asdict(r)))
```
and it prints `{"done": 0, "pending": 0}`

relud: long way of saying: i'll just use string constants

PR #111, opened 2018-10-15
Improve docstring for _publish

relud:
I did a review of all of ingestion-edge with @mythmon and the docstring wasn't clear enough about why `PublishError` was being retried manually, so for a while we thought it wasn't necessary.

This updates the docstring to be more clear about the fact that `PublishError` is non-transient and has to be retried so that we don't reject messages in the batch that aren't suffering a permanent failure.

I also changed the direct api call to use a dict instead of `PubsubMessage`. I found out that works while reading the api docstrings and I think it's better than touching the pubsub sdk internals more than necessary.

### DIFF
--- a/ingestion-edge/ingestion_edge/publish.py
+++ b/ingestion-edge/ingestion_edge/publish.py
@@ -9,7 +9,6 @@ from flask import Flask, request
 from functools import partial
 from google.cloud.pubsub_v1 import PublisherClient
 from google.cloud.pubsub_v1.publisher.exceptions import PublishError
-from google.cloud.pubsub_v1.types import PubsubMessage
 from persistqueue import SQLiteAckQueue
 from persistqueue.exceptions import Empty
 from typing import Dict, Tuple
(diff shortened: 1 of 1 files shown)


ingestion-edge/ingestion_edge/publish.py
```diff
@@ -44,21 +43,25 @@
...
+
+    Note that this method will block until it gets a result, but gevent workers
+    should yield the coroutine during IO to handle multiple connections per
+    worker.
     """
     try:
         return client.publish(topic, data, **attrs).result()
     except PublishError:
-        # Batch failed because pubsub rejected at least one message
+        # Batch failed because pubsub permanently rejected at least one message
```
mythmon: So if this is the message that was permanently rejected, and this method raises `PublishError`, then that would be caught by the general `except:` clause on line 97, which would cause the message to be `nack`ed. If I'm understanding right, that would mean the message would be put back in the queue and attempted again later? If it is a permanent failure, is that the desired outcome?

relud: a message with a permanent failure shouldn't ever end up in the queue, but if it does then it would be retried indefinitely.

The expected behavior is that a message that raises `PublishError` never goes into the queue in the first place, and results in a 500 error at line 134.

mythmon: Could a transient error "outrank" a `PublishError`? That is, could you get a message that at first causes a transient error to happen, because of some outage, and gets added to the queue, and then in `flush` the message gets processed again, and causes a permanent failure? If that happened, the queue would fill up with permanent-error-causing messages.

relud: hmm, you're right, that is definitely broken. I will file an issue to fix that bug and add a test case for it.

relud: filed #112

PR #114, opened 2018-10-19
Add design documentation for whole pipeline

relud:
Mostly just copied information from the [GCP Architecture Proposal](https://docs.google.com/document/d/18CaLmI7lUYUC2pHsfCFM3TFBdzn-PPj4kpLPxhwon3A)

I left out the `Workflows` section, because it was incomplete, but I have it in another branch.


docs/architecture.mmd
```diff
@@ -0,0 +1,23 @@
+graph TD
```
jklukas: As the docs structure is getting more complex, there's risk of it turning into a wasteland of random files. I expect thus would feel cleaner if we moved md documents into README.md inside subdirectories along with any supporting files the README needs. So this would become `docs/architecture/`

docs/design.md
```diff
@@ -0,0 +1,144 @@
...
+  - [Saving as JSON](#saving-as-json)
+  - [Use destination tables?](#use-destination-tables)
+  - [Use views for user-facing data?](#use-views-for-user-facing-data)
+- [Known Issues](#known-issues)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Architecture Diagram
+
+![alt text](architecture.svg "Architecture Diagram")
```
jklukas: Can you put a comment in here about how to modify the mermaid diagram?

jklukas: Also, replace the `alt text` placeholder here.

jklukas: I think the color coding in the diagram is really useful, so you could put some text after the diagram to explicitly call out what the colors mean and link to the subprojects where the relevant code lives.

docs/design.md
```diff
@@ -0,0 +1,144 @@
...
+        including delayed `submission_date` and usually some amount of data
+loss.
+   - The PubSub Topic for data is configurable based on `uri`.
+   - Expected topics are Structured Ingestion, Telemetry, and Pioneer.
+- The Dataflow Landfill Sink job copies data from a PubSub Raw Topic to Cloud
+  Storage.
+   - Data can be read from Cloud Storage for backfill, recovery, or testing.
+- The Dataflow Decoder job decodes data from a PubSub Raw Topic to a PubSub
+  Decoded Topic.
+   - Decoding data applies these transforms in order:
```
jklukas: Consider linking to the `decoder` subdirectory of `ingestion-beam` here since these steps map directly to the files there.

docs/design.md
```diff
@@ -0,0 +1,144 @@
...
+transform.
+
+Dataflow jobs achieve at least once delivery by *not* using GroupByKey
+transforms and *not* falling more than 7 days behind in processing.
+
+## Design Decisions
+
+### Kubernetes and PubSub
+
+They are simple, scalable, managed services. Using them as opposed to a compute
```
jklukas: I hesitate to describe Kubernetes as "simple". It presents a relatively simple and now industry-standard interface for complex orchestration.

Perhaps "they are scalable managed services that are well-integrated standards in the GCP environment

docs/design.md
```diff
@@ -0,0 +1,144 @@
...
+One of the primary challenges of building a real-world data pipeline is
+anticipating and adapting to changes in the schemas of messages flowing through
+the system. Strong schemas and structured data give us many usability and
+performance benefits, but changes to the schema at one point in the pipeline
+can lead to processing errors or dropped data further down the pipeline.
+
+Saving as a raw JSON string allows us to gracefully handle new fields added
+upstream without needing to specify those fields completely before they are
+stored. The JSON can be integrated back into the table's schema via a load
+within BigQuery after adding the a column.
```
jklukas:
```suggestion
within BigQuery after adding the column.
```

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
+# Differences from AWS Architecture
+
+This document explains how GCP Ingestion differs from the AWS implementation.
+
+[AWS Data Platform Architecture](https://mana.mozilla.org/wiki/display/SVCOPS/Telemetry+-+Data+Pipeline+Architecture)
```
jklukas: This link looks like it's part of the table of contents. I'd move it to the sentence above and give it the text "AWS implementation (mozilla-internal diagram)"

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+- [Messages Always Delivered to Message Queue](#messages-always-delivered-to-message-queue)
+- [Landfill is Downstream from Message Queue](#landfill-is-downstream-from-message-queue)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+
+## Replace Heka Message Format with PubSub Message Format
+
+Heka messages are protobuf, and have a special framing format. PubSub messages
+are newline delimited JSON. The PubSub message format is being used both because
```
jklukas: "PubSub messsages are newline delimited JSON" doesn't sound right to me. The GCS files we store will be in ndjson format, but a Pubsub message itself does not necessarily contain JSON, and even when it does, it's probably a single json message rather than ndjson.

Perhaps this secion should be framed as "Replace Heka Message Format with Compressed Pubsub-formatted JSON"

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+it works well with PubSub and also because it can be read using Dataflow the
+same way from PubSub and Cloud Storage with minimal effort.
+
+## Replace EC2 Edge with Kubernetes Edge
+
+The AWS data platform uses EC2 instances running an NGinX module to encode HTTP
+requests as Heka messages, writes them to Kafka using librdkafka and directly to
```
jklukas:
```suggestion
requests as Heka messages, writing them to Kafka using librdkafka and directly to
```

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+files on disk. librdkafka handles buffering and batching when writing to Kafka.
+Files on disk are rotated with cron and uploaded to S3. On shutdown files are
+forcefully rotated and uploaded. The sizing of the EC2 instance cluster is
+effectively static, but is configured to scale up if needed.
+
+The EC2 instances have been replaced with a Kubernetes cluster. This decision
+was made by the Cloud Operations team to simplify operational support for them.
+
+The NGinX module has been replaced by an HTTP service running in Docker. A
+number of factors informed the decision to rewrite the edge, including:
```
jklukas: The "including:" here makes it feel like there should be another level of nesting here and we should have level 3 headings

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Replace Kafka with PubSub
+
+Comparison:
+
+|                | Kafka in AWS                    | PubSub                      |
+|----------------|---------------------------------|-----------------------------|
+| Managed by     | Ops                             | Google                      |
+| Access control | Security groups, all-or-nothing | Cloud IAM, per-topic        |
```
jklukas: Kafka does support per-topic ACLs, but I imagine it was too complex for us to easily take advantage of.

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+| Scaling        | Manual                          | Automatic                   |
+| Cost           | Per EC2 instance                | Per GB, min charge 1 KB/req |
+| Data Storage   | 2-3 days, depends on throughput | 7 days for unacknowledged   |
```
jklukas: 2-3 days here is again something we configured, not a limitation of Kafka as a technology. 

Perhaps the important difference is that storage is fixed and exhaustible for Kafka, but auto-scaling for PubSub.

docs/differences_from_aws.md
```diff
@@ -0,0 +1,101 @@
...
+GCP, the failsafe data store is downstream from the message queue (PubSub).
+
+This makes the edge and Dataflow landfill loader simpler. The edge doesn’t have
+to ensure that pending messages are safely offloaded on shutdown, because
+messages are only left pending when PubSub is unavailable, and scaling down is
+disabled while messages are pending. The Dataflow landfill loader doesn’t have
+to ensure pending messages are safely offloaded because it only acks messages
+after they are uploaded, ensuring at least once delivery.
+
+This design is possible because of two changes from AWS. First, the Kubernetes
```
jklukas:
```suggestion
This design is possible because of two changes compared to our AWS implementation. First, the Kubernetes
```

docs/differences_from_aws.md
relud: I updated the header for this column to "Kafka in AWS Data Pipeline" to indicate that this isn't a kafka limitation, it's an "our kafka setup" limitation

docs/differences_from_aws.md
relud: I changed this to "Configured in GB per EC2 instance" because I feel like that cuts to the heart of the issue. PubSub's advantage is that it's an amount of time, not storage.

docs/differences_from_aws.md
relud: I did this instead, because it felt awkward to include a link description in the sentence:

> This document explains how GCP Ingestion differs from the [AWS Data Platform Architecture](https://mana.mozilla.org/wiki/display/SVCOPS/Telemetry+-+Data+Pipeline+Architecture)

added advantage being that link isn't guaranteed to be a diagram, jason just chose to put all the info into the diagram instead of more words.

docs/differences_from_aws.md
relud: part of the problem is that the heka message format and the pubsub message format aren't actually what's being swapped out here. it's heka framed protobuf vs compressed ndjson, and the json schema varies. i'll update this section to reflect that.

docs/differences_from_aws.md
relud: counter-suggestion: 
```suggestion
requests as Heka messages and then write them to Kafka using librdkafka and directly to
```

jklukas: that's fine.

docs/design.md
relud: I updated the wording here to describe PubSub and Kubernetes separately

docs/design.md
relud: I added a diagram describing the color meanings right before this

relud: I put the commit in the mermaid file, and made the mermaid file the alt text of the image

jklukas:
I'm going ahead and merging these docs as they look in great shape to me and they'll be a useful reference immediately. Any further fixups can happen in a new PR.

PR #115, opened 2018-10-22
Format ingestion-edge with black

relud:
https://github.com/ambv/black


.circleci/config.yml
```diff
@@ -18,6 +18,18 @@ jobs:
           name: Ensure markdown tables of contents are up to date
           command: ./.circleci/doctoc-check.sh
 
+  black:
```
jklukas: I'd prefer to call this task `format` to describe what it does rather than referring to the specific tool it uses. Same for `bin/black` -> `bin/format`.

relud: will do

PR #120, opened 2018-10-23
Move test requirements into a separate document

relud:


relud:
cc @jklukas if you want to chime in on anything

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+      - require disk queue is not touched in absence of PubSub errors
+- full recovery from downtime must not take longer than the downtime period
+
+### Recovery Tests
+
+Recovery tests should the following properties in addition to load test
+properties:
+
+- do not need to run in 1 hour
+- do not block deployment to production
```
relud: this is a change to current requirements from before, everything else is more or less copied or rephrased

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+   - present, absent, invalid, and wrong-type payload JSON
+   - present, absent, invalid, and wrong-type payload fields
+   - duplicate documents
+   - ingestion-beam: inputs and outputs return same schema received
+
+### Integration Tests
+
+Integration tests should have the following properties:
+
+- suite runs in less than 1 minute in CI
```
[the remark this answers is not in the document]
jklukas: This context on timing is helpful. Let's put this in the summary at the top of the Test Categories section.

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+      - 100% and 10% of external service requests time out
+      - 100% and 10% external service requests return 500
+   - ingestion-edge:
+      - when running out of disk queue space due to PubSub server errors
+      - require disk queue is not touched in absence of PubSub errors
+- full recovery from downtime must not take longer than the downtime period
+
+### Recovery Tests
+
+Recovery tests should the following properties in addition to load test
```
jklukas:
```suggestion
Recovery tests should have the following properties in addition to load test
```

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+properties:
+
+- do not need to run in 1 hour
```
jklukas: So this means they _may_ run for longer than 1 hour?

jklukas: It might be useful to adopt [RFC requirement levels](https://www.ietf.org/rfc/rfc2119.txt) for this document, so each line begin like "MUST x", "MAY y", "SHOULD NOT z"

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+      - PubSub cannot TCP connect
+      - Cloud Storage cannot TCP connect
+      - Dataflow disk is full
+- clean up external resources when complete
+
+### Load Tests
+
+Load tests should have the following properties:
+
+- can run in parallel in less than 1 hour
```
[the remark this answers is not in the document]
relud: i'm definitely referring to load testing against separate stacks

docs/test_requirements.md
```diff
@@ -0,0 +1,131 @@
...
+   in a GCP environment configured to mimic production configuration as
+   closely as possible.
+
+## Test Categories
+
+### Unit Tests
+
+Unit tests should have the following properties:
+
+- suite runs in less than 5 seconds in CI
```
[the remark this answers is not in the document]
relud: > I think our current Maven tests take at least 15 seconds.

Yes, but they don't all fall under unit tests. I think I can get them running in under 5 seconds.

Also, when this says `in CI` it's meant to include CI's parallelism. If we have 15 seconds of unit tests, but CI splits them into 4 groups that each complete in time, that's sufficient.

I will add a comment indicating this.

relud:
Okay, non-trivial updates to address review, requesting another look

PR #143, opened 2018-10-26
Add an SLA for gcp-ingestion

relud:
Phrasing is largely modeled after https://cloud.google.com/pubsub/sla

Thresholds were made up based on what I think is reasonable to achieve by the end of Q4.

Requesting review from data-platform before soliciting feedback from operations.

### DIFF
--- a/docs/sla.md
+++ b/docs/sla.md
@@ -1,20 +1,38 @@
+# Overview
+
+This document describes a prototype for how Downtime on GCP Ingestion will be
+measured and handled.
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Disclaimer and Purpose](#disclaimer-and-purpose)
+- [Service Level Agreement (SLA)](#service-level-agreement-sla)
+  - [Service Level Objective](#service-level-objective)
+  - [Definitions](#definitions)
+  - [Exclusions](#exclusions)
+  - [Additional Information](#additional-information)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# Disclaimer and Purpose
+
+**This agreement is intended solely for those directly running, writing, and
+managing GCP Ingestion. It is not an agreement, implicit or otherwise, with any
+other parties.** This agreement is a prototype that should be treated as a goal
+that will require adjustments.
+
+The purpose of this agreement is first and foremost to encourage behavior that
+reduces Downtime. The secondary purpose is to establish clear expectations for
+how to respond to Downtime and how much Downtime is considered acceptable.
+
 # Service Level Agreement (SLA)
 
 In production the Covered Service will provide a minimum Biweekly Uptime
 Percentage determined by the Service Level Objective (SLO). If the Covered
 Service does not meet the SLO then a Stability Work Period should be assigned
 to engineers supporting the Covered Service.
-
-<!-- START doctoc generated TOC please keep comment here to allow auto update -->
-<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
-
-
-- [Service Level Objective](#service-level-objective)
-- [Definitions](#definitions)
-- [Exclusions](#exclusions)
-- [Additional Information](#additional-information)
-
-<!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Service Level Objective
 
--- a/docs/sla.md
+++ b/docs/sla.md
@@ -1,0 +1,74 @@
+# Service Level Agreement (SLA)
+
+In production the Covered Service will provide a minimum Biweekly Uptime
+Percentage determined by the Service Level Objective (SLO). If the Covered
+Service does not meet the SLO then a Stability Work Period should be assigned
+to engineers supporting the Covered Service.
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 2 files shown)


relud:
cc @sunahsuh

docs/sla.md
```diff
@@ -0,0 +1,74 @@
...
+exceeds 1 hour for a batch sink job or 1 minute for a decoder or streaming sink
+job.
+
+**"Downtime Period"** means a period of 60 consecutive seconds of Downtime.
+Intermittent Downtime for a period of less than 60 consecutive seconds will not
+be counted towards any Downtime Periods.
+
+**"Biweekly Uptime Percentage"** means total number of minutes in two weeks,
+minus the number of minutes of Downtime suffered from all Downtime Periods in
+two weeks, divided by the total number of minutes in two weeks.
```
jklukas: Is this a rolling two weeks, or are we going to have some defined start date and end date for each biweekly period?

docs/sla.md
```diff
@@ -0,0 +1,74 @@
...
+
+**"Stability Work"** means work on preventing downtime while all other features
+and work are suspended.
+
+**"Stability Work Period"** means a period of time spent on Stability Work
+determined below.
+
+| Covered Service | Biweekly Uptime Percentage | Time spent on Stability Work |
```
jklukas: Is the "time spent" here referring to a team's time? Or to person-weeks of effort? So a promise of "3 months" of time spent could potentially be 3 engineers dedicated for 1 month?

docs/sla.md
relud: rolling period. i'll update this to say that.

docs/sla.md
relud: it's "time spent" as in a dedicated "3 months" window that applies to relevant people. so if it's three engineers, they're still dedicated for 3 months.

relud: also, it may be 3 months for one person, and another person assigned for review.

relud: Also, this is not intended to indicate that time can't be taken to task switch as a fatigue reducing measure. Just that this is the only assigned work for the period, and that it is the only work that should be deployed to production during the period.

relud:
> if we're going to merge this to the repo in the near term, I think there needs to be a disclaimer at the top that the contract here is currently aspirational.

this won't be merged unless operations and management agree to the terms, which I very much doubt will happen anywhere near quickly or without modification.

I think it would be very good to have a highlighted disclaimer indicating that this is a prototype SLA as an agreement between the developers/managers and themselves, and is not an agreement, implied or otherwise, with anyone outside that group. I think it would also be good to include something like a mission statement for the SLA, which in this case is primarily to respond to downtime with behavior that reduces future downtime and secondarily to establish internal expectations regarding downtime. I'll try to write something like that up and add it.

> We need to understand how we'll be measuring all the quantities here

for the server error and message age, I'm confident it will be available via stackdriver. I think that it's something operations will be in charge of measuring, and I chose the language there specifically based on what I thought would be reasonable to evaluate automatically.

I have no idea how we should represent the downtime due to misbehaving code. I chose effecting `at least 0.1%` of traffic, but now that I think about it what I was really trying to express was a discrepancy requiring backfill (as opposed to requiring modification of the spec). Downtime would be measured relative to the period that requires backfill, but that's actually a proxy for measuring how much the issue impacted or could have impacted external parties.

I think that misbehaving code will cause far longer downtime periods. I'm uncertain about how handle how that relates to the "Stability Work Period", because I don't think every beam job bug that goes unnoticed for more than 2 days should realistically result in dropping non-stability quarterly goals. On the other hand I feel like that would be very effective at minimizing back fills due to code here.

> who is accountable for confirming and enforcing the SLA

Stackdriver can monitor the non-code portion of the SLA and send out an email indicating a breach of SLA and the expected response. I expect we should be able to set up an integration between stackdriver and statuspage to make it more visible.

We would have a postmortem as normal, so for enforcement I expect that would include checking in to determine which engineers it would be most effective and appropriate to assign a "Stability Work Period" to minimize future downtime. I also expect this would implicitly include re-evaluating the documented length of the "Stability Work Period" if it is deemed inappropriately long or short.

> we should be extra careful about what we're messaging at this stage

Agreed, thorough and careful communication about this is vital to avoiding unintentional side-effects.

relud:
I said: 

> I have no idea how we should represent the downtime due to misbehaving code. I chose effecting `at least 0.1%` of traffic, but now that I think about it what I was really trying to express was a discrepancy requiring backfill (as opposed to requiring modification of the spec). Downtime would be measured relative to the period that requires backfill, but that's actually a proxy for measuring how much the issue impacted or could have impacted external parties.

I didn't change it though, because I'm having trouble finding wording that is concise and unambiguous.

relud:
@whd I'd like to invite you to review at this point. I'm specifically interested in your personal opinion on the document.

Afterward (still before merging) I'll schedule a meeting with relevant parties to discuss things like manager agreement and how this effects monitoring and alerting.

docs/sla.md
```diff
@@ -0,0 +1,95 @@
...
+| Covered Service | Biweekly Uptime Percentage |
+|-----------------|----------------------------|
+| ingestion-edge  | 99.99%                     |
+| ingestion-beam  | 99.5%                      |
+
+## Definitions
+
+**"Covered Service"** means ingestion-edge and ingestion-beam jobs.
+
+**"Downtime"** on ingestion-edge means for at least 0.1% of requests [the
```
whd: The spec was not "followed" is vague as to what is doing the following, and is potentially vague as to the meaning of "following".

relud:
this has some external issues, i'm going to close this pr for now, and keep the branch around for later use if needed.

PR #146, opened 2018-10-29
Add dockerhub config

whd:
There's some yaml formatting inconsistency WRT indenting lists which I've left as-is. I've also only added filtering to the jobs related to building the edge. I'm open to any restructuring that accomplishes the goal of pushing images to dockerhub on merges to master.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,9 +30,9 @@ jobs:
           name: Ensure ingestion-edge is formatted by black
           command: black --check --diff ingestion-edge
 
-  ingestion-edge:
+  ingestion-edge-build:
     docker:
-    - image: docker/compose:1.22.0
+    - image: docker:18.02.0-ce
     steps:
     - checkout
     - setup_remote_docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -60,6 +87,33 @@ jobs:
           COMPOSE_FILE: ingestion-edge/docker-compose.yml
           COMPOSE_PROJECT_NAME: circleci-ingestion-edge
 
+  ingestion-edge-deploy:
```
relud: I think we should use an early return like [jeff's recent article](https://circleci.com/blog/managing-secrets-when-you-have-pull-requests-from-outside-contributors/), docker-compose to push the image, and not split the jobs because it makes the local cli more troublesome. something like:

in `ingestion-edge/docker-compose.yml`:
<pre>
  web:
    image: &image "${DOCKERHUB_REPO-ingestion-edge}:${CIRCLE_TAG-latest}"
</pre>

in `.circleci/config.yml`
<pre>
    - run:
        name: Early return for no deploy
        command: |
                if [ -n "$CIRCLE_PR_NUMBER" ] || ([ "${CIRCLE_BRANCH}" != "master" ] && [ -z "${CIRCLE_TAG}" ]); then
                  echo "Nothing to deploy"
                  circleci step halt
                fi
    - run:
        name: Deploy to Dockerhub
        command: |
          echo $DOCKER_PASS | docker login -u $DOCKER_USER --password-stdin
          docker-compose push web
</pre>

whd:
@relud I've updated the PR to push in the style you suggest. I am not sure if this style will affect downstream jenkins processing but I'll cross that bridge when I get to it. The ingestion-beam error appears to be unrelated:

    Error: Could not find or load main class org.apache.maven.surefire.booter.ForkedBooter
    ...
    [ERROR] Failed to execute goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test (default-test) on project ingestion-beam: Execution default-test of goal org.apache.maven.plugins:maven-surefire-plugin:2.12.4:test failed: The forked VM terminated without saying properly goodbye. VM crash or System.exit called ? -> [Help 1]

PR #148, opened 2018-10-30
Add tests for check_disk_bytes_free

### LINKED ISSUE #147: Return something when check_disk_bytes_free runs and passes
https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-edge/ingestion_edge/dockerflow.py#L39

Be it the empty list or perhaps something more informational such as the number of bytes free. The current behavior is to fail with `'NoneType' object is not iterable` when both `QUEUE_PATH` and `MINIMUM_DISK_FREE_BYTES` are set and the check passes (it does however fail correctly when it should).

I almost wrote a patch but that would require writing a test :open_mouth:

relud:
fixes #147

### DIFF
--- a/ingestion-edge/tests/test_api.py
+++ b/ingestion-edge/tests/test_api.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, you can obtain one at http://mozilla.org/MPL/2.0/.
 
+from dataclasses import dataclass
 from datetime import datetime
 from dateutil.parser import parse
 from sanic.testing import SanicTestClient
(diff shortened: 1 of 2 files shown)


ingestion-edge/tests/test_api.py
```diff
@@ -11,14 +12,85 @@
...
+
+
+def test_heartbeat_noop(client: SanicTestClient):
+    def check(app):
+        assert app is client.app
+        return []
+
+    with patch("ingestion_edge.dockerflow.check_disk_bytes_free", new=check):
+        _, r = client.get("/__heartbeat__")
+        assert r.json == {
```
jklukas: I generally try to leave a context manager as soon as possible, similar to keeping the body of a `try` as short as possible. So, I would have this assert indented left here and in the other tests.

I don't evidence that this is a standard followed elsewhere, so just something to think about.

relud: i don't see how it would hurt, and i know reducing the size of try blocks is recommended

PR #152, opened 2018-11-01
Add a brief description of the migration plan.

mreid-moz:
Credit to :whd for the content, any errors are courtesy of :mreid.

### DIFF
--- a/docs/edge_migration.md
+++ b/docs/edge_migration.md
@@ -2,6 +2,19 @@
 
 This document outlines plans to migrate edge traffic from AWS to GCP using the code in this repository.
 
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+
+- [Current state](#current-state)
+- [Phase 1](#phase-1)
+- [Phase 2](#phase-2)
+- [Phase 3](#phase-3)
+- [Phase 3 (alternative)](#phase-3-alternative)
+- [Phase 4](#phase-4)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
 ## Current state
 
 Today, data producers send data to the ingestion stack on AWS as described [here](https://github.com/mozilla/firefox-data-docs/blob/042fddcbf27aa5993ee5578224200a3ef65fd7c7/src/concepts/pipeline/data_pipeline_detail.md#ingestion).
--- a/docs/edge_migration.md
+++ b/docs/edge_migration.md
@@ -20,7 +20,7 @@ This "expected" data does _not_ include recording and replaying traffic that we 
 
 During this phase, we will collect logging info to determine exactly what data is being ignored by limiting to this "expected" data.
 
-Why have a GCP PubSub topic instead of running an HTTP output from the AWS consumer directly? The main reason is that we want 100% of data in a PubSub topic anyway, for staging purposes. This way we can have a production GCP Edge ingestion stack writing to production GCP resources, while being able to e.g. simulate load and various outage conditions using the same data in a staging environment. We could could in theory do the stage testing using the prod GCP edge output topic, assuming the edge has no issues. This will be the eventual end state of the system when there's no more AWS, but we are currently reusing most of the hindsight tooling for stage testing since it's already written and working in production.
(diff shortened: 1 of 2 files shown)


docs/edge_migration.md
```diff
@@ -0,0 +1,73 @@
+# Plans for migrating edge traffic to GCP Ingestion
+
+This document outlines plans to migrate edge traffic to go through the code in this repository.
```
whd: Nit: migration happens from one place to another. I'd suggest "migrate edge traffic from AWS to GCP" or similar.

docs/edge_migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Current state
+
+Today, data producers send to a data ingestion stack on AWS as described [here](https://github.com/mozilla/firefox-data-docs/blob/042fddcbf27aa5993ee5578224200a3ef65fd7c7/src/concepts/pipeline/data_pipeline_detail.md#ingestion).
```
whd: Nit: while somewhat redundant, leaving the indirect object unspecified makes this statement ambiguous as to what is being sent. I would suggest "data producers send data to a ..." or "data producers produce data to a ...".

docs/edge_migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+Timeline: **Q4 2018**
+
+```
+Data Producers -> AWS Edge -> Kafka -> AWS Hindsight -> PubSub -> GCP Hindsight w/ HTTP Output -> GCP Edge
+```
+
+In this configuration we get 100% of "expected" data going to the GCP pipeline without any risk of affecting production AWS processing.
+
+This "expected" data does _not_ include recording and replaying traffic that we currently throw away at the edge (e.g. pre-unified telemetry). There may be some other subtle differences from the original incoming data from producers, such as missing some headers that are not currently being stored in Landfill. On the whole, this a good enough approximation of 100% of the data we care actually about, but for the purposes of testing the new system, we will need to care about operation while processing data we don't care about, see more detail in Phase 3 below.
```
whd: There are multiple errors here. I know I wrote this and I'm correcting myself, but now that this is going into actual documentation I would suggest:

    On the whole, this is a good approximation of 100% of the data we actually care about. We also need to test operation of the new system while processing data we don't care about; see more detail in Phase 3 below.

docs/edge_migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Phase 4
+
+Timeline: **TBD**
+
+```
+Data Producers -> GCP Edge
+```
+
+This will happen when we are confident that there is no risk of data loss by switching the endpoint to the GCP stack.
```
whd: I'm not sure this is an accurate characterization of why Phase 4 isn't going to happen for a while. I think we will be reasonably certain that we won't be losing data on the GCP infra long before we've fully migrated off all the AWS-based infrastructure and data processing, but I might be wrong.

whd:
I now notice that doctoc and spelling are failing CI. Those should probably be fixed before merging this.

docs/edge_migration.md
mreid-moz: My intention here was "this won't happen *until* we are confident bla bla". It may be significantly later, depending on the realities of sunsetting things in AWS. I'll clarify.

docs/edge_migration.md
```diff
@@ -0,0 +1,73 @@
...
+
+During this phase, we will collect logging info to determine exactly what data is being ignored by limiting to this "expected" data.
+
+Why have a GCP PubSub topic instead of running an HTTP output from the AWS consumer directly? The main reason is that we want 100% of data in a PubSub topic anyway, for staging purposes. This way we can have a production GCP Edge ingestion stack writing to production GCP resources, while being able to e.g. simulate load and various outage conditions using the same data in a staging environment. We could could in theory do the stage testing using the prod GCP edge output topic, assuming the edge has no issues. This will be the eventual end state of the system when there's no more AWS, but we are currently reusing most of the hindsight tooling for stage testing since it's already written and working in production.
```
akkomar: typo: `We could could`

```suggestion
Why have a GCP PubSub topic instead of running an HTTP output from the AWS consumer directly? The main reason is that we want 100% of data in a PubSub topic anyway, for staging purposes. This way we can have a production GCP Edge ingestion stack writing to production GCP resources, while being able to e.g. simulate load and various outage conditions using the same data in a staging environment. We could in theory do the stage testing using the prod GCP edge output topic, assuming the edge has no issues. This will be the eventual end state of the system when there's no more AWS, but we are currently reusing most of the hindsight tooling for stage testing since it's already written and working in production.
```

jklukas:
I ran `./bin/update-toc` and committed the results. Good to merge now?

mreid-moz:
Thanks Jeff! Yep, I think that was the last thing.

PR #153, opened 2018-11-06
Optionally discard geo_city based on presence in a city size file

### LINKED ISSUE #144: Optionally discard geo_city based on presence in a city size file
https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_ingest/io_modules/decoders/moz_ingest/common.lua#L26-L28

jklukas:
Closes #144

I realized a slightly improved way of accessing a single file via the
Beam FileSystems API and decided it's simple enough to use directly rather
than wrapping up in a utility class, so `util/File.java` goes away here.

The sample city file for `milton` is taken from [lua_sandbox_extensions]
(https://github.com/mozilla-services/lua_sandbox_extensions/blob/e9a472704a88cf7f714ba4974436701558850c44/moz_ingest/tests/cities1000-test.txt)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -81,14 +100,20 @@ public PubsubMessage apply(PubsubMessage value) {
           foundCity.inc();
 
           attributes.put("geo_country", response.getCountry().getIsoCode());
-          attributes.put("geo_city", response.getCity().getName());
+
+          City city = response.getCity();
+          if (cityAllowed(city.getGeoNameId())) {
+            attributes.put("geo_city", response.getCity().getName());
```
relud:
```suggestion
            attributes.put("geo_city", city.getName());
```

relud:
>I realized a slightly improved way of accessing a single file via the
Beam FileSystems API and decided it's simple enough to use directly rather
than wrapping up in a utility class, so util/File.java goes away here.

awesome!

PR #156, opened 2018-11-06
Replace CompositeTranform with Beam's PTranform.compose

### LINKED ISSUE #155: Remove CompositeTransform
We now have `PTransform.compose` in the Beam API, so can replace all usage.

jklukas:
Closes #155


.circleci/config.yml
```diff
@@ -95,7 +95,7 @@ jobs:
         command: cd ingestion-beam && mvn clean test
     - run:
         name: Report code coverage
-        command: bash <(curl -s https://codecov.io/bash) -F ingestion-beam
+        command: bash <(curl -s https://codecov.io/bash) -F beam
```
relud: why?

jklukas: Apparently, dashes are not allowed:

https://docs.codecov.io/docs/flags#section-how-to-create-flags

jklukas: codecov is currently ignoring this upload due to the bad flag name

PR #157, opened 2018-11-06
Report code coverage for ingestion-edge

### LINKED ISSUE #123: Setup code coverage reporting for ingestion-edge

jklukas:
Closes #123

### DIFF
--- a/ingestion-edge/docker-compose.yml
+++ b/ingestion-edge/docker-compose.yml
@@ -38,12 +38,13 @@ services:
       ]
   test:
     image: *image
-    entrypoint: py.test
+    entrypoint: pytest
     command:
     - --flake8
     - --docstyle
     - --mypy
     - --mypy-ignore-missing-imports
+    - --cov=./
     depends_on:
     - pubsub
     - web
--- a/ingestion-edge/docker/pytest-upload-coverage
+++ b/ingestion-edge/docker/pytest-upload-coverage
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Wrapper for pytest that also submits coverage data.
+
+set -eox pipefail
+
+pytest "$@"
+bash <(curl -s https://codecov.io/bash) -F ingestion_edge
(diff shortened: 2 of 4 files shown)


ingestion-edge/docker-compose.yml
```diff
@@ -38,15 +38,12 @@ services:
...
-    - --flake8
-    - --docstyle
-    - --mypy
-    - --mypy-ignore-missing-imports
+    entrypoint: bin/test
     depends_on:
     - pubsub
     - web
+    env_file:
+    - docker-compose.test.env
```
relud: I think we should leave this as part of the circle config, or make it `env: {CIRCLE: "${CIRCLE}"}` so that this file doesn't get applied when locally running `docker-compose` after locally running `circleci build --job ingestion-edge`

ingestion-edge/bin/test
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -eox pipefail
+
+py.test \
+    --flake8 \
+    --docstyle \
+    --mypy \
+    --mypy-ignore-missing-imports \
+    --cov=./
```
relud: I think we should leave these args in docker-compose, and here use `"$@"` so that `docker-compose run --no-deps --rm test tests` continues to work as expected

ingestion-edge/docker-compose.yml
jklukas: Refactored to have a docker-compose.circleci.env that contains all the changes specific to running in CI

ingestion-edge/bin/test
jklukas: Done.

ingestion-edge/docker/pytest-upload-coverage
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Wrapper for pytest that also submits coverage data.
+
+set -eox pipefail
+
+pytest $@
```
relud:
```suggestion
pytest "$@"
```

relud: this will make bash expansion preserve argument groups, in case they contain spaces.

PR #163, opened 2018-11-06
Make ingestion-edge metadata headers configurable

### LINKED ISSUE #13: Make edge server's metadata headers configurable

relud:
fixes #13

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -53,7 +53,7 @@ jobs:
           docker-compose
           up
           --build
-          --exit-code-from test
+          --abort-on-container-exit
           --force-recreate
           --quiet-pull
         environment:
--- a/ingestion-edge/ingestion_edge/config.py
+++ b/ingestion-edge/ingestion_edge/config.py
@@ -49,3 +49,13 @@ ROUTE_TABLE = [Route(*route) for route in json.loads(environ.get("ROUTE_TABLE", 
 QUEUE_PATH = environ.get("QUEUE_PATH", "queue")
 
 MINIMUM_DISK_FREE_BYTES = int(environ.get("MINIMUM_DISK_FREE_BYTES", 0)) or None
+
+METADATA_HEADERS = {
+    header: header.replace("-", "_")
+    for raw in environ.get(
+        "METADATA_HEADERS",
+        "content-length,dnt,date,user-agent,x-forwarded-for,x-pingsender-version",
+    ).split(",")
+    for header in (raw.strip().lower(),)
+    if header
+}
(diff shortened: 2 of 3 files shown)


.circleci/config.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           docker-compose
           up
           --build
-          --exit-code-from test
+          --abort-on-container-exit
```
relud: i had something cause `web` to exit before `test` could be started and it returned 0 because test wasn't run >_<. this flag is a subset of `--exit-code-from=<container>` that appears to use the exit code of the container that caused the abort, which is what we want.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -123,13 +127,11 @@
...
-            content_length=request.headers.get("Content-Length"),
-            date=request.headers.get("Date"),
-            dnt=request.headers.get("DNT"),
             host=request.host,
-            user_agent=request.headers.get("User-Agent"),
-            x_forwarded_for=request.headers.get("X-Forwarded-For"),
-            x_pingsender_version=request.headers.get("X-Pingsender-Version"),
+            **{
+                attr: request.headers.get(header)
+                for header, attr in metadata_headers.items()
```
relud: i chose to require `header` and `attr` separately so the server doesn't need to execute `header.replace("-","_")` all the time

we could instead modify the decoder to use dashes or modify `config.py` to generate the map from a list, this was just the easiest way to implement it.

ingestion-edge/ingestion_edge/config.py
```diff
@@ -49,3 +49,17 @@ class Route:
...
+            "METADATA_HEADERS",
+            '{"Content-Length":"content_length",'
+            '"DNT":"dnt",'
+            '"Date":"date",'
+            '"User-Agent":"user_agent",'
+            '"X-Forwarded-For":"x_forwarded_for",'
+            '"X-Pingsender-Version":"x_pingsender_version"}',
+        )
+    )
+)
```
jklukas: I'd prefer a triple-quoted string here to make this more readable. Perhaps ideal would be breaking out the default:

```
METADATA_HEADERS_DEFAULT = '''
{
  "Content-Length": "content_length",
  ...
}
'''

METADATA_HEADERS = dict(json.loads(environ.get("METADATA_HEADERS", METADATA_HEADERS_DEFAULT)))
```

ingestion-edge/ingestion_edge/publish.py
jklukas: If we know we always want the attr name to follow this format, then I like the idea of generating this map in `config.py`. Especially if that lets us change the `METADATA_HEADERS` format to be a comma-separated string rather than a full json blob.

ingestion-edge/ingestion_edge/config.py
relud: i used a comma-separated string instead, and i feel like that addresses the readability. lmk if you still think it should be separated.

jklukas: That looks great.

PR #166, opened 2018-11-08
Fix container termination behavior for SIGTERM

### LINKED ISSUE #164: Fix container termination behavior via SIGTERM
https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-edge/Dockerfile#L9

It looks like `CMD_PREFIX` was borrowed from Normandy[1][2]. I don't see where it is used over there, but I didn't look carefully.

At any rate, on GKE when deleting pods (particularly when increasing terminationGracePeriodSeconds by a significant amount) I noticed that the edge app container does not respond to the TERM signal it is being sent. See https://stackoverflow.com/questions/43584760/gunicorn-graceful-stopping-with-docker-compose for what I believe is the exact same scenario. The reason we don't see this timeout in docker-compose is because [we run python directly](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-edge/docker-compose.override.yml#L8) in that configuration.

There are a couple of ways to fix this. There's a bunch of information about the various forms of `CMD` and `ENTRYPOINT` [3][4][5][6]. The simplest fix is to prefix the current shell-style `CMD` with exec. This is presumably what `CMD_PREFIX` might have been for, and I've confirmed that setting `CMD_PREFIX` to "exec" in the environment does in fact fix the container shutdown behavior. I think unless we have a compelling reason to keep `CMD_PREFIX` around we should just drop it and prefix `CMD` with exec. However, it's also recommended not to use the shell forms for `CMD` and `ENTRYPOINT`[6] and this round of debugging has made me realize why that might be, so an alternative path forward would be to migrate to exec-style syntax.

For reference, Normandy uses an exec-style `ENTRYPOINT` (and `CMD`), but has a wrapper script to do (among other things) environment variable interpolation. It might be good to adopt a similar approach.

[1] https://github.com/mozilla/normandy/blob/master/Dockerfile#L34
[2] https://github.com/mozilla/normandy/blob/master/bin/run.sh#L10
[3] https://docs.docker.com/engine/reference/builder/#understand-how-cmd-and-entrypoint-interact
[4] https://docs.docker.com/engine/reference/builder/#entrypoint
[5] http://goinbigdata.com/docker-run-vs-cmd-vs-entrypoint/
[6] https://docs.docker.com/engine/reference/builder/#cmd

relud:
fixes #164


relud:
wow, exec does not work at all how i thought it did v_v. Fixed the PR

PR #167, opened 2018-11-08
Move Decoder tests into transform-specific suites

### LINKED ISSUE #94: Move ingestion-beam decode tests
current decode tests are against the transforms, so they should be separated based on the main file they are testing

jklukas:
We remove `DecoderTest` completely, factoring tests out to individual
classes corresponding to the transform they test.
Most test method names have been changed to `testExpand`
(since the class name now expresses which transform is under test).

Otherwise, the content of the test code is intended to be identical.

Closes #94


relud:
or `testOutput`, which is the best i've been able to think of

jklukas:
testOutput sounds fine. Will change.

PR #190, opened 2018-11-14
Add BigQueryIntegrationTest that accesses BigQuery in GCP

### LINKED ISSUE #95: Test bigquery output in ingestion-beam
bq doesn't have an emulator, so i think this will have two parts.

1) use a fake http service and hit localhost. may not work.
2) use cloud-build on restricted branches and test actual bq output. needs to generate unique table names and reliably clean up all resources after running.

jklukas:
Closes #95

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -10,10 +10,15 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <beam.version>2.8.0</beam.version>
         <exec.mainClass>com.mozilla.telemetry.Sink</exec.mainClass>
         <!-- Workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925 -->
         <argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
+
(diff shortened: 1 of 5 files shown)


ingestion-beam/pom.xml
```diff
@@ -10,10 +10,15 @@
...
-        <beam.version>2.8.0</beam.version>
         <exec.mainClass>com.mozilla.telemetry.Sink</exec.mainClass>
         <!-- Workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925 -->
         <argLine>-Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
+
+        <!-- Keep these dependency versions in sync with those pulled in by beam;
+             check https://mvnrepository.com/artifact/org.apache.beam -->
+        <beam.version>2.8.0</beam.version>
+        <google-cloud.version>1.32.0</google-cloud.version>
+        <jackson.version>2.9.5</jackson.version>
```
jklukas: This should also prevent dependabot from trying to update these. I think it's not able to parse back through the logic of versions declared via property references.

jklukas:
I am quite pleased with the setup and teardown time for BigQuery. The test runs in a few seconds.

jklukas:
Broke out additional testing need to #191

PR #215, opened 2018-11-19
Test BigQuery failedInserts

### LINKED ISSUE #191: Test BigQuery failed inserts

jklukas:
The test here exercises failedInserts as currently implemented,
but the implementation still leaves some things to be desired:

- we lose attributes for failedInserts (because we get back only the
  TableRow; not the original PubsubMessage)
- we rely on the less performant BigQueryIO.insertTableRows method
  rather than passing a function to convert PubsubMessage to TableRow,
  since there's no hook for error handling in the conversion function

I will write a follow-up issue for these problems, but they look like
they'd require upstream Beam changes to address.

Closes #191


ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -161,6 +162,8 @@
               BigQueryIO.writeTableRows().withMethod(BigQueryIO.Write.Method.STREAMING_INSERTS)
                   .withCreateDisposition(BigQueryIO.Write.CreateDisposition.CREATE_NEVER)
                   .withWriteDisposition(BigQueryIO.Write.WriteDisposition.WRITE_APPEND)
+                  .skipInvalidRows()
+                  .withFailedInsertRetryPolicy(InsertRetryPolicy.retryTransientErrors())
```
jklukas: Without setting this retry policy, any failed insert bubbled up to a pipeline execution exception rather than getting captured in getFailedInserts. I asked about this on the beam user list since it doesn't seem to be documented behavior.

PR #219, opened 2018-11-20
Add PubsubIntegrationTest to ingestion-beam

### LINKED ISSUE #207: PubSub integration tests for ingestion-beam
We should exercise both PubSub input type and PubSub output type. We can use the Pubsub emulator provided by Google.

jklukas:
Adds one test case each for exercising InputType.pubsub and OutputType.pubsub

Also introduces `numShards` options, since sharding specification is required when an unbounded source is involved.

Closes #207

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/PubsubIntegrationTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/PubsubIntegrationTest.java
@@ -1,0 +1,193 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry;
+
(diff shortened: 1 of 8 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/PubsubIntegrationTest.java
```diff
@@ -0,0 +1,193 @@
...
+    PAssert.that(output).containsInAnyOrder(inputLines);
+
+    // This runs in the background and returns immediately due to setBlockOnRun above.
+    PipelineResult result = pipeline.run();
+
+    // The wait here is determined empirically; it's not entirely clear why it takes this long.
+    System.err.println("Waiting 15 seconds to make sure we've processed all messages...");
+    result.waitUntilFinish(Duration.millis(15000));
+    System.err.println("Done waiting; now cancelling the pipeline so the test can finish.");
+    result.cancel();
```
relud: looks good for now, filed #220 to iterate when we can

Issue #208, opened 2018-11-16
Exception tests for transforms


jklukas:
Write tests that exercise each of the understood cases within our PTransform classes where an exception could be raised.

PR #225, opened 2018-11-21
Test schema failure modes in ValidateSchema

jklukas:
Partially fulfills #208


ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -56,7 +56,7 @@ public static JSONObject readJSONObject(byte[] data) throws IOException {
     // Read data into a tree
     TreeNode tree = MAPPER.readTree(data);
     // Check that we have an object, because treeToValue won't
-    if (!tree.isObject()) {
+    if (tree == null || !tree.isObject()) {
```
jklukas: `tree.isObject` was throwing a NullPointerException in the case of an empty byte array.

PR #227, opened 2018-11-21
Increase ingestion-beam coverage

jklukas:
Various new test cases and some dead code removal.

Partially addresses #208

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputFileFormat.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputFileFormat.java
@@ -17,6 +17,9 @@ public enum OutputFileFormat {
 
     /** Return this PubsubMessage payload as text. */
     public String encodeSingleMessage(PubsubMessage message) {
+      if (message == null || message.getPayload() == null) {
+        return "";
+      }
       return new String(message.getPayload());
     }
 
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
@@ -100,13 +100,7 @@ public class DynamicPathTemplate implements Serializable {
       String name = placeholderNames.get(i);
       String value = attributes.get(name);
       if (value == null) {
-        String defaultValue = placeholderDefaults.get(i);
-        if (defaultValue == null) {
-          throw new IllegalArgumentException("The destination specified via --output"
-              + " includes an attribute placeholder `" + name + "' with no default,"
-              + " so a value is required, but this message has no attribute `" + name + "'.");
-        }
-        value = defaultValue;
+        value = placeholderDefaults.get(i);
       }
       placeholderValues.add(value);
     }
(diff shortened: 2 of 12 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputFileFormat.java
```diff
@@ -17,6 +17,9 @@
 
     /** Return this PubsubMessage payload as text. */
     public String encodeSingleMessage(PubsubMessage message) {
+      if (message == null || message.getPayload() == null) {
+        return "";
+      }
```
jklukas: This is a behavioral change, but seems like a reasonable choice for the scenario?

ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
```diff
@@ -100,13 +100,7 @@ public String replaceDynamicPart(List<String> values) {
       String name = placeholderNames.get(i);
       String value = attributes.get(name);
       if (value == null) {
-        String defaultValue = placeholderDefaults.get(i);
-        if (defaultValue == null) {
-          throw new IllegalArgumentException("The destination specified via --output"
-              + " includes an attribute placeholder `" + name + "' with no default,"
-              + " so a value is required, but this message has no attribute `" + name + "'.");
-        }
-        value = defaultValue;
```
jklukas: Since changing to requiring defaults, this situation can never occur, so this was dead code.

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
```diff
@@ -15,7 +15,7 @@
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
-public class DecoderMainTest {
+public class DecoderMainTest extends Decoder {
```
jklukas: In a few places, I make a test class extend a class whose constructor otherwise would never get called, and thus would show up as uncovered.

jklukas:
codecov.io is down at the moment, causing test failures. We can rerun tests once they're back up.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputFileFormat.java
relud: fine for whole message null, but payload null should be able to encode, because it can exist and can have attributes, such as an error message.

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
relud: wouldn't it be better to have a test that calls the constructor?

ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputFileFormat.java
jklukas: Agreed that it's possible to have a message with a null payload. This particular bit of code is for `text` output where we are throwing away attributes and just encoding the payload as a String. If we don't special-case the null payload here, the next line calls `new String(null)` which throws NPE.

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
jklukas: Yeah, that would be more explicit. Will do.

PR #233, opened 2018-11-27
Decoder mixed errors test

### LINKED ISSUE #209: Single integration test exercising multiple message edge cases

jklukas:
Writing this test uncovered a few bugs and some additional testing nuance, so there are several things going on. I've split some of those changes out in separate commits, but the final commit that adds the test is still rather large and contains some additional changes. I'll add inline comments to explain more of the supporting changes.

Closes #209


ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
```diff
@@ -34,42 +37,46 @@ public static void main(String[] args) {
 
     final DecoderOptions.Parsed options = DecoderOptions.parseDecoderOptions(
         PipelineOptionsFactory.fromArgs(args).withValidation().as(DecoderOptions.class));
-
     final Pipeline pipeline = Pipeline.create(options);
-    final PTransform<PCollection<PubsubMessage>, ? extends POutput> errorOutput = options
-        .getErrorOutputType().write(options);
+    final List<PCollection<PubsubMessage>> errorCollections = new ArrayList<>();
```
jklukas: One of the bugs discovered here is that we were creating multiple transforms for each step's error output, but the configurations are identical, so they create potentially identical filenames and overwrite each other. We now collect error collections in a list and combine to one error output transform.

ingestion-beam/src/main/java/com/mozilla/telemetry/Sink.java
```diff
@@ -27,17 +30,21 @@ public static void main(String[] args) {
 
     final SinkOptions.Parsed options = SinkOptions.parseSinkOptions(
         PipelineOptionsFactory.fromArgs(args).withValidation().as(SinkOptions.class));
-
     final Pipeline pipeline = Pipeline.create(options);
-    final PTransform<PCollection<PubsubMessage>, ? extends POutput> errorOutput = options
-        .getErrorOutputType().write(options);
+    final List<PCollection<PubsubMessage>> errorCollections = new ArrayList<>();
```
jklukas: Same error output change in Sink as in Decoder.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -55,9 +55,9 @@
 
   @Description("URI of a redis server that will be used for deduplication")
   @Validation.Required
-  ValueProvider<URI> getRedisUri();
+  ValueProvider<String> getRedisUri();
```
jklukas: It looks like Beam was passing this to Jackson; Jackson was expecting a JSON blob that it could deserialize into a URI, which isn't what we want.

So I changed this to String and we now have a separate parsing step.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/ErrorOutputType.java
```diff
@@ -5,19 +5,24 @@
...
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder;
+import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.POutput;
+import org.apache.beam.sdk.values.TypeDescriptor;
 
 /**
- * Enumeration of error output types that each provide a {@code write} method.
+ * Enumeration of error output types that each provide a {@code writeFailures} method.
```
jklukas: We now have a concrete `write` method that calls the abstract `writeFailures` method. This allows us to consolidate logic for the new `includeStackTraces` configuration option.

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
```diff
@@ -20,25 +27,39 @@
...
   public TemporaryFolder outputFolder = new TemporaryFolder();
 
-  private String outputPath;
-
-  @Before
-  public void initialize() {
-    outputPath = outputFolder.getRoot().getAbsolutePath();
-  }
+  @Rule
+  public final RedisServer redis = new RedisServer();
```
jklukas: Since RedisServer is now needed in two places, I implemented a junit `Rule` for it that gets used here and in DecoderTest

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
```diff
@@ -20,25 +27,39 @@
...
+  }
+
+  /** Reset the ObjectMapper configurations we changed. */
+  @AfterClass
+  public static void tearDown() throws Exception {
+    Field mapperField = Json.class.getDeclaredField("MAPPER");
+    mapperField.setAccessible(true);
+    ObjectMapper mapper = (ObjectMapper) mapperField.get(null);
+    mapper.disable(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS);
+  }
```
jklukas: This reflection business is a bit unfortunate, but I think it's good to keep the mapper private. In the future, it would be nice to build more robust testing utilities that can compare PubsubMessages rather than strings and could avoid this need.

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderTest.java
```diff
@@ -1,323 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
jklukas: This file was supposed to be deleted in a previous PR where I broke this into separate classes. I somehow screwed that up and this file got included again before merging. Cleaning that up here.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/DeduplicateTest.java
```diff
@@ -23,87 +24,78 @@
...
     Integer ttlSeconds = parsedOptions.getDeduplicateExpireSeconds().get();
-
-    int redisPort = new redis.embedded.ports.EphemeralPortProvider().next();
-    // create testing redis server
-    final RedisServer redis = RedisServer.builder().port(redisPort).setting("bind 127.0.0.1")
-        .build();
-    final URI redisUri = URI.create("redis://localhost:" + redisPort);
-    redis.start();
-
-    try {
```
jklukas: Changes in this block are just indentation because we no longer need the try/catch with the RedisServer rule

ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
jklukas: Wrote up https://github.com/mozilla/gcp-ingestion/issues/234

PR #236, opened 2018-11-27
Add StorageIntegationTest for beam GCS -> GCS pipeline

### LINKED ISSUE #210: Decoder GCS -> GCS integration test
We should:

- produce a non-trivial input file on GCS with raw ping inputs
- produce the expected output file on GCS
- write a test that validates expected and actual outputs match

jklukas:
Closes #210 

Note two preliminary commits:

- move all integration tests into a subpackage
- remove one unneeded option from DecoderMainTest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,14 +83,14 @@ jobs:
     - image: maven
     steps:
     - checkout
-    - restore_cache:
+    - restore_cache: &restore_cache_beam
         keys:
           # when lock file changes, use increasingly general patterns to restore cache
           # https://circleci.com/docs/2.0/caching/#maven-java-and-leiningen-clojure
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -116,12 +116,16 @@ jobs:
...
           fi
     - checkout
+    - restore_cache:
+        <<: *restore_cache_beam
     - restore_cache:
         keys:
           # Use same cache as the ingestion-beam job.
           - maven-repo-v1-{{ .Branch }}-{{ checksum "ingestion-beam/pom.xml" }}
           - maven-repo-v1-{{ .Branch }}-
           - maven-repo-v1-
```
relud: looks like the duplicate should be removed?

jklukas: indeed

PR #237, opened 2018-11-28
Add ResultWithErrors to refactor handling of error collections

### LINKED ISSUE #235: Refactor MapElementsWithErrors to return custom object
We currently have `MapElementsWithErrors` return a `PCollectionTuple`. This works, but requires us to extract specific collections based on tags. It would be cleaner to define a more specific result type something like:

```
@AutoValue
public abstract static class Result<T> extends PInput, POutput {
  public PCollection<T> successes();
  public PCollection<PubsubMessage> failures();

  public Result<T> appendFailureCollectionTo(List<PCollection<PubsubMessage> failureCollections) {
  failureCollections.add(failures());
  }
}
```

Then our main class code could go from:

```
    pipeline.apply("input", options.getInputType().read(options))
        .apply("collect input parsing errors", PTransform.compose((PCollectionTuple input) -> {
          errorCollections.add(input.get(ToPubsubMessageFrom.errorTag));
          return input.get(ToPubsubMessageFrom.mainTag);
        }))
    ...
```

to:

```
    pipeline.apply("input", options.getInputType().read(options))
        .appendFailureCollectionTo(errorCollections)
        .successes()
    ...
```

jklukas:
Closes #235

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
@@ -12,7 +12,6 @@ import com.mozilla.telemetry.decoder.GzipDecompress;
 import com.mozilla.telemetry.decoder.ParsePayload;
 import com.mozilla.telemetry.decoder.ParseUri;
 import com.mozilla.telemetry.decoder.ParseUserAgent;
-import com.mozilla.telemetry.transforms.DecodePubsubMessages;
 import java.util.ArrayList;
 import java.util.List;
 import org.apache.beam.sdk.Pipeline;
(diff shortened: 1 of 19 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
```diff
@@ -32,48 +30,39 @@
...
-        })).apply("parseUri", new ParseUri())
-        .apply("collect parseUri errors", PTransform.compose((PCollectionTuple input) -> {
-          errorCollections.add(input.get(ParseUri.errorTag));
-          return input.get(ParseUri.mainTag);
-        })).apply("validateSchema", new ValidateSchema())
-        .apply("collect validateSchema errors", PTransform.compose((PCollectionTuple input) -> {
-          errorCollections.add(input.get(ValidateSchema.errorTag));
-          return input.get(ValidateSchema.mainTag);
-        })).apply("decompress", new GzipDecompress())
+    // Trailing comments are used below to prevent rewrapping by google-java-format.
```
jklukas: I'm open to pushback about whether using `//` to control google-java-format behavior is a road we're willing to go down. I've been finding this Decoder class more difficult to read than it needs to be, though.

relud: using `//` lgtm, and i definitely agree that it's much more readable

PR #252, opened 2018-11-30
Document need to specify sharding for streaming mode

### LINKED ISSUE #245: Add comment to beam readme about when --outputNumShards is required

jklukas:
Closes #245

### DIFF
--- a/ingestion-beam/README.md
+++ b/ingestion-beam/README.md
@@ -22,6 +22,7 @@ Apache Beam jobs.
     - [Locally](#locally)
     - [On Dataflow](#on-dataflow)
     - [On Dataflow with templates](#on-dataflow-with-templates)
+    - [In streaming mode](#in-streaming-mode)
 - [Decoder Job](#decoder-job)
   - [Transforms](#transforms)
     - [Parse URI](#parse-uri)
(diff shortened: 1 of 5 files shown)


ingestion-beam/README.md
```diff
@@ -303,6 +304,28 @@ gcloud dataflow jobs show "$JOB_ID"
...
+```
+
+When interacting with an unbounded source or sink like `pubsub`, Beam
+will implicitly set itself to streaming execution mode, which imposes some
+additional constraints on configuration of the underlying `FileIO.Write` transform.
+As discussed in the
+[Beam documentation for `FileIO.Write#withNumShards`](https://beam.apache.org/releases/javadoc/2.8.0/org/apache/beam/sdk/io/FileIO.Write.html#withNumShards-int-),
+batch mode is most efficient when the runner is left to determine sharding,
+so `numShards` options should normally be left to their default of `0`, but
+streaming mode can't perform the same optimizations and an exception will be thrown
```
whd: nit: thrown when?

jklukas:
You make a good point about validating the condition in code. I'll look for the condition and raise a custom exception.

jklukas:
Picked nits and added validation in the options parsing code.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -147,6 +147,28 @@ static Parsed parseSinkOptions(SinkOptions options) {
...
+          + " --outputType=file and the input is unbounded (--inputType=pubsub);"
+          + " Dataflow recommends starting with twice the value of --maxWorkers or 10"
+          + " (see https://github.com/apache/beam/pull/1952)");
+    }
+    if (options.getInputType() == InputType.pubsub
+        && options.getErrorOutputType() == ErrorOutputType.file
+        && options.getErrorOutputNumShards() == 0) {
+      throw new IllegalArgumentException("Missing required parameter: "
+          + " --errorOutputNumShards must be set to an explicit non-zero value when"
+          + " --errorOutputType=file and the input is unbounded (--inputType=pubsub);"
```
whd: There's a bunch of mostly duplicated string here but it's likely not worth the effort to dedupe it.

jklukas:
> looks like it's only going to tell you you're missing the first one

Good usability point. I just squashed in a change to accumulate multiple validation errors and include them all in the message.

PR #257, opened 2018-12-05
Document reliability goals

relud:


docs/reliability.md
```diff
@@ -0,0 +1,84 @@
...
+
+| Component      | Biweekly Uptime Percentage |
+|----------------|----------------------------|
+| ingestion-edge | 99.99%                     |
+| ingestion-beam | 99.5%                      |
+
+## Definitions
+
+**"Downtime"** on ingestion-edge means for at least 0.1% of requests a status
+code lower than 500 was not successfully returned. On ingestion-beam it means
```
mreid-moz: Do we yet know how this will be measured and monitored? If so, let's add some detail / links to monitoring here. If not, let's file a follow up issue to figure that piece out.

relud: filed #296

PR #259, opened 2018-12-06
Add python pubsub emulator

relud:
this should speed up ingestion-edge CI because we no longer need to download the `mozilla/gcloud-sdk-emulators` image, and it also repurposes the `UpdateTopic` API to configure test conditions needed for integration tests.


jklukas:
I don't see any difference in test timing with this change. The `ingestion-edge` job still takes ~2 minutes. I'm not wrapping my head around the UpdateTopic configuration, though, and how that's necessary for integration tests.

ingestion-edge/pubsub_emulator/__init__.py
```diff
@@ -0,0 +1,169 @@
...
+        else:
+            for ack_id in request.ack_ids:
+                try:
+                    subscription.pulled.pop(ack_id)
+                except KeyError:
+                    context.set_code(grpc.StatusCode.NOT_FOUND)
+        return empty_pb2.Empty()
+
+    def UpdateTopic(self, request, context):
+        """Repurpose UpdateTopic API for setting up test conditions."""
```
jklukas: Need some more documentation here about what's going on what specific repurposing is going on here, and how to use the setting of statuscode to alter behavior.

ingestion-edge/pubsub_emulator.py
```diff
@@ -0,0 +1,200 @@
...
+                    subscription.pulled.pop(ack_id)
+                except KeyError:
+                    context.set_code(grpc.StatusCode.NOT_FOUND)
+        return empty_pb2.Empty()
+
+    def UpdateTopic(self, request, context):
+        """Repurpose UpdateTopic API for setting up test conditions.
+
+        :param request.topic.name: Name of the topic that needs overrides.
+        "param request.update_mask.paths: A list of overrides, over the form
```
jklukas:
```suggestion
        :param request.update_mask.paths: A list of overrides, of the form
```

ingestion-edge/pubsub_emulator.py
```diff
@@ -0,0 +1,200 @@
...
+        "key=value".
+
+        Valid override keys are "status_code" and "sleep". An override value of
+        "" disables the override.
+
+        For the override key "status_code" the override value indicates the
+        status code that should be returned with an empty response by Publish
+        requests, and non-empty override values must be a property of
```
jklukas:
```suggestion
        requests, and non-empty override values must be a property of type
```

relud: actually i meant that it has to be an attribute of the object

jklukas:
looks good

PR #262, opened 2018-12-06
Create things_that_suck.md

relud:


relud:
cc @mreid-moz

docs/things_that_suck.md
```diff
@@ -0,0 +1,30 @@
+# Overview
```
jklukas: How about we call this `painpoints.md`?

PR #264, opened 2018-12-06
Bug: accessing ValueProvider at runtime in writeFile

### LINKED ISSUE #258: Don't access ValueProviders at compile time
@whd reported a problem where missing `--output` on template compilation throws an exception about value not being available until runtime. And indeed the following line is executed at compile time rather than runtime, breaking the ValueProvider guarantees:

        DynamicPathTemplate pathTemplate = new DynamicPathTemplate(outputPrefix.get());

We should check for other calls to `.get()` on ValueProvider and ensure they're not happening at compile time. And add tests that exercise this.

jklukas:
Closes #258

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
@@ -25,6 +25,7 @@ import org.apache.beam.sdk.io.gcp.pubsub.PubsubIO;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder;
 import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.options.ValueProvider.NestedValueProvider;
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -115,19 +116,25 @@
...
-                .to(pathTemplate.staticPrefix).withNaming(placeholderValues -> Write.defaultNaming(
-                    pathTemplate.replaceDynamicPart(placeholderValues), format.suffix()))));
+        .compose(
+            input -> input.apply("fixedWindows", Window.into(FixedWindows.of(windowDuration)))
+                .apply("writeFile", FileIO.<List<String>, PubsubMessage>writeDynamic()
+                    // We can't pass the attribute map to by() directly since MapCoder isn't
+                    // deterministic;
+                    // instead, we extract an ordered list of the needed placeholder values.
+                    // That list is later available to withNaming() to determine output location.
+                    .by(message -> pathTemplate.get().extractValuesFrom(message.getAttributeMap()))
```
relud: is it possible to do this type of thing later to make all our inputs use value providers? because that would be ideal imo.

PR #283, opened 2018-12-17
Add top-level README

jklukas:
What else would be useful here?


mreid-moz:
I think quick links to `ingestion-beam` and `ingestion-edge` with short descriptions could be useful.

jklukas:
> I think quick links to ingestion-beam and ingestion-edge with short descriptions could be useful.

Done

PR #287, opened 2018-12-18
Make flush a background task

relud:
after discussing with @whd we determined that we can't rely on kube to do anything other than liveness check, readiness check, and waiting some period for clean shutdown after SIGTERM (or some other signal). as such, flush needs to be run as a background task in Sanic.


ingestion-edge/ingestion_edge/publish.py
```diff
@@ -68,42 +70,78 @@
         return await async_wrap(future)
 
 
-async def flush(request: Request, q: SQLiteAckQueue) -> response.HTTPResponse:
-    """Flush messages from the local queue to pubsub.
+def _set_status(q, message, future):
```
jklukas: Should we have type hints here to be consistent with the rest of the file?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -68,42 +70,78 @@
...
+    # run forever
+    while True:
         try:
-            message = q.get(block=False)
-        except Empty:
+            # flush until the queue is empty or a publish fails
+            while True:
+                pending, size = [], 0
+                for _ in range(1000):
+                    if size > 5000000:
```
jklukas: Where do these literals 1000 and 5000000 come from? They should probably be tunable configuration parameters or named constants.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -135,6 +173,10 @@
         ).items()
         if value is not None
     }
+    # assert valid pubsub message
+    for value in attrs.values():
+        if len(value.encode("utf8")) > 1024:
```
jklukas: Can you add a comment here linking to PubSub documentation that specifies this limit?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -162,10 +204,14 @@ def init_app(app: Sanic):
...
     q = SQLiteAckQueue(**queue_config)
-    # route flush
-    app.add_route(partial(flush, q=q), "/__flush__", name="flush")
+    # flush continuously on app start
+    app.add_task(flush(q=q))
     # get metadata_headers config
     metadata_headers = app.config["METADATA_HEADERS"]
+    # validate attribute keys
+    for attribute in metadata_headers.values():
+        if len(attribute.encode("utf8")) > 256:
```
jklukas: Can you add a link for this limitation too?

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -68,42 +70,78 @@
...
-            q.nack(message)
-            # raise from bare except
-            raise
-        else:
-            q.ack(message)
-            result[DONE] += 1
-    return response.json(result, HTTP_STATUS.OK)
+            pass
+        # wait 60 seconds then flush again
+        await asyncio.sleep(60)
```
jklukas: Should the 60 seconds be configurable?

ingestion-edge/ingestion_edge/publish.py
relud: 1000 is the max batch size for the pubsub api , and 5000000 was chosen somewhat arbitrarily to keep the batch from running over the 10MB limit without serializing to protobuf because we can't reuse the serialized message in the client.

that said, they should totally be tunable, and be tuned based on load testing results. will fix.

ingestion-edge/ingestion_edge/publish.py
relud: yes

relud:
> Is there more context on why we're making this change? Previously, were we relying on kube to periodically call the flush endpoint?

after discussing with @whd we determined that we can't rely on kube to do anything other than liveness check, readiness check, and waiting some period for clean shutdown after SIGTERM (or some other signal).

> Is the thought that this approach is simpler?

it definitely wasn't easier to test 'v_v, but it is easier than configuring an extra kube service to hit `/__flush__` on every running container.

> Would we potentially still want to be able to force a flush via endpoint as part of shutdown?

good idea, i think we should register an `after_server_stop` handler that attempts one flush on clean shutdown. will fix.

> Do we also need to change documentation that references a flush endpoint as part of this change?

yes absolutely, will fix.

PR #289, opened 2018-12-19
Remove and parse "metadata" field in ParsePayload

### LINKED ISSUE #279: Appending "metadata" field could result in invalid JSON
In the [`AddMetadata`](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java#L31) decoder, we insert a top-level `"metadata"` field in the JSON representation. If the payload itself already includes a `"metadata"` field we could end up with a duplicate key. 

Do we enforce that `"metadata"` is not allowed at the top level of incoming submissions? 

This could be handled at the schema level or in this codebase (or both).

jklukas:
Resolves #279

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -21,6 +21,13 @@ import org.everit.json.schema.loader.SchemaLoader;
 import org.json.JSONException;
 import org.json.JSONObject;
 
+/**
+ * A {@code PTransform} that parses the message's payload as a {@link JSONObject}, sets
+ * some attributes based on the content, and validates that it conforms to its schema.
+ *
+ * <p>There are several unrelated concerns all packed into this single transform so that we
+ * incur the cost of parsing the JSON only once.
+ */
 public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
 
   private static class SchemaNotFoundException extends Exception {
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -97,11 +104,29 @@ protected PubsubMessage processElement(PubsubMessage element)
...
+    Object untypedMetadata = json.remove("metadata");
+    if (untypedMetadata instanceof JSONObject) {
+      JSONObject metadata = (JSONObject) untypedMetadata;
+      for (String key : metadata.keySet()) {
+        Object value = metadata.get(key);
+        if (value != null) {
+          attributes.put(key, metadata.get(key).toString());
+        }
+      }
+    }
```
jklukas: Doing this here is definitely imperfect, since we won't have access to these new attributes for any earlier stages of the pipeline. Thus, we wouldn't be able to run the full Decoder based on a file in `text` format because ParseUri would throw errors about not having a uri attribute.

But we could build a future variant of the pipeline that calls this `ParsePayload` transform without the preceding steps that exist in Decoder.

In any case, this seems a bit safer than just straight-up dropping a "metadata" field if it exists.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -97,11 +104,29 @@ protected PubsubMessage processElement(PubsubMessage element)
...
+
+    // Remove any top-level "metadata" field if it exists, and attempt to parse it as a
+    // key-value map of strings, adding all entries as attributes.
+    Object untypedMetadata = json.remove("metadata");
+    if (untypedMetadata instanceof JSONObject) {
+      JSONObject metadata = (JSONObject) untypedMetadata;
+      for (String key : metadata.keySet()) {
+        Object value = metadata.get(key);
+        if (value != null) {
+          attributes.put(key, metadata.get(key).toString());
```
jklukas: One could argue that we should drop a non-string value or throw an exception rather than coercing it to a string and continuing. I'm open to feedback there.

relud: i think this is fine for now

PR #293, opened 2018-12-20
Dynamic destinations for BigQuery output

### LINKED ISSUE #291: Dynamic destinations for BigQuery output
Similar to what we allow for file-based output, we should allow the destination table for BigQuery output to depend on attributes of the message. We might have an option that looks like:

    --output='someProject.${document_namespace}.${document_type}_${document_version}'

Under the hood, we'd use the `.to()` variant of BigQueryIO that takes a function of the input values to determine the destination.

jklukas:
Closes #291

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -5,13 +5,57 @@
 package com.mozilla.telemetry.transforms;
 
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.common.collect.Maps;
 import com.mozilla.telemetry.util.Json;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+import org.apache.beam.sdk.coders.KvCoder;
+import org.apache.beam.sdk.io.gcp.bigquery.TableDestination;
+import org.apache.beam.sdk.io.gcp.bigquery.TableDestinationCoderV2;
+import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.options.ValueProvider;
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
```diff
@@ -89,16 +90,45 @@ public void canWriteToBigQuery() throws Exception {
     String input = Resources.getResource("testdata/json-payload.ndjson").getPath();
     String output = String.format("%s:%s", projectId, tableSpec);
 
-    PipelineResult result = Sink.run(new String[] { "--inputFileFormat=text", "--inputType=file",
-        "--input=" + input, "--outputFileFormat=text", "--outputType=bigquery",
-        "--output=" + output, "--errorOutputType=stderr" });
+    PipelineResult result = Sink
+        .run(new String[] { "--inputFileFormat=text", "--inputType=file", "--input=" + input,
+            "--outputType=bigquery", "--output=" + output, "--errorOutputType=stderr" });
```
jklukas: Unrelated change: there was a spurious `"--outputFileFormat=text"` that I removed here.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -5,13 +5,57 @@
...
+import org.apache.beam.sdk.io.gcp.bigquery.TableDestinationCoderV2;
+import org.apache.beam.sdk.io.gcp.bigquery.TableRowJsonCoder;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.commons.text.StringSubstitutor;
 
-public class PubsubMessageToTableRow extends MapElementsWithErrors<PubsubMessage, TableRow> {
+public class PubsubMessageToTableRow
```
jklukas: The name of this class is now a bit of a stretch, since it returns a KV, but I couldn't think of any clearer alternative.

PR #294, opened 2018-12-20
Make PublisherClient ephemeral

relud:


jklukas:
On first read, this looks fine. Can you provide context on the motivation for the change? Does this fix something that was broken?

relud:
This makes testing easier

PR #299, opened 2018-12-21
Make geoCityDatabase and geoCityFilter runtime parameters

jklukas:
Also fixes up some docs and adds a workaround for default value of deduplicateExpireDuration

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
@@ -26,20 +26,19 @@ public interface DecoderOptions extends SinkOptions, PipelineOptions {
 
   @Description("Path (local or gs://) to GeoIP2-City.mmdb")
   @Validation.Required
-  String getGeoCityDatabase();
+  ValueProvider<String> getGeoCityDatabase();
 
-  void setGeoCityDatabase(String value);
+  void setGeoCityDatabase(ValueProvider<String> value);
 
   @Description("Path (local or gs://) to newline-delimited text file listing city names to allow"
       + " in geoCity information; cities not in the list are considered too small to ensure"
       + " user anonymity so we won't report geoCity in that case."
       + " If not specified, no limiting is performed and we always report valid geoCity values.")
-  String getGeoCityFilter();
+  ValueProvider<String> getGeoCityFilter();
 
-  void setGeoCityFilter(String value);
+  void setGeoCityFilter(ValueProvider<String> value);
 
-  @Description("Source of messages to mark as seen for deduplication."
-      + " Allowed sources are:"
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -59,9 +58,9 @@
 
   void setRedisUri(ValueProvider<String> value);
 
-  @Description("Duration for which message ids should be stored for deduplication. "
-      + "Defaults to 24h. Allowed formats are: Ns (for seconds, example: 5s), "
-      + "Nm (for minutes, example: 12m), Nh (for hours, example: 2h).")
+  @Description("Duration for which message ids should be stored for deduplication."
+      + " Allowed formats are: Ns (for seconds, example: 5s), "
+      + " Nm (for minutes, example: 12m), Nh (for hours, example: 2h).")
```
relud: there's now a double-space here

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -59,9 +58,9 @@
 
   void setRedisUri(ValueProvider<String> value);
 
-  @Description("Duration for which message ids should be stored for deduplication. "
-      + "Defaults to 24h. Allowed formats are: Ns (for seconds, example: 5s), "
-      + "Nm (for minutes, example: 12m), Nh (for hours, example: 2h).")
+  @Description("Duration for which message ids should be stored for deduplication."
+      + " Allowed formats are: Ns (for seconds, example: 5s), "
```
relud:
```suggestion
      + " Allowed formats are: Ns (for seconds, example: 5s),"
```

ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -91,9 +91,9 @@
 
   void setIncludeStackTrace(Boolean value);
 
-  @Description("Fixed window duration. Defaults to 10m. Allowed formats are: "
```
relud: these two aren't value providers, why are they losing their default value from the doc string?

jklukas: Because default gets automatically injected into the help output. This one looks like:

```
  --windowDuration=<String>
    Default: 10m
    Fixed window duration. Allowed formats are: Ns (for seconds, example: 5s),
    Nm (for minutes, example: 12m), Nh (for hours, example: 2h).
```

That may have been a change in a more recent version of beam.

relud: awesome!

PR #302, opened 2018-12-21
Move black to same location as other style checks

relud:


jklukas:
I removed format-python from the list of required checks for the repo.

Issue #246, opened 2018-11-29
Maven complains about threads after submitting Dataflow pipeline


jklukas:
cc @whd 

[One of Google's examples](https://github.com/GoogleCloudPlatform/DataflowTemplates/blob/2296750e6f379f6fd97806d3e195967dfaf0767d/src/main/java/com/google/cloud/teleport/templates/PubsubToText.java#L51) sets `-Dexec.cleanupDaemonThreads=false`.

I took a first look at [Maven's docs for that option](https://www.mojohaus.org/exec-maven-plugin/java-mojo.html#cleanupDaemonThreads) and haven't yet developed an opinion about whether that's the right thing to do.

jklukas:
Note that this isn't a problem when using maven to create a Dataflow template, and then executing a pipeline based on the template.

jklukas:
Fixed in https://github.com/mozilla/gcp-ingestion/pull/303

Issue #242, opened 2018-11-28
Add error info to failed BigQuery inserts


jklukas:
This is listed as a TODO in `OutputType.java`.

jklukas:
We will want to enable extended errors so we can call `getFailedInsertsWithErr()`. Relevant info:

- [InsertAll response JSON schema](https://cloud.google.com/bigquery/docs/reference/rest/v2/tabledata/insertAll#response)
- [BigQuery troubleshooting docs](https://cloud.google.com/bigquery/troubleshooting-errors#sample-error-response)

jklukas:
Fixed in #303

PR #303, opened 2018-12-21
Beam fixups

jklukas:
Solves a number of issues/irritations around deploying Decoder. More info inline.


ingestion-beam/pom.xml
```diff
@@ -194,6 +194,9 @@
                         </configuration>
                     </execution>
                 </executions>
+                <configuration>
+                  <cleanupDaemonThreads>false</cleanupDaemonThreads>
+                </configuration>
```
jklukas: Fixes #246

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -81,9 +81,10 @@ private static void loadSchema(String name) throws SchemaNotFoundException {
   }
 
   private static Schema getSchema(String name) throws SchemaNotFoundException {
-    final Schema schema = schemas.get(name);
+    Schema schema = schemas.get(name);
     if (schema == null) {
-      throw SchemaNotFoundException.forName(name);
+      loadSchema(name);
+      schema = schemas.get(name);
```
jklukas: Decoder running in Dataflow would always die on ParsePayload because it could never find the schemas.

The documentation for PTransform explains how PTransform is marked Serializable because it might be captured by a DoFn's closure, but its implementations of Serializability are no-ops. So this is actually documented behavior I hadn't previously understood.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -157,12 +162,31 @@
           .withMethod(BigQueryIO.Write.Method.STREAMING_INSERTS) //
           .withCreateDisposition(BigQueryIO.Write.CreateDisposition.CREATE_NEVER) //
           .withWriteDisposition(BigQueryIO.Write.WriteDisposition.WRITE_APPEND).skipInvalidRows()
+          .withExtendedErrorInfo().ignoreUnknownValues()
```
jklukas: If our payload contains additional fields not in the destination table schema, this setting makes us just drop those fields rather than erroring out.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -157,12 +162,31 @@
...
+                  attributes.put("error_location", errorProto.getLocation());
+                  attributes.put("error_reason", errorProto.getReason());
+                }
+                if (bqie.getError().getErrors().size() > 1) {
+                  // If there are additional errors, we include the entire JSON response.
+                  attributes.put("insert_errors", bqie.getError().toString());
+                }
+                byte[] payload = bqie.getRow().toString().getBytes();
+                return new PubsubMessage(payload, attributes);
+              }));
```
jklukas: Closes #242

PR #312, opened 2018-12-27
Add integration tests for ingestion-edge

relud:


ingestion-edge/tests/integration/conftest.py
```diff
@@ -17,11 +21,11 @@ def pytest_addoption(parser: _pytest.config.argparsing.Parser):
         help="Server to run tests against",
     )
     parser.addoption(
-        "--no-create-pubsub-resources",
-        action="store_false",
-        dest="create_pubsub_resources",
-        default=True,
-        help="Don't create PubSub resources for tests",
```
relud: this option doesn't really make sense imo after moving pubsub resources to pytest fixtures

relud:
This is the first round of integration tests that I've gotten working and cleaned up.

I have a few more ready submit after these are merged.

ingestion-edge/tests/integration/publish/test_healthy.py
```diff
@@ -0,0 +1,121 @@
...
+            ([".", "p/a/t/h", "MAX_LENGTH"]),  # uri_suffix
+            # list of None so cycle doesn't generate infinite tests
+            [None]
+            * max(
+                5, 4 + len(set(METADATA_HEADERS.keys()).difference({"content-length"}))
+            ),
+        )
+    ],
+)
+def test_submit_success(
```
relud: `test_submit_success` is an expansion of `test_publish` from the former `test_api.py` file, expanded to cover more cases, especially around message and attribute size.

ingestion-edge/tests/integration/test_dockerflow.py
```diff
@@ -0,0 +1,27 @@
...
+def test_lbheartbeat(requests_session: requests.Session, server: str):
+    r = requests_session.get(server + "/__lbheartbeat__")
+    r.raise_for_status()
+    assert r.text == ""
+
+
+def test_version(requests_session: requests.Session, server: str):
+    r = requests_session.get(server + "/__version__")
+    r.raise_for_status()
+    assert sorted(r.json().keys()) == ["build", "commit", "source", "version"]
```
relud: moved these tests out of `test_api.py` for better separation of concerns.

ingestion-edge/tests/integration/publish/test_healthy.py
```diff
@@ -0,0 +1,121 @@
...
+from ingestion_edge.config import METADATA_HEADERS
+from itertools import cycle, chain
+from typing import Dict, Optional
+import gzip
+import pytest
+
+# avoid haveing long strings in parametrized test
+# names by outsourcing the actual values
+DATA = {
+    b"GZIPPED": gzip.compress(b"data"),
```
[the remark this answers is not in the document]
relud: i had at one time intended to use something closer to a default map with a default value equal to the key and this was a holdover from that. i changed it to be string keys, because it doesn't matter.

ingestion-edge/tests/integration/publish/test_healthy.py
```diff
@@ -0,0 +1,121 @@
...
+            ),  # headers
+            ([".", "p/a/t/h", "MAX_LENGTH"]),  # uri_suffix
+            # list of None so cycle doesn't generate infinite tests
+            [None]
+            * max(
+                5, 4 + len(set(METADATA_HEADERS.keys()).difference({"content-length"}))
+            ),
+        )
+    ],
+)
```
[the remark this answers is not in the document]
relud: I moved the cases outsize of the `parametrize`, I agree it definitely improves readability.

As for cycling through args and uri_suffix in sync, it's a side effect i'm aware of. I wouldn't say it's intended so much as allowed. The real goal is to each value, but each combination of each value makes an excessive number of tests, so I eventually settled on this.

Issue #137, opened 2018-10-26
Add integration tests for ingestion-edge with insufficient pubsub permissions


relud:

relud:
fixed

relud:
fixed

Issue #136, opened 2018-10-26
Add integration tests for ingestion-edge where pubsub resource does not exist


relud:

relud:
fixed

PR #317, opened 2018-12-28
Add more integration tests for ingestion-edge

relud:
This also fixes an issue where `PublisherClient` would try to deliver the message for ten minutes on 500 and then the retry exception wouldn't be caught and due to threading would never return.


relud:
fixes #136 
fixes #137

PR #332, opened 2019-01-02
Use context.abort in pubsub_emulator

relud:
`context.set_code` was the wrong way to return specific status codes in with grpc.
`context.abort` raises an exception, so the code no longer needs to ensure a valid return after an abort. This removed line 102, and simplified cases where an `else` was needed to ensure work was not done following an exception.
`context.abort` takes a status message, which can be helpful when debugging.


ingestion-edge/pubsub_emulator.py
```diff
@@ -59,60 +59,55 @@ def create_server(self):
...
         except KeyError:
-            context.set_code(grpc.StatusCode.NOT_FOUND)
-        else:
-            message_ids = [uuid.uuid4().hex for _ in range(len(request.messages))]
-            for _id, message in zip(message_ids, request.messages):
-                message.message_id = _id
-            for subscription in subscriptions:
-                subscription.published.extend(request.messages)
+            context.abort(grpc.StatusCode.NOT_FOUND, "Topic not found")
+        message_ids = [uuid.uuid4().hex for _ in range(len(request.messages))]
```
jklukas: Style suggestion unrelated to the content of your changes:

```suggestion
        message_ids = [uuid.uuid4().hex for _ in request.messages]
```

PR #333, opened 2019-01-02
Add types to pubsub_emulator api calls

relud:


ingestion-edge/pubsub_emulator.py
```diff
@@ -56,23 +56,29 @@ def create_server(self):
...
         try:
             self.topics.pop(request.topic)
         except KeyError:
             context.set_code(grpc.StatusCode.NOT_FOUND)
         return empty_pb2.Empty()
 
-    def CreateSubscription(self, request, context):  # noqa: D403
+    def CreateSubscription(
+        self, request: pubsub_pb2.Subscription, context: grpc.ServicerContext
+    ):  # noqa: D403
```
jklukas: You can probably get rid of these D403 comments by wrapping CreateSubscription in double backticks to mark it as inline code in ReStructured Text syntax.

PR #336, opened 2019-01-03
Account for Google forwarding rule IP in GeoCityLookup

### LINKED ISSUE #327: Geo indicates Mountain View 80% of the time
Unclear why this is happening at this point, but it looks like the majority of pings flowing through GCP are being decoded with geo indicating Mountain View. This includes payloads I've sent from Worthington, OH through AWS telemetry tee-d to GCP. Also, payloads I've sent directly to the GCP endpoint.

jklukas:
Fixes #327

### DIFF
--- a/ingestion-beam/README.md
+++ b/ingestion-beam/README.md
@@ -388,8 +388,11 @@ unmodified.
 ### GeoIP Lookup
 
 1. Extract `ip` from the `x_forwarded_for` attribute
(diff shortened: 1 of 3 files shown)


whd:
I sent a couple of pings from eu-west1 in AWS and they are now showing up as from Dublin, so this fix appears to be effectual.

PR #342, opened 2019-01-04
Apply gzip compression and decompression by default

### LINKED ISSUE #324: Compression strategy
We have observed a Decoder failure with a message too large to publish to Pubsub. This is likely due to the payload being gzipped when sent to the edge, and then the Decoder is unzipping the content and publishing the unzipped payload, which is substantially larger than the original message.

Firefox should never send a payload larger than 10 MB, and the existing AWS edge should be unzipping and rejecting any payload larger than 10 MB uncompressed, but the error we saw actually complained about a 7 MB payload. Pubsub requires a base64 encoded payload, which ends up being 4 bytes per 3 bytes of input, so the limit for an uncompressed and unencoded payload to Pubsub is actually 7 MB rather than 10 MB.

I need to look into whether the Pubsub service and clients have any built-in support for compression, or whether we should be handling this ourselves.

jklukas:
Closes #324


ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -66,6 +74,12 @@
 
   void setErrorOutputType(ErrorOutputType value);
 
+  @Description("Compression format for --errorOutputType=file")
+  @Default.Enum("GZIP")
+  Compression getErrorOutputFileCompression();
+
+  void setErrorOutputFileCompression(Compression value);
+
```
jklukas: The file compression formats can't be ValueProviders due to the FileIO API lacking such an overload.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/DecompressPayload.java
```diff
@@ -0,0 +1,92 @@
...
+    private final Counter compressedInput = Metrics.counter(Fn.class, "compressed-input");
+    private final Counter uncompressedInput = Metrics.counter(Fn.class, "uncompressed-input");
+
+    @Override
+    public PubsubMessage apply(PubsubMessage value) {
+      byte[] payload = value.getPayload();
+      if (enabled.isAccessible() && !enabled.get()) {
+        // Compression has been explicitly turned off in options, so return unchanged message.
+        return value;
+      } else if (!isGzip(payload)) {
```
relud: is detecting gzip here more efficient than just catching an IOException in the next block?

jklukas: I don't have evidence at this point to indicate how expensive it is to let decompression throw an exception on uncompressed input. I added this primarily due to observing metrics on the running Decoder job. This decompression step was the second most expensive step behind ParsePayload.

I'm open to reverting this if you're concerned about adding the complexity.

jklukas: Removed. We can measure this and add the implementation back as part of the zstd project if we take that up.

PR #347, opened 2019-01-07
 Don't catch Throwable in MapElementsWithErrors

jklukas:
In testing the Decoder, I've seen some records in the error collection due to
OutOfMemoryError. We likely want the worker to fail in this case rather than
send the message to error output.

In general, it's a bit dangerous to catch Throwable since system Errors
(which don't inherit from exception) are generally due to some unrecoverable
state that means the process needs to go down quickly.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
@@ -222,8 +222,8 @@ public enum OutputType {
 
     @Override
     public ResultWithErrors<? extends POutput> expand(PCollection<PubsubMessage> input) {
-      return ResultWithErrors.of(input.apply(inner),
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -223,7 +223,7 @@ public EmptyErrors(PTransform<PCollection<PubsubMessage>, ? extends POutput> inn
     @Override
     public ResultWithErrors<? extends POutput> expand(PCollection<PubsubMessage> input) {
       return ResultWithErrors.of(input.apply(inner),
-          input.getPipeline().apply(Create.empty(PubsubMessageWithAttributesCoder.of())));
+          input.getPipeline().apply("Empty error collection", Create.empty(PubsubMessageWithAttributesCoder.of())));
```
jklukas: Small unrelated change. It's difficult to understand in the Dataflow UI what this Create.Values instance means.

PR #349, opened 2019-01-08
Filter large payloads to error output when writing to BigQuery

jklukas:
We found that the pipeline was throwing runtime exceptions due to 404 responses relating to payloads over 10 MB.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry.transforms;
+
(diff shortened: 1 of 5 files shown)


jklukas:
I'm looking into why the integration test keeps crashing.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -0,0 +1,52 @@
...
+package com.mozilla.telemetry.transforms;
+
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.metrics.Counter;
+import org.apache.beam.sdk.metrics.Metrics;
+
+public class LimitPayloadSize extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
+
+  /** Exception to throw for messages whose payloads exceed the configured size limit. */
+  public static class PayloadTooLargeException extends RuntimeException {
```
relud: why does this inherit RuntimeException instead of CheckedException?

jklukas: I will change it to inherit from Exception. This is mostly just habit of preferring runtime exceptions since exception checking seems to be largely regarded as a failed feature of Java. Checked is fine in this case.

PR #355, opened 2019-01-10
Implement pubsub batch for asyncio

relud:
This resolves three main issues with the default batch:
- Handle RetryError in PublisherClient.publish
- Raise ValueError when single message exceeds max bytes
- Return asyncio.Task instead of google.api_core.futures.Future

Fixing those issues in turn fixes 4 new integration tests.


ingestion-edge/ingestion_edge/publish.py
```diff
@@ -70,10 +82,15 @@
...
-    # Get PubSub timeout
-    timeout = app.config.get("PUBLISH_TIMEOUT_SECONDS")
     # Initialize PubSub client
+    timeout = app.config.get("PUBLISH_TIMEOUT_SECONDS", None)
     client = PublisherClient()
+    client.api.publish = partial(
+        client.api.publish,
+        retry=Retry(TRANSIENT_ERRORS, deadline=timeout),
+        timeout=timeout,
+    )
```
relud: python threads are unkillable, so they have to time out on their own. The publish api call happens in a thread, so we need to use the built in `retry` and `timeout` or we get unusual behavior where everyone waiting times out, but the api call is still retrying in a thread.

relud: this is also why `FLUSH_PUBLISH_TIMEOUT_SECONDS` went away, because we don't control batching we have to set a timeout to apply to all batches.

ingestion-edge/tests/integration/conftest.py
```diff
@@ -100,8 +102,14 @@ def server(
             assert process.poll() is None  # server still running
             yield "http://localhost:%d" % ports.pop()
         finally:
-            process.kill()
-            process.wait()
+            try:
+                # allow one second for graceful termination
+                process.terminate()
```
relud: this allows `--cov` to pick up coverage from integration tests when `--server` isn't specified

ingestion-edge/ingestion_edge/util.py
```diff
@@ -4,41 +4,102 @@
...
-    """Wrap a threading-like future in an async future that can be awaited.
+class AsyncioBatch:
+    """Batch for google.cloud.pubsub_v1.PublisherClient in asyncio.
 
-    example:
-
-    async def func():
-        future = ...
-        await wrap(future)
+    Work around for https://github.com/googleapis/google-cloud-python/issues/7104
```
jklukas:
```suggestion
    Workaround for https://github.com/googleapis/google-cloud-python/issues/7104
```

Apparently, Webster's has the noun as "work-around", but the rest have it as a single unhyphenated word. I do what the OED says and they say "workaround".

This is a very important comment.

ingestion-edge/ingestion_edge/util.py
```diff
@@ -4,41 +4,102 @@
...
+            return []
+
+        response = await asyncio.get_running_loop().run_in_executor(
+            None, self.client.api.publish, self.topic, self.messages
+        )
+
+        if len(response.message_ids) != len(self.messages):
+            raise PublishError("Some messages not successfully published")
+
+        return response.message_ids
```
jklukas: I'm still wrapping my head around much of python's async features, so reading this is a good learning experience. I like the way this reads.

Is there any need to cancel tasks once we commit? If `full` is set due to hitting max size, what happens to the `monitor` task? Does it still sleep and then set `full` again? Does that prevent this class instance from getting cleaned up?

relud: > Is there any need to cancel tasks once we commit?

I tried to do that in flush, it introduced a weird bug that i couldn't figure out how to fix where it propagated down the stack to `run_in_exector` and raised `concurrent.futures.CancelledError` instead of `asyncio.CancelledError`, and that would in turn kill `uvloop`, instead of raising back up the stack.

> If `full` is set due to hitting max size, what happens to the `monitor` task? Does it still sleep and then set `full` again?

yes, but setting `full` again is a noop because the event is already set.

Note that `asyncio.Event` stays set and waiting on it after that is a noop, where `asyncio.Condition` will only notify pending tasks and waiting on it after that waits for the next notify. This is the exact opposite of what I expect from the names, but it's been years since I last dealt with concurrency primitives.

> Does that prevent this class instance from getting cleaned up?

nope, the task will end, the other tasks will already all be complete, there will be no references, and it will be cleaned up.

PR #360, opened 2019-01-11
Modify JSON payloads to make submission_timestamp a top-level field

### LINKED ISSUE #339: Insert submission_timestamp as a top-level field in Decoder
Per [BiqQuery docs on column-partitioned tables](https://cloud.google.com/bigquery/docs/creating-column-partitions#limitations):

> The partitioning column must be either a scalar DATE or TIMESTAMP column. While the mode of the column may be REQUIRED or NULLABLE, it cannot be REPEATED (array-based). Additionally, the partitioning column must be a top-level field. You cannot use a leaf field from a RECORD (STRUCT) as the partitioning column.

I had been hoping to partition all tables by `metadata.submission_timestamp`, but this violates the requirement about nested columns.

It looks like we'll have to insert `submission_timestamp` at the top level of the payload rather than having it inside a `metadata` struct.

jklukas:
Closes #339

I've already deployed this code to Dataflow and recreated the `core_v9` and `core_v10` tables to match the new schema and to partition on submission_timestamp. Records are flowing in!

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -5,6 +5,7 @@
 package com.mozilla.telemetry.transforms;
 
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.Maps;
 import com.mozilla.telemetry.util.Json;
 import java.io.IOException;
@@ -21,6 +22,14 @@ import org.apache.beam.sdk.values.KV;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.commons.text.StringSubstitutor;
 
+/**
+ * Parses JSON payloads using Google's JSON API model library, emitting a BigQuery-specific
+ * TableRow.
+ *
+ * <p>We also perform some light manipulation of the parsed JSON to match details of our table
+ * schemas in BigQuery. In particular, we pull out submission_timestamp to a top level field so we
+ * can use it as our partitioning field.
+ */
 public class PubsubMessageToTableRow
     extends MapElementsWithErrors<PubsubMessage, KV<TableDestination, TableRow>> {
 
@@ -48,7 +57,7 @@ public class PubsubMessageToTableRow
           + tableSpecTemplate.get());
     }
     TableDestination tableDestination = new TableDestination(tableSpec, null);
-    TableRow tableRow = Json.readTableRow(element.getPayload());
+    TableRow tableRow = buildTableRow(element.getPayload());
     return KV.of(tableDestination, tableRow);
   }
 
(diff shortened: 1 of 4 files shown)


jklukas:
My thought is that `submission_timestamp` is naturally metadata and that the transformation we're making here is a BigQuery implementation detail, so we should delay the transformation until writing to BigQuery. Other outputs might not have the same restriction and it would be cleaner to leave submission_timestamp in metadata. But yes, it also spreads out the logic a bit. And other sinks might have the same restriction on partitioning.

jklukas:
I am going to merge now, though, to fix current issues, and we can continue discussing.

relud:
> Other outputs might not have the same restriction

one of the main uses i'm expecting from the json-in-s3 output is loading it into bigquery with a new schema, which means that it also needs this.

also, I think the decoding should output a single consistent schema across all outputs, so that people get a consistent experience and aren't confused when they need to switch between the two.

jklukas:
> the decoding should output a single consistent schema across all outputs

That's great if we can achieve it and optimizing for that would likely mean not using nested schemas at all, since that's one of the places where support in various systems will be different. Flattening metadata in with other fields makes it less obvious, though, what's metadata and what's not.

What would you think about a convention of prepending `meta_` to the names of attributes populated from pipeline metadata? The prefix would give an indication of what fields are populated from metadata vs. populated from the payload, and behavior would likely be more consistent across destination systems.

relud:
i think that for flattened systems we should indicate nesting using a field seperator (ideally `.`, but may not be available), so that things feel like they are in similar locations. this has been done in the past for heka messages, iirc.

in athena we have to flatten metadata to be mutable without a version bump, even though it supports nesting, but i would prefer to avoid flattening unless we absolutely have to.

relud:
also the flattening prefix used right now is `metadata_`

https://bugzilla.mozilla.org/show_bug.cgi?id=1403608

jklukas:
> one of the main uses i'm expecting from the json-in-s3 output is loading it into bigquery with a new schema, which means that it also needs this.

We could define a workflow for backfilling that involves a non-partitioned staging table with submission_timestamp inside the `metadata` struct (to match the JSON on disk), then we do a:

```
# Approximate syntax
SELECT 
  metadata.submission_timestamp AS submission_timestamp, 
  * EXCEPT metadata.submission_timestamp 
FROM staging_table 
INTO destination_table
```

This is admittedly more complicated.

@relud - Do you feel pretty confident that `submission_timestamp` is a special case? Have you seen instances of any of the other metadata fields needing special handling? It may well be that this is the pragmatic thing to do.

relud:
> Do you feel pretty confident that `submission_timestamp` is a special case?

hmm. clustering (bigquery's kind-of sorting thing) might also have this limitation, will check. @whd you have more experience with this, do you think needing a metadata field to be top-level like `submission_timestamp` is an outlier or the norm?

another thought: perhaps we should keep this in the bigquery output, but not remove the value from the metadata struct. that way the standard place is present, and the bigquery-specific location has solid explanation to not be part of the default schema everywhere.

PR #368, opened 2019-01-15
Send records to error output if BQ table does not exist.

jklukas:
This will be necessary to unblock moving to batch loading for BigQuery,
since BigQueryIO raises a pipeline execution exception in case of nonexistent
table or any single row being rejected.

Once this is merged, I will take down the BQ sink Dataflow job, delete
placeholder tables we had previously created, and launch a new job.
This may significantly improve performance of the sink as well, since we
won't be sending a large volume of data we know won't validate given current
table schemas.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -4,15 +4,25 @@
 
 package com.mozilla.telemetry.transforms;
 
+import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableRow;
+import com.google.cloud.bigquery.BigQuery;
+import com.google.cloud.bigquery.BigQueryOptions;
+import com.google.cloud.bigquery.TableId;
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -38,6 +48,11 @@ public static PubsubMessageToTableRow of(ValueProvider<String> tableSpecTemplate
   }
 
   private final ValueProvider<String> tableSpecTemplate;
+  private final Cache<TableId, Boolean> tableExistenceCache = CacheBuilder.newBuilder()
+      .expireAfterWrite(Duration.ofMinutes(1)).build();
```
relud: is this relevant to here: (does the cache not get serialized and this value goes null outside testing?) https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L52-L54

jklukas: That's possible. Good call. Will follow up with a PR to make this also a transient field we initialize on first use.

PR #372, opened 2019-01-16
Add file-based (batch) loading for BigQuery

### LINKED ISSUE #353: Add BigQuery batch loader
We should probably rely on the batch mode for most things, and we can figure out how to special case streaming for certain things down the line.

jklukas:
Closes #353 

Depends on #370 and will change to base of master once that's merged.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
@@ -14,6 +14,7 @@ import com.mozilla.telemetry.transforms.PubsubMessageToTableRow;
 import com.mozilla.telemetry.transforms.ResultWithErrors;
 import com.mozilla.telemetry.util.DerivedAttributesMap;
 import com.mozilla.telemetry.util.DynamicPathTemplate;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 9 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -164,56 +166,64 @@
...
-              }));
+          .ignoreUnknownValues();
+
+      if (writeMethod == BigQueryWriteMethod.streaming) {
+        writeTransform = writeTransform
+            .withFailedInsertRetryPolicy(InsertRetryPolicy.retryTransientErrors()).skipInvalidRows() //
+            .withExtendedErrorInfo();
+      } else if (writeMethod == BigQueryWriteMethod.file_loads) {
+        if (inputType == InputType.pubsub) {
+          // Trigger frequency must be set iff we're in streaming mode.
```
relud:
```suggestion
          // Trigger frequency must be set iff we're in batch mode.
```

jklukas: Nomenclature here gets iffy here, so I'll clarify the comment. Trigger frequency must be set if and only if the overall pipeline is in streaming mode (due to pubsub input) and the bigquery output is in file_loads mode.

relud: oooh, okay, i see now

PR #385, opened 2019-01-22
Drop duplicate messages

jklukas:
Currently, duplicate messages go to error output, but we have no particular reason to save them.

cc @whd

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
@@ -6,6 +6,7 @@ package com.mozilla.telemetry.decoder;
 
 import com.google.common.primitives.Ints;
 import com.mozilla.telemetry.transforms.MapElementsWithErrors;
+import com.mozilla.telemetry.transforms.ResultWithErrors;
 import com.mozilla.telemetry.util.Time;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
```diff
@@ -59,8 +62,22 @@ Jedis getJedis() {
...
+      PTransform<PCollection<PubsubMessage>, ResultWithErrors<PCollection<PubsubMessage>>> //
+      removeDuplicates(ValueProvider<URI> uri) {
+    return PTransform.compose((PCollection<PubsubMessage> input) -> {
+      ResultWithErrors<PCollection<PubsubMessage>> result = input.apply(new RemoveDuplicates(uri));
+      PCollection<PubsubMessage> errorsStrippedOfDuplicates = result.errors()
+          .apply(Filter.by((PubsubMessage message) -> !Optional
+              .ofNullable(message.getAttribute("exception_class"))
+              .filter(RemoveDuplicates.DuplicateIdException.class.getName()::equals).isPresent()));
+      return ResultWithErrors.of(result.output(), errorsStrippedOfDuplicates);
+    });
```
relud: I think it would be better to route duplicates to a separate output when caught, rather than filtering them out here.

PR #387, opened 2019-01-23
Improve the error collection interface

### LINKED ISSUE #241: Closer consideration of naming transforms
Reading through some of the naming logic in the Beam codebase, it looks like it does a pretty good job of inferring display names for transforms.

I'm running a test right now where I remove most of the explicit names in our `.apply("explicit name", transform)` calls. The Dataflow console generally shows the class name (like ParseUri) in this case:

screen shot 2018-11-28 at 2 03 00 pm

I'd argue that letting beam supply the class name in most cases makes for a better experience. What you see in the console directly relates to the class name being executed. In the case of `AnonymousTransform`, we can certainly still supply explicit names to aid in comprehension.

### LINKED ISSUE #386: Route duplicate messages to separate pcollection
in #385 we added filtering to drop duplicates, but it happens after routing to the error pcollection and transforming the message into a error message. it would be more efficient if we just routed duplicates differently from other exceptions, so that we could ignore the duplicates pcollection instead of filtering.

jklukas:
See the new WithErrors class for the main new concept here.
The transforms that emit errors now return a result that has an errorsTo()
method that makes their usage in the top-level Sink and Decoder pipelines
simpler.

This also closes #386 as it refactors the Deduplicate.java code to use the
new error interface.

All the logic for our inputs and outputs have been factored into an io module
and also now use the new error interface consistently.

These refactorings also aid in how the pipeline is displayed in the dataflow UI
so I've removed most of the explicit string names as we apply transforms
since the class names are now instructive. Thus also closes #241.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
```diff
@@ -4,128 +4,169 @@
...
+          TupleTag<PubsubMessage> errorTag) {
+        return new AutoValue_Deduplicate_RemoveDuplicates_Result(tuple, outputTag, duplicateTag, errorTag);
+      }
+
+      /**
+       * Ignore the collection of duplicates and return a {@link Result} of just the
+       * non-duplicate output and the errors.
+       */
+      public WithErrors.Result<PCollection<PubsubMessage>> ignoreDuplicates() {
+        return WithErrors.Result.of(tuple().get(outputTag()), outputTag(), tuple().get(errorTag()), errorTag());
```
relud: I don't see where `WithErrors.Result.of` is defined to tag pairs of pcollections and tags. could you explain to me what's happening here?

ingestion-beam/src/main/java/com/mozilla/telemetry/Sink.java
```diff
@@ -50,14 +50,14 @@ public static PipelineResult run(SinkOptions.Parsed options) {
 
     // Trailing comments are used below to prevent rewrapping by google-java-format.
     pipeline //
-        .apply("input", options.getInputType().read(options)) //
-        .addErrorCollectionTo(errorCollections).output() //
+        .apply(options.getInputType().read(options)).errorsTo(errorCollections) //
```
relud: :tada: this is a lot nicer

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -0,0 +1,59 @@
...
+      return input //
+          .apply(PubsubIO.readMessagesWithAttributes().fromSubscription(options.getInput()))
+          .apply(ToPubsubMessageFrom.identity());
+    }
+  }
+
+  /** Implementation of reading from local or remote files. */
+  public static class FileInput extends Read {
+
+    private final SinkOptions.Parsed options;
```
relud:
```suggestion
    private final Parsed options;
```

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -0,0 +1,59 @@
...
+ * Implementations of reading from the sources enumerated in {@link
+ * com.mozilla.telemetry.options.InputType}.
+ */
+public abstract class Read
+    extends PTransform<PBegin, WithErrors.Result<PCollection<PubsubMessage>>> {
+
+  /** Implementation of reading from Pub/Sub. */
+  public static class PubsubInput extends Read {
+
+    private final SinkOptions.Parsed options;
```
relud:
```suggestion
    private final Parsed options;
```

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -0,0 +1,59 @@
...
+          .apply(ToPubsubMessageFrom.identity());
+    }
+  }
+
+  /** Implementation of reading from local or remote files. */
+  public static class FileInput extends Read {
+
+    private final SinkOptions.Parsed options;
+
+    public FileInput(Parsed options) {
```
relud: `com.mozilla.telemetry.io.Read` methods are accepting `Parsed options`, while `com.mozilla.telemetry.io.Write` methods are accepting individual options. We should probably do one or the other.

PR #391, opened 2019-01-24
ingestion-beam style cleanup

jklukas:
A variety of changes to be more consistent in our style internally and more
consistent with some conventions of Beam.

We give all "user-facing" transforms private constructors and public static
factory methods, returning a Singleton instance in case of transforms that
take no parameters.

Also more consistent in skipping explicit names for transforms when they
simply repeat the class name.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/WithErrors.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/WithErrors.java
@@ -11,7 +11,6 @@ import java.util.Map;
 import javax.annotation.Nullable;
 import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
-import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.PInput;
@@ -85,7 +84,6 @@ public class WithErrors {
     @Override
     public void finishSpecifyingOutput(String transformName, PInput input,
         PTransform<?, ?> transform) {
-      errors().setCoder(PubsubMessageWithAttributesCoder.of());
     }
   }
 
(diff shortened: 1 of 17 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/DeduplicateTest.java
```diff
@@ -39,8 +39,8 @@ public void testOutput() {
 
     // Create new PubsubMessage with element as document_id attribute
     final MapElements<String, PubsubMessage> mapStringsToId = MapElements
-        .into(new TypeDescriptor<PubsubMessage>() {
-        }).via(element -> new PubsubMessage(new byte[0], ImmutableMap.of("document_id", element)));
+        .into(TypeDescriptor.of(PubsubMessage.class))
+        .via(element -> new PubsubMessage(new byte[0], ImmutableMap.of("document_id", element)));
```
jklukas: We remove a few instances of `new TypeDescriptor` with the simpler `TypeDescriptor.of`

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/WithErrors.java
```diff
@@ -85,7 +84,6 @@ public Pipeline getPipeline() {
     @Override
     public void finishSpecifyingOutput(String transformName, PInput input,
         PTransform<?, ?> transform) {
-      errors().setCoder(PubsubMessageWithAttributesCoder.of());
```
relud: do we just not need this anymore?

jklukas: I think this depends on what type information is available to the particular transform. There is a `PubsubCoderProviderRegistrar` that gets loaded as part of the google cloud beam package. In this particular case of the error collection for `WithErrors`, the collection has a concrete type of `PCollection<PubsubMessage>`, so it's able to use the registered coder.

For MapElementsWithErrors, the output collection type is a type parameter, so the coder registry is not able to infer. Thus, we still do need to explicitly set coder when the output collection type is pubsubmessage.

relud: ah, makes sense. thanks for the explanation!

PR #405, opened 2019-01-30
Pull schemas from GCS rather than building into jar

### LINKED ISSUE #398: Consider GCS-based schema artifacts
Right now, schemas are pulled in as resources during the build process by a script. Other artifacts such as the GeoIP database are stored in GCS and referenced as parameters.

We should consider treating schemas the same way we treat other artifacts. This decouples the build / template generation steps from the specific assets, which will make issuing dataflow updates via the update API not require new templates be generated for new schemas. Probably we'd want to make the schemas a zip bundle (you could even hit github directly), and make it a required parameter for the decoder job.

From discussion last week @jklukas  didn't have any objection, but I may have oversimplified things and there might be other considerations.

jklukas:
Closes #398

This new approach is consistent with how we pull in the geo database and how
we run tests for it. Instead of pulling from GitHub at maven build time,
developers are instructed to run a script to pull down an archive locally.
Then, we rely on Beam's Filesystems library to transparently handle a local
or GCS path.

We also use this opportunity to factor out all the code for parsing schemas to a separate module. So we get to remove a lot of code from ParsePayload.


mreid-moz:
cc @fbertsch and @acmiyaguchi re: schema handling

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
```diff
@@ -0,0 +1,145 @@
...
+        if (!i.canReadEntryData(entry)) {
+          LOG.warn("Unable to read tar file entry: " + entry.getName());
+          continue;
+        }
+        String[] components = entry.getName().split("/");
+        if (components.length > 2 && "schemas".equals(components[1])) {
+          String name = String.join("/", Arrays.copyOfRange(components, 2, components.length));
+          if (entry.isDirectory()) {
+            tempDirs.add(name);
+          } else if (name.endsWith(".json")) {
```
relud: I think that @acmiyaguchi is going to be outputting `bigquery` schemas as `.bigquery.json` files, so this may cause issues when that is merged.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
```diff
@@ -0,0 +1,145 @@
...
+    this.schemasLocation = schemasLocation;
+  }
+
+  @VisibleForTesting
+  public int numLoadedSchemas() {
+    ensureSchemasLoaded();
+    return schemas.size();
+  }
+
+  private void loadAllSchemas() throws SchemaNotFoundException, IOException {
```
relud: I don't see where `SchemaNotFoundException` can be thrown

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -38,6 +38,13 @@
 
   void setGeoCityFilter(ValueProvider<String> value);
 
+  @Description("Path (local or gs://) to a .tar.gz file containing json schemas; the expected"
+      + " format is the output of GitHub's archive endpoint for mozilla-pipeline-schemas:"
+      + " https://github.com/mozilla-services/mozilla-pipeline-schemas/archive/dev.tar.gz")
```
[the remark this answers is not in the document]
whd: This will be handled by the ops infra more or less analogously to how it's handled now i.e. the schemas will be recompiled before the ops artifact is generated (we should do this follow-up at any rate of course).

PR #406, opened 2019-01-30
Update python linting libs

jklukas:
Closes #388 #400 #401


jklukas:
This did not work. Still getting a pyflakes error.

PR #426, opened 2019-02-11
Truncate error payloads too large for Pubsub

### LINKED ISSUE #416: Make decoder error output pubsub-friendly
The plan is to truncate payload / attributes to within pubsub limitations and keep a counter for such messages. As to whether we drop the errors entirely or truncate oversized payloads, it probably won't make much of a difference from an analysis or reprocessing perspective.

The plan for error topology is to have decoder errors be written back to pubsub, but sink errors be written directly to GCS (for now, anyway).

jklukas:
Closes #416


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/CompressPayload.java
```diff
@@ -24,38 +27,59 @@
...
+  }
+
+  @VisibleForTesting
+  PubsubMessage compress(PubsubMessage message) {
+    message = PubsubConstraints.ensureNonNull(message);
+    byte[] compressedBytes = compress(message.getPayload(), compression.get());
+    if (compressedBytes.length > maxCompressedBytes) {
+      byte[] truncated = Arrays.copyOfRange(message.getPayload(), 0, maxCompressedBytes);
+      truncationCounter.inc();
+      return compress(new PubsubMessage(truncated, message.getAttributeMap()));
```
relud: why are we doing this instead of?
```suggestion
      return new PubsubMessage(compress(truncated, compression.get()), message.getAttributeMap());
```

PR #444, opened 2019-02-19
Test edge without docker-compose for better circleci caching

relud:
still use docker-compose for releases


ingestion-edge/tests/unit/conftest.py
```diff
@@ -26,6 +26,16 @@ def client() -> PublisherClient:
     return client
 
 
+@pytest.fixture(autouse=True)
+def bad_pubsub():
+    if "PUBSUB_EMULATOR_HOST" not in os.environ:
+        os.environ["PUBSUB_EMULATOR_HOST"] = "."
```
relud: this allows `PublisherClient` to be instantiated and fail if it's actually used (which in unit tests it isn't)

ingestion-edge/docker-compose.circleci.yml
```diff
@@ -4,6 +4,23 @@ version: '3'
...
+    command:
+    - --black
+    - --flake8
+    - --docstyle
+    - --mypy
+    - --mypy-ignore-missing-imports
+    - --ignore=tests/load
+    - .
+    - tests/integration
+    - --server=http://web:8000
```
relud: entrypoint is gone because we no longer upload coverage from here

ingestion-edge/constraints.txt
```diff
@@ -53,5 +53,5 @@ urllib3==1.24.1 --hash=sha256:61bf29cada3fc2fbefad4fdf059ea4bd1b4a86d2b6d15e1c7c
...
...41f6a14592d2908 --hash=sha256:6b2e03d69afa8d20253455e67b64de1a82ff8612db105113cccec35d3f8429f0 --hash=sha256:79ca7cdda7ad4e3663ea3c43bfa8637fc5d5604c7737f19a8964781abbd1148d --hash=sha256:7fd2dd9a856f72e6ed06f82facfce01d119b88457cd4b47b7ae501e8e11eba9c --hash=sha256:82c0354ac39379d836719a77ee360ef865377aa6fdead87909d50248d0f05f4d --hash=sha256:8f3b956d11c5b301206382726210dc1d3bee1a9ccf7aadf895aaf31f71c3716c --hash=sha256:91ec98640220ae05b34b79ee88abf27f97ef7c61cf525eec57ea8fcea9f7dddb --hash=sha256:952be9540d83dba815569d5cb5f31708801e0bbfc3a8c5aef1890b57ed7e58bf --hash=sha256:99ac266af38ba1b1fe13975aea01ac0e14bb5f3a3200d2c69f05385768b8568e --hash=sha256:9fa122e7adb24232247f8a89f2d9070bf64b7869daf93ac5e19546b409e47e96 --hash=sha256:a0873eadc4b8ca93e2e848d490809e0123eea154aa44ecd0109c4d0171869584 --hash=sha256:cb998bd4d93af46b8b49ecf5a72c0a98e5cc6d57fdca6527ba78ad89d6606484 --hash=sha256:e02e57346f6a68523e3c43bbdf35dde5c440318d1f827208ae455f6a2ace446d --hash=sha256:e79a5a896bcee7fff24a788d72e5c69f13e61369d055f28113e71945a7eb1559 --hash=sha256:ee55eb6bcf23ecc975e6b47c127c201b913598f38b6a300075f84eeef2d3baff --hash=sha256:f1414e6cbcea8d22843e7eafdfdfae3dd1aba41d1945f6ca66e4806c07c4f454
```
relud: added missing osx hashes for two dependencies, so that `pip install -r` works there

ingestion-edge/README.md
```diff
@@ -84,82 +90,86 @@ environment variables:
...
 ```
 
 ### Integration Tests
 
-Update dependencies installed in the container as needed
-
-```bash
-docker-compose build
-```
-
```
relud: removed comment indicating builds because that's already covered in the build section

relud:
also noteworthy: this fixes codecov integration for edge so that filenames match git repo

ingestion-edge/docker-compose.circleci.yml
jklukas: I'm a bit worried that these pytest configuration flags are duplicated in many places now, so seems like there's a good likelihood they'll drift apart from one another. Could we pull most of this configuration into a `pytest.ini` or wrapper of some sort that will be consumed in both the docker-compose case and the circleci case?

relud: pulled the pytest flags into `bin/test` wrapper

relud:
> consolidate the pytest flags to be reused by the various invocations

fixed in 85946cc, requesting another review because i feel a lot has changed

ingestion-edge/bin/test
```diff
@@ -0,0 +1,98 @@
...
+! ${HELP:-false} || echo "$USAGE"
+
+# exit early on --no-pytest
+${PYTEST:-true} || exit 0
+
+if ${HELP:-false}; then
+    python3.7 -m pytest "${ARGS[@]}" | sed 1d
+else
+    exec python3.7 -m pytest "${ARGS[@]}"
+fi
```
jklukas: Did you use some sort of framework for generating `bin/test` script? It's some pretty advanced bash option parsing. I find it quite readable, but it _is_ rather complex so I have some worry about maintenance burden. If someone fresh to the repo had to add a new option, for example, it might take a while to get up to speed on how to do so.

I was imagining a thinner wrapper that would just call a single line like `pytest --black --bunch-of-other-shared-args $@` but it looks like that may not be flexible enough to meet the needs of the various invocations here.

If this is building on some existing example, I'd feel better if we linked to that prior art and were able to reference that to incorporate future updates, etc.

If this is entirely new code, I rate it a _wow_ 😮 for high cleanliness and level of documentation for a bash script.

relud: it's entirely new, but it's not my first time doing option parsing in bash 😊

relud: i'll sit on this and think of ways to reduce complexity

relud: @jklukas okay, what do you think of 24ea67e ? I need to document `VENV` and `CLEAN_RELOCATES` somewhere, but i'm not sure where best to do it.

jklukas: I do think the refactor here is more maintainable. I would say VENV and `CLEAN_RELOCATES` can be documented in the README where you give example invocations of the `bin/` scripts.

PR #448, opened 2019-02-21
Generize SchemaStore for specific schema types

acmiyaguchi:
This proposed PR refactors the SchemaStore into an abstract class with two concrete implementors -- `JSONSchemaStore` and `AvroSchemaStore`. This will make it easier to handle an `AvroIO` or `ParquetIO` writer that accepts schemas in the same directory structure as `mozilla-pipeline-schemas`.

For reference, the AvroIO module documentation can be found [here](https://beam.apache.org/releases/javadoc/2.3.0/org/apache/beam/sdk/io/AvroIO.html). I'm thinking about integrating this schema store with the section on `Writing Avro files`.


acmiyaguchi:
Any ideas about the ingestion-edge image that's failing?

jklukas:
> Any ideas about the ingestion-edge image that's failing?

cc @relud

relud:
investigating...

PR #449, opened 2019-02-21
Send duplicate message metadata to error output

### LINKED ISSUE #446: Duplicates are dropped
We should update the documentation if we're not going to stream all duplicates to the error stream: https://github.com/mozilla/gcp-ingestion/blob/master/docs/decoder.md#decoding-errors.

A separate discussion would be whether to in fact keep the duplicates as errors for downstream analysis. This would allow for more comprehensive parity monitoring. Deduping is interesting to me because the implementation is completely different on the different pipelines, but I don't feel strongly that we need to keep them. In the AWS pipeline, we generated metadata-only dedup messages for cost/efficiency purposes.

jklukas:
Fixes #446 

cc @whd @trink

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
@@ -23,15 +23,19 @@ import org.apache.beam.sdk.Pipeline;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.Flatten;
+import org.apache.beam.sdk.transforms.MapElements;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
 import org.apache.beam.sdk.values.PCollectionTuple;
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
```diff
@@ -210,6 +231,12 @@ protected PubsubMessage processElement(PubsubMessage element) {
 
   ////////
 
+  private static class DuplicateIdException extends Exception {
+    DuplicateIdException() {
+      super("A message with this documentId has already been successfully processed.");
```
jklukas: Added a descriptive message.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/Deduplicate.java
```diff
@@ -105,6 +109,23 @@ public static Result of(PCollectionTuple tuple, TupleTag<PubsubMessage> outputTa
...
 
+      /**
+       * Strip the payload from duplicate messages and add them to the error collection, returning
+       * a {@link Result} of the non-duplicate output and the error collection.
+       */
+      public WithErrors.Result<PCollection<PubsubMessage>> sendDuplicateMetadataToErrors() {
+        PCollection<PubsubMessage> duplicateMetadata = tuple().get(duplicateTag())
+            .apply("DropDuplicatePayloads", MapElements //
+                .into(TypeDescriptor.of(PubsubMessage.class))
+                .via(message -> FailureMessage.of("Duplicate",
```
jklukas: Passing the literal "Duplicate" to be populated as "exception_type". Per @trink, this description is more obvious than "Deduplicate.RemoveDuplicates"

Issue #263, opened 2018-12-06
ingestion-beam-integration is flaky


relud:
looks like bigquery is failing https://circleci.com/gh/mozilla/gcp-ingestion/4094 maybe we need to add a retry?

jklukas:
https://cloud.google.com/bigquery/streaming-data-into-bigquery#dataavailability suggests that we could be seeing occasional unavailability of the streaming buffer, and we may be able to detect that by checking errors on the job.

jklukas:
This may be solved by #290, but let's watch for recurrences before closing.

jklukas:
Just saw another failure: https://circleci.com/gh/mozilla/gcp-ingestion/5376

jklukas:
Haven't seen this in a while. I think the retries are working.

PR #464, opened 2019-02-25
Add retries in BigQueryIntegrationTest to address race

jklukas:
We sometimes have tests fail due to empty query results. This is likely due
to immediately making a query after inserting rows.

Addresses #263


jklukas:
@relud ping for review

PR #470, opened 2019-02-26
Add option to write Avro to GCS

### LINKED ISSUE #24: Add options to landfill to read from and/or write to gcs in avro format

acmiyaguchi:
This PR will fix #24, ~~but is still in a draft form~~.

I'd like to document what's been done so far and the gaps in my knowledge. As I work through writing unit and integration tests in the next days, I figure that I'll figure out most of the problems that I'm facing.

---

Avro is implemented as a new output type, though it overlaps with the file writer. There are a few things I don't understand fully yet though:

From https://beam.apache.org/releases/javadoc/2.10.0/org/apache/beam/sdk/io/FileIO.Write.html
`via(Contextful<Contextful.Fn<UserT,OutputT>> outputFn, Contextful<Contextful.Fn<DestinationT,FileIO.Sink<OutputT>>> sinkFn) `

In `FileOutput`:
- `UserT = PubsubMessage`
- `OutputT = String`, representing text or json
- `DestinationT = List<String>`, representing placeholder values in the dynamic output destination

https://github.com/mozilla/gcp-ingestion/blob/7eb429193558cd1da8e6bb9b80b0986aaec0cd7d/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java#L127

This interface doesn't work for avro because of side-input per dynamic destination. However, the sink function only has access to the `DestinationT` type, which doesn't map to the SchemaStore even as side-input (in other words, `AvroSchemaStore.getSchema: List<String> -> Schema` doesn't always exist).

For `AvroOutput`:
- `UserT = PubsubMessage`
- `OutputT = GenericRecord`, representing a serialized avro record
- `DestinationT = String`, representing the path to the avro schema

The rules for the avro destination requires the schema to exist, otherwise serialization will fail. Instead of using the `AvroIO.sinkViaGenericRecords` method inside of `FileIO.Write.via`, I extend the `DynamicAvroDestinations` interface and apply it directly to the input stream and side-input.


ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -148,6 +173,125 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+
+      public GenericRecord formatRecord(PubsubMessage message) {
+        // TODO: unhandled exception
+        Schema schema = sideInput(schemaStore).getSchema(message.getAttributeMap());
+        return formatter.formatRecord(message, schema);
+      }
+
+      public Schema getSchema(String schemaPath) {
+        // TODO: unhandled exception
+        return sideInput(schemaStore).getSchema(schemaPath);
```
acmiyaguchi: I haven't figured out what to do with unhandled `SchemaNotFoundException`s, it doesn't seem like there's a way to bubble up the exceptions into the pipeline.

acmiyaguchi:
After a bit of refactoring, I figured out how to add side-input to `Contextful.Fn` to provide an `Avro.IO` sink. The writer now resembles the FileOutput class, except it passes around the pubsub attributes as the destination type. This might be troublesome because it gets used as a group by key, but it seems like the right way to map messages to a schema.

Now I'm getting this strange error in my tests. 

```
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 3 s <<< FAILURE! - in com.mozilla.telemetry.SinkAvroTest
[ERROR] testJsonToAvro(com.mozilla.telemetry.SinkAvroTest)  Time elapsed: 2.947 s  <<< ERROR!
org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.IllegalStateException: Value only available at runtime, but accessed from a non-runtime context: RuntimeValueProvider{propertyName=errorOutputPubsubCompression, default=null}
        at com.mozilla.telemetry.SinkAvroTest.testJsonToAvro(SinkAvroTest.java:53)
Caused by: java.lang.IllegalStateException: Value only available at runtime, but accessed from a non-runtime context: RuntimeValueProvider{propertyName=errorOutputPubsubCompression, default=null}
```

jklukas:
> After a bit of refactoring, I figured out how to add side-input to `Contextful.Fn` to provide an `Avro.IO` sink. The writer now resembles the FileOutput class, except it passes around the pubsub attributes as the destination type. This might be troublesome because it gets used as a group by key, but it seems like the right way to map messages to a schema.

The file output has to deal with this, too. See `DynamicPathTemplate#extractValuesFrom`. That takes in the attribute map and returns a list of just the values for the attributes needed to fill out the template.

ingestion-beam/src/test/java/com/mozilla/telemetry/SinkAvroTest.java
```diff
@@ -0,0 +1,59 @@
...
+
+  @Test
+  public void testJsonToAvro() {
+    String input = Resources.getResource("testdata/avro-message-single-doctype.ndjson").getPath();
+    String schemas = Resources.getResource("testdata/avro-schema-test.tar.gz").getPath();
+    String output = outputPath + "/out";
+
+    Sink.main(new String[] { "--inputFileFormat=json", "--inputType=file", "--input=" + input,
+        "--outputType=avro", "--output=" + output, "--outputFileCompression=UNCOMPRESSED",
+        "--schemaLocation=" + schemas });
```
jklukas: I think you're getting an value provider exception here because error output defaults to Pubsub. Try setting `--errorOutputFileCompression=UNCOMPRESSED --errorOutputType=file`.

acmiyaguchi:
The trouble using the `DynamicPathTemplate` are the placeholders are based on the output path, which can omit the document namespace, type, and version. I considered fixing the first 3 elements of the list so it can be used to retrieve schemas from the store, but accommodating the change requires some changes. It might be worth more refactoring after I write up the test suite.

relud:
> which can omit the document namespace, type, and version

if we are outputting avro and those are omitted i think we should throw an exception

ingestion-beam/src/test/java/com/mozilla/telemetry/SinkAvroTest.java
acmiyaguchi: Thanks for the pointer! This error has been particularly puzzling.

acmiyaguchi:
> > which can omit the document namespace, type, and version
> 
> if we are outputting avro and those are omitted i think we should throw an exception

Which would mean the output path would always have to be in some form of `.../${document_namespace}/${document_type}/${document_version}/...`? Taking a second look at the class, I could probably add another public method to retrieve the map for the `SchemaStore.getSchema` method. The code I have so far is probably functional though, I'm looking forward to writing the integration tests soon.

relud:
> Which would mean the output path would always have to be in some form of `.../${document_namespace}/${document_type}/${document_version}/...`?

no, it would only mean `DynamicPathTemplate.placeholderNames` would contain `"document_namespace"`, `"document_type"`, and `"document_version"`

acmiyaguchi:
After figuring out that `--errorOutputType=stdout` gives useful information about exceptions in the pipeline, I have a new set of questions.

* What is the expected behavior of the `util.Json` class? When I read an ndjson document with base64 payload into a pubsub message, I see two different behaviors. If `message` is a `PubsubMessage` with an integer payload, then `Json.asString(message)` returns a json object where the payload field is a string (presumably the base64 encoded message). If I create a new string via `new String(message.getPayload())`, I get the decoded json message. Why does the `Json.asString` method decode the payload?
  - I can decode the payload separately using the second method, but I need to figure out why the `DatumReader<GenericRecord>.read` method doesn't return a GenericRecord when reading an integer via the Avro JsonDecoder. 
     * `Caused by: java.lang.ClassCastException: java.lang.Integer cannot be cast to org.apache.avro.generic.GenericRecord`
  - I can create a helper function to do the former behavior (after understanding the expected behavior of the toString method), but it would mean the schemas from mps need to be embedded into a wrapper pubsub message schema.
* How do I propagate runtime errors within the pipeline without breaking the pipeline itself?
   - It seems like the `UncheckedIOException` will generate a new pubsub message with an attribute tagged with a wrapped `IOException`. If I throw a different runtime error like `AvroTypeException` or `SchemaNotFoundException`, the pipeline will immediately fail. Is there a way to keep the pipeline alive for certain error-cases? This would be helpful for debugging in combination with the stdout error output.

ingestion-beam/src/test/java/com/mozilla/telemetry/SinkAvroTest.java
acmiyaguchi: For the unit tests, it's more useful to set `--errorOutputType=stdout`.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -148,6 +168,109 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+        GenericRecord result;
+        try {
+          message = Json.asString(element);
+          DatumReader<GenericRecord> reader = new GenericDatumReader<>(schema);
+          Decoder decoder = DecoderFactory.get().jsonDecoder(schema, message);
+          result = reader.read(null, decoder);
+        } catch (IOException e) {
+          throw new UncheckedIOException(e);
+        } catch (AvroTypeException e) {
+          throw e;
```
acmiyaguchi: This breaks the pipeline during tests. I'd like to propagate this up with something like a `UncheckedRuntimeException`.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -148,6 +168,109 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+      this.outputPrefix = outputPrefix;
+      this.windowDuration = windowDuration;
+      this.numShards = numShards;
+      this.compression = compression;
+      this.inputType = inputType;
+      this.schemaStore = AvroSchemaStore.of(schemasLocation);
+    }
+
+    /** A functor that constructs a GenericRecord from a PubsubMessage. */
+    public class RecordFormatter implements AvroIO.RecordFormatter<PubsubMessage> {
```
acmiyaguchi: This functor does more work than I originally thought. If there are any schema incompatibilities, the `jsonDecoder` will throw an `AvroTypeException`. I'm not sure what happens to unspecified fields, but it might be possible to stuff them into a `residue` field here. Additionally, the schemas can be modified after the fact here.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
jklukas: It may be a limitation of `sinkViaGenericRecords` that it doesn't allow any facility for handling exceptions raised while formatting records. We might need to be responsible for doing more of that checking ourselves.

acmiyaguchi:
I'm starting to write some integration tests now that some of the basic unit tests are passing, but there are a few caveats to the code that should be changed. Here are some status notes. 

* Unit tests are passing, which presumably means that pubsub messages can be encoded into avro
* Unit tests are generated from a python script for ease of modification.
* The `FileIO.Write` interface doesn't allow us to catch any exceptions from what I can see. This is troubling because documents that don't exist in the schema store or do not conform to the schema will bring down the pipeline.
* I'm starting to write integration tests that generate avro schemas inline from the `schemas.tar.gz` file. I may have to limit this a subset of documents since it seems like the schema transpiler is not accepting some of the jsonschemas.

acmiyaguchi:
It looks like some of the generated avro schemas are not actually valid schemas when running the following command:

```bash
# first make sure all the dependencies exist
$ ./bin/download-schemas

# convert json schemas into avro schemas using the jsonschema_transpiler
$ ./bin/generate-avro-schemas.sh

# download and convert corresponding samples from telemetry-parquet/sanitized-landfill-sample/v3
$ ./bin/download-sampled-landfill.py

# run locally
$ ./bin/mvn compile exec:java -Dexec.args="\
    --inputFileFormat=json \
    --inputType=file \
    --input=avro-landfill-integration.ndjson  \
    --outputType=avro \
    --output=tmp/output/out \
    --schemaLocation=avro-schema.tar.gz \
    --errorOutputType=stdout \
"
```

Here's an example error:
```
INFO: Deleting temporary file /root/project/tmp/output/.temp-beam-2019-03-13_23-21-39-1/a40d6407-4813-4ade-948d-4b96c62c848e
    at java.lang.Thread.Mar 13, 2019 11:21:50 PM org.apache.beam.sdk.io.WriteFiles$WriteShardsIntoTempFilesFn processElement
INFO: Opening writer 4164672e-472a-464b-9b96-f7567ddf22b6 for window [-290308-12-21T20:00:00.000Z..-290308-12-21T20:10:00.000Z) pane PaneInfo{isFirst=true, isLast=true, timing=ON_TIME, index=0, onTimeIndex=0} destination {document_namespace=telemetry, document_type=heartbeat, document_version=4}
run (Thread.java:834)
Caused by: java.lang.NullPointerException
    at org.apache.beam.sdk.io.FileIO$Write$ViaFileBasedSink$1$1.write (FileIO.java:1327)
    at org.apache.beam.sdk.io.WriteFiles.writeOrClose (WriteFiles.java:549)
    at org.apache.beam.sdk.io.WriteFiles.access$1000 (WriteFiles.java:111)
    at org.apache.beam.sdk.io.WriteFiles$WriteShardsIntoTempFilesFn.processElement (WriteFiles.java:717)
Mar 13, 2019 11:21:50 PM org.apache.beam.sdk.io.FileBasedSink$Writer open
SEVERE: Beginning write to  failed, closing channel.
org.apache.avro.SchemaParseException: "record" is not a defined name. The type of the "application" field must be a defined name or a {"type": ...} expression.
        at org.apache.avro.Schema.parse(Schema.java:1265)
        at org.apache.avro.Schema$Parser.parse(Schema.java:1032)
        at org.apache.avro.Schema$Parser.parse(Schema.java:1004)
        at com.mozilla.telemetry.schemas.AvroSchemaStore.loadSchemaFromArchive(AvroSchemaStore.java:31)
        at com.mozilla.telemetry.schemas.AvroSchemaStore.loadSchemaFromArchive(AvroSchemaStore.java:12)
        at com.mozilla.telemetry.schemas.SchemaStore.loadAllSchemas(SchemaStore.java:126)
        at com.mozilla.telemetry.schemas.SchemaStore.ensureSchemasLoaded(SchemaStore.java:138)
        at com.mozilla.telemetry.schemas.SchemaStore.getSchema(SchemaStore.java:44)
````

File: ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java [hunk 5]
```diff
@@ -148,6 +166,98 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+    public WithErrors.Result<PDone> expand(PCollection<PubsubMessage> input) {
+      final PCollectionView<AvroSchemaStore> schemaSideInput = input.getPipeline()
+          .apply(Create.of(schemaStore)).apply(View.asSingleton());
+
+      ValueProvider<DynamicPathTemplate> pathTemplate = NestedValueProvider.of(outputPrefix,
+          DynamicPathTemplate::new);
+      ValueProvider<String> staticPrefix = NestedValueProvider.of(pathTemplate,
+          value -> value.staticPrefix);
+
+      ParDo.SingleOutput<PubsubMessage, KV<Map<String, String>, GenericRecord>> intoGenericRecord = ParDo
```
acmiyaguchi:
```
java.lang.IllegalStateException: Unable to return a default Coder for Write.AvroOutput/PubsubMessageToGenericRecord/ParMultiDo(Anonymous).output [PCollection]. Correct one of the following root causes:
  No Coder has been manually specified;  you may do so using .setCoder().
  Inferring a Coder from the CoderRegistry failed: Cannot provide coder for parameterized type org.apache.beam.sdk.values.KV<java.util.Map<java.lang.String, java.lang.String>, org.apache.avro.generic.GenericRecord>: Unable to provide a Coder for org.apache.avro.generic.GenericRecord.
  Building a Coder using a registered CoderProvider failed.
  See suppressed exceptions for detailed failures.
  Using the default output Coder from the producing PTransform failed: PTransform.getOutputCoder called.
```

jklukas: I hadn't totally thought through this, but looks like this is going to be a significant problem. It looks like AvroCoder requires a single schema be supplied. I don't think it's possible to store a PCollection containing GenericRecord of varying schemas. I believe AvroIO.Sink delays encoding to GenericRecord until after a destination has been determined, so from then on all records have the same schema and be encoded together in a single collection.

Looking at AvroIO's "Writing Data to Multiple Destinations" example, it looks like they define a custom DynamicAvroDestinations subclass that includes a hook for calling a method to encode to GenericRecord. That interface doesn't give us a way to bubble up exceptions to a separate collection.

Perhaps it's possible for us to hijack the interface to route to a failure destination if the message isn't encodable. We could have a special schema with just a single string field, and we could shove the payload in there. Our error output then would end up being avro-encoded. I'm not entirely sure this approach is feasible given the constraints of the interface.

Another possibility is that we maintain this custom ParDo, but only use it for a "dry run" of encoding to find records that are going to cause exceptions. We'd try encoding and if it passes, we'd pass through the unencoded payload string rather than the GenericRecord. Then, AvroIO.Sink would have to do the compute of encoding again, but we'd be guaranteed that it won't throw exceptions since we've already validated the avro encoding.

Another option would be to modify AvroIO itself to allow the exception handling we need and PR that code upstream. That's likely an ambitious task, though, so would only be feasible as a longer-term refactor rather than a solution for the purposes of getting this avro pipeline unblocked.

acmiyaguchi:
This passes locally for me. It looks like the testing scripts need to be modified to generate the avro files.

I am currently getting OOM errors, but I'm closer to actually getting this to work properly. 

```
Caused by: java.lang.OutOfMemoryError: Java heap space
    at java.util.Arrays.copyOf (Arrays.java:3745)
    at java.io.ByteArrayOutputStream.grow (ByteArrayOutputStream.java:120)
    at java.io.ByteArrayOutputStream.ensureCapacity (ByteArrayOutputStream.java:95)
    at java.io.ByteArrayOutputStream.write (ByteArrayOutputStream.java:156)
    at org.apache.beam.sdk.util.StreamUtils.getBytesWithoutClosing (StreamUtils.java:59)
    at org.apache.beam.sdk.coders.StringUtf8Coder.decode (StringUtf8Coder.java:96)
    at org.apache.beam.sdk.coders.StringUtf8Coder.decode (StringUtf8Coder.java:37)
    at org.apache.beam.sdk.util.CoderUtils.decodeFromSafeStream (CoderUtils.java:115)
    at org.apache.beam.sdk.util.CoderUtils.decodeFromByteArray (CoderUtils.java:98)
    at org.apache.beam.sdk.util.CoderUtils.decodeFromByteArray (CoderUtils.java:92)
    at org.apache.beam.sdk.util.CoderUtils.clone (CoderUtils.java:141)
    at org.apache.beam.sdk.util.MutationDetectors$CodedValueMutationDetector.<init> (MutationDetectors.java:113)
    at org.apache.beam.sdk.util.MutationDetectors.forValueWithCoder (MutationDetectors.java:44)
    at org.apache.beam.runners.direct.ImmutabilityCheckingBundleFactory$ImmutabilityEnforcingBundle.add (ImmutabilityCheckingBundleFactory.java:112)
    at org.apache.beam.runners.direct.BoundedReadEvaluatorFactory$BoundedReadEvaluator.processElement (BoundedReadEvaluatorFactory.java:151)
    at org.apache.beam.runners.direct.DirectTransformExecutor.processElements (DirectTransformExecutor.java:160)
    at org.apache.beam.runners.direct.DirectTransformExecutor.run (DirectTransformExecutor.java:124)
    at java.util.concurrent.Executors$RunnableAdapter.call (Executors.java:515)
    at java.util.concurrent.FutureTask.run (FutureTask.java:264)
    at java.util.concurrent.ThreadPoolExecutor.runWorker (ThreadPoolExecutor.java:1128)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run (ThreadPoolExecutor.java:628)
    at java.lang.Thread.run (Thread.java:834)
```

acmiyaguchi:
> I am currently getting OOM errors, but I'm closer to actually getting this to work properly.
> 
> ```
> Caused by: java.lang.OutOfMemoryError: Java heap space
>```

This is solved by prepending `MAVEN_OPTS="-Xmx4g"` to maven directly. The docker script needs to be modified to export variables into the container.

I can now start to fix some of the bugs that I've found while testing the integration test scripts. There are a couple of things that I want to note here. 

* It may be useful to rewrite the `download-sampled-landfill` script with proper arguments for limiting the set of documents and the number of documents per type. Otherwise the number of errors blows up significantly.
* The `GuidedJsonDecoder` is failing on a variety of documents. In particular, it's not setting default values correctly: 
```json
{
  "attributeMap": {
    "error_message": "org.apache.avro.AvroTypeException: Expected field name not found: campaign",
    "error_type": "com.mozilla.telemetry.io.Write$AvroOutput$3@734fccac",
    "document_namespace": "telemetry",
    "exception_class": "org.apache.avro.AvroTypeException",
    "stack_trace": "[com.mozilla.telemetry.avro.GuidedJsonDecoder.doAction(GuidedJsonDecoder.java:204), org.apache.avro.io.parsing.Parser.advance(Parser.java:88), com.mozilla.telemetry.avro.GuidedJsonDecoder.readIndex(GuidedJsonDecoder.java:464), org.apache.avro.io.ResolvingDecoder.doAction(ResolvingDecoder.java:290), org.apache.avro.io.parsing.Parser.advance(Parser.java:88), org.apache.avro.io.ResolvingDecoder.readIndex(ResolvingDecoder.java:267), org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:179), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:153), org.apache.avro.generic.GenericDatumReader.readField(GenericDatumReader.java:232), org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:222), org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:175), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:153), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:145...",
    "document_version": "10",
    "document_type": "core"
  },
  "payload": "..."
}
```
* It may be useful to extend MapWithErrors to use side inputs. On the other hand, I found it much easier to define the ParDo transform inline using side inputs and output tags.
* A single broken schema in the SchemaStore will cause the pipeline to fail by continuously writing to the error-stream. In this case, I was using an old version of the jsonschema-transpiler that did not strip `$schema` fields in records.

acmiyaguchi:
This works successfully after making the patches to the Avro JsonDecoder! No errors are encountered once the data is validated. 

```bash
#!/bin/bash

# Pick a bucket to store files in
BUCKET="gs://$(gcloud config get-value project)"

# create a test input file
gsutil cp avro-landfill-integration.ndjson $BUCKET/input.json
gsutil cp avro-schema.tar.gz $BUCKET/schema.tar.gz

# Set credentials; beam is not able to use gcloud credentials
export GOOGLE_APPLICATION_CREDENTIALS="$HOME/test-avro-ingest-credentials.json"

# consume messages from the test file, decode and re-encode them, and write to a bucket
    --runner=Dataflow \
    --inputFileFormat=json \
    --inputType=file \
    --input=$BUCKET/input.json  \
    --outputType=avro \
    --output=$BUCKET/output/\${document_namespace:-NONE}/\${document_type:-NONE}/\${document_version:-0}/out \
    --schemaLocation=$BUCKET/schema.tar.gz \
    --errorOutputType=file \
"
```

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
acmiyaguchi: I've solved this by creating a DoFn that handles side-input and tuple-tags for a success and error PCollection.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
acmiyaguchi: Most of the logic here was moved into a custom JsonDecoder in #506.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
acmiyaguchi: I went with the first approach of stuffing the avro-encoded data into a Pubsub message. I think this is better than the second approach of a dry run, because writing and reading the binary data is probably faster than decoding the JSON payload twice.

I would be interested in seeing the options for adding error-handling closer to the FileIO writer. In a similar vein, I think another useful feature would be dynamic Parquet outputs, since I've already worked around the error handling in this PR.

acmiyaguchi:
I've cleaned up this branch in preparation for final review. The backup of the branch history can be found in my repository as `avro-writer-history-backup`.

There a few things that still need to be done:

[ ] Ensure tests are passing
[ ] Modify `generate-avro-test-resources` script to update tar.gz only when uncompressed contents change
[ ] Refactor `generate-avro-test-resources` into two commits with generated schemas and generate test data
[ ] Create a separate SchemaStore PR for fixing bug with AvroSchemaStore implementation
[ ] Miscellaneous editing
[ ] README updates (?)

@jklukas I think this is ready for feedback -- I can incorporate changes as I rework the pieces above.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -148,6 +169,104 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+      };
+      final TupleTag<PubsubMessage> errorTag = new TupleTag<PubsubMessage>() {
+      };
+
+      ValueProvider<DynamicPathTemplate> pathTemplate = NestedValueProvider.of(outputPrefix,
+          DynamicPathTemplate::new);
+      ValueProvider<String> staticPrefix = NestedValueProvider.of(pathTemplate,
+          value -> value.staticPrefix);
+
+      ParDo.MultiOutput<PubsubMessage, PubsubMessage> intoGenericRecord = ParDo
```
acmiyaguchi: This should be renamed to something like `encodeAvro`

ingestion-beam/src/main/java/com/mozilla/telemetry/options/OutputType.java
```diff
@@ -41,6 +42,16 @@ public Write write(SinkOptions.Parsed options) {
     }
   },
 
+  avro {
+
+    /** Return a PTransform that writes to local or remote files. */
+    public Write write(SinkOptions.Parsed options) {
+      return new AvroOutput(options.getOutput(), options.getParsedWindowDuration(),
+          options.getOutputNumShards(), options.getOutputFileCompression(), options.getInputType(),
+          options.getSchemaLocation());
```
acmiyaguchi: This may need to be updated to include the recent schema aliasing changes.

ingestion-beam/src/test/java/com/mozilla/telemetry/SinkAvroTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    Sink.main(new String[] { "--inputFileFormat=json", "--inputType=file", "--input=" + input,
+        "--outputType=avro", "--output=" + output, "--outputFileCompression=UNCOMPRESSED",
+        "--schemaLocation=" + schemas, "--errorOutputFileCompression=UNCOMPRESSED",
+        "--errorOutputType=file", "--errorOutput=" + errorOutput, });
+
+    assertThat("error count", getPrefixFileCount(outputPath, "err"), Matchers.greaterThan(0L));
+    assertThat("foo output count", getPrefixFileCount(outputPath + "/foo", "out"),
+        Matchers.is(-1L));
+  }
```
acmiyaguchi: Add a test asserting the correctness of a record. Additionally, a test verifying the error output is including the correct exceptions. These tests were written before anything worked, so these tests are based on the expected file outputs.

ingestion-beam/bin/download-sampled-landfill.py
```diff
@@ -0,0 +1,105 @@
...
+    """Given a directory path to a json schema in the mps directory, generate
+    the fully qualified name in the form `{namespace}.{doctype}.{docver}`."""
+    elements = path.split("/")
+    doctype, docver = elements[-1].split(".")[:-2]
+    namespace = elements[-3]
+    return f"{namespace}.{doctype}.{docver}"
+
+
+def load_schemas(path):
+    """return a dictionary containing "{namespace}.{doctype}.{docver}" to validator"""
```
jklukas:
```suggestion
    """Return a dictionary containing "{namespace}.{doctype}.{docver}" to validator"""
```

ingestion-beam/bin/download-sampled-landfill.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
```
jklukas: Remove .py extension to match other executables in this directory

jklukas: Please add a module docstring explaining the purpose and usage of this executable.

ingestion-beam/bin/generate-avro-schemas.sh
```diff
@@ -0,0 +1,69 @@
...
+cd "$(dirname "$0")/.." || exit
+
+if [ ! -x  "$(command -v jsonschema-transpiler)" ]; then
+    echo "jsonschema-transpiler is not installed"
+    echo "Run 'cargo install --git https://github.com/acmiyaguchi/jsonschema-transpiler.git --branch dev'"
+    exit 1
+fi
+
+if [[ ! -f schemas.tar.gz ]]; then
+    echo "Run 'bin/download-schemas'"
```
jklukas: I still have in my mind that we should be generating avro and bq schemas directly in mozilla-pipeline-schemas. It would make it very easy to inspect the source schemas via the normal GitHub UI, and it also would mean that the avro and bq schemas would get packaged up in the mps tarball rather than needing to pull in a separate tarball for avro schemas or to generate a tarball here.

Has a decision been made to run the schema transpiler elsewhere for the production pipeline?

ingestion-beam/src/main/java/com/mozilla/telemetry/avro/BinaryRecordFormatter.java
```diff
@@ -0,0 +1,39 @@
...
+ * Decodes an incoming PubsubMessage that contains an avro encoded payload.
+ */
+public class BinaryRecordFormatter implements RecordFormatter<PubsubMessage> {
+
+  static DecoderFactory factory = new DecoderFactory();
+
+  @Override
+  public GenericRecord formatRecord(PubsubMessage element, Schema schema) {
+    InputStream in = new ByteArrayInputStream(element.getPayload());
+    DatumReader<GenericRecord> reader = new GenericDatumReader<>(schema);
```
jklukas: Do you know how expensive it is to contruct this reader? Should we maintain a local cache of readers as a private field?

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
jklukas: I'd like to see this factored out to a PTransform that extends MapElementsWithErrors. That will hide most of the boilerplate for handling the exception (unless there's additional nuance here I missed).

ingestion-beam/bin/generate-avro-schemas.sh
acmiyaguchi: I opened an issue in mozilla-services/mozilla-pipeline-schemas#306, which includes a link to a branch where all the generated schemas can be browsed. The trouble is the increase in the number of dependencies. Adding the transpiler and python3 dependencies would make it even hard to contribute to the schema repo. I'm still mulling over the best way to keep it cross-platform and accessible.

@fbertsch and I have been discussing an Airflow DAG for generating the schema artifacts that also integrate probe information. Reading tarballs in GCS from this output might be another alternative to checking everything into mps.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
acmiyaguchi: I made a note in https://github.com/mozilla/gcp-ingestion/pull/470#issuecomment-478786548 about this ParDo, since I originally tried to write it as a PTransform. I had difficulty including side-input into the PTransform, which is why I use the ParDo here.

I could probably refactor this so the PTransform contains a static schema store that it uses independently from the one in this scope. Would that be preferable?

ingestion-beam/src/main/java/com/mozilla/telemetry/avro/BinaryRecordFormatter.java
acmiyaguchi: I don't know how expensive this is to construct, it might be worth profiling this somehow. I imagine most of the time will be spent on decoding the binary file though.

ingestion-beam/bin/generate-avro-schemas.sh
jklukas: Thanks for starting that discussion. I've jumped in there with my thoughts on the benefits of checking in the generated avro and bq schemas.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
jklukas: I don't feel strongly about it. The side input detail seems like sufficient motivation to define it directly as a ParDo here. I'd be happy if you just add a comment indicating that we're not using  MapElementsWithErrors on account of the side input.

ingestion-beam/src/main/java/com/mozilla/telemetry/avro/BinaryRecordFormatter.java
jklukas: Don't block this PR on the performance investigation. We can observe performance of the job and look more into it if it's looking like this stage of the pipeline is a hot spot.

acmiyaguchi:
This is ready for review, albeit missing two small unit tests that I've noted. 

The branch in `bin/download-schemas` needs to be rebased on a more recent version of mps. As a temporary fix, you may replace `jklukas` with `acmiyaguchi` to get fixes necessary for transpiling all of the schemas.

To verify that the new sink works as expected:

```
$ bin/download-schemas
$ bin/generate-avro-schemas
$ pip3 install --user boto3 python-rapidjson
$ bin/download-sampled-landfill
```
Finally, run the following command:
```
MAVEN_OPTS="-Xmx4g" mvn compile exec:java -Dexec.args="\
    --inputFileFormat=json \
    --inputType=file \
    --input=avro-landfill-integration.ndjson  \
    --outputType=avro \
    --output=tmp/output/\${document_namespace:-NONE}/\${document_type:-NONE}/\${document_version:-0}/out \
    --schemaLocation=avro-schema.tar.gz \
    --errorOutputType=file \
   --errorOutput=tmp/error/
```

```
$ tree tmp/
tmp/
└── output
    ├── activity-stream
    │   └── impression-stats
    │       └── 1
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00000-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00001-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00002-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00003-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00004-of-00006.avro.gz
    │           └── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00005-of-00006.avro.gz
    ├── coverage
    │   └── coverage
    │       └── 1
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00000-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00001-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00002-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00003-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00004-of-00006.avro.gz
    │           └── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00005-of-00006.avro.gz
    ├── eng-workflow
    │   └── build
    │       └── 1
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00000-of-00006.avro.gz
    │           ├── out--290308-12-21T20:00:00.000Z--290308-12-21T20:10:00.000Z-00001-of-00006.avro.gz
...
```

ingestion-beam/bin/generate-avro-schemas
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
+# Generate a new archive with avro schemas from the mozilla-pipeline-schema
```
jklukas:
```suggestion

# Generate a new archive with avro schemas from the mozilla-pipeline-schema
```

acmiyaguchi:
> Since you've pioneered a technique for generating a tarball from schemas checked into the repo, perhaps it would be better to switch to that technique here as well.

I don't think the actual branch matters, as long as it's on [https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/ba09bf7f0de547fcd0719e960f6eb5402d33a4f4](ba09bf7) or higher. Is there a reason for not using the mozilla-services/mozilla-pipeline-schemas:dev repo directly?

jklukas:
> Is there a reason for not using the mozilla-services/mozilla-pipeline-schemas:dev repo directly?

The branch on my fork added a test/test schema with no fields. That schema is referenced in tests that include schema validation. That's the only difference.

acmiyaguchi:
I'm going to merge this in as is, with confidence that this will work based on the integration tests in [this earlier comment](https://github.com/mozilla/gcp-ingestion/pull/470/files#issuecomment-481063633).

I'll leave some of the issues in this thread to be fixed in smaller, iterative PRs going forward.

Issue #503, opened 2019-03-26
Add PTransform for converting PubsubMessage into GenericRecord


acmiyaguchi:
A [GenericRecord](http://avro.apache.org/docs/1.8.2/api/java/org/apache/avro/generic/GenericRecord.html?is-external=true) represents a deserialized Avro record. This structure can be written to disk via `AvroIO` or `ParquetIO` using an Avro schema fetched from `AvroSchemaStore`. 

This transform groups by namespace, type, and version and applies a format function that takes a PubsubMessage and Schema and generates a GenericRecord. The format function may throw when the fully qualified schema name is not found or when the data reader finds an unexpected or missing element in the incoming document.

More concretely, the PTransform needs to implement the `org.apache.beam.sdk.io.AvroIO.RecordFormatter` interface. Due to [AVRO-1582](https://issues.apache.org/jira/browse/AVRO-1582), we must implement a custom JSON decoder for the `GenericDatumReader` function. In short, the included JSON decoder in the java Avro library expects documents in the format`{"<NAME>": {"<TYPE>": <VALUE>}}` instead of the standard `{"<NAME>": <VALUE>}`

This transform also handles a few other things:
* Casting JSON elements into strings when specified by the schema (due to incompatible typing or incomplete information)
* Renaming field names (names should match the regex `^[a-zA-Z_][a-ZA-Z0-9_]*`).

acmiyaguchi:
I'm going to mark this as done, since the custom implementation of the JsonDecoder in #506 implements most of this logic.

PR #506, opened 2019-03-28
Implement formatter for decoding PubsubMessage payloads into GenericRecords

acmiyaguchi:
This address a significant portion of #503.

This formatter will convert a PubsubMessage into a GenericRecord, which can then be written out to Avro or Parquet using the corresponding Sink. The decoder does the following things:

* Decode arbitrary JSON documents given an appropriate Avro schema
* Uses the schema to serialize sub-documents into strings when appropriate
* Renames field names to adhere to BigQuery column name rules

There are many unused operations that are unimplemented in this class. There are a few Avro types that could prove useful in the future, but the current schemas generated by the [jsonschema-transpiler](https://github.com/acmiyaguchi/jsonschema-transpiler) uses a subset.


ingestion-beam/src/main/java/com/mozilla/telemetry/avro/GuidedJsonDecoder.java
```diff
@@ -0,0 +1,410 @@
...
+import org.apache.avro.io.parsing.JsonGrammarGenerator;
+import org.apache.avro.io.parsing.Parser;
+import org.apache.avro.io.parsing.Symbol;
+import org.apache.avro.util.Utf8;
+
+/**
+ * A custom JSON decoder based on the Parsing JsonDecoder in the
+ * `org.apache.avro.io.JsonDecoder` package.
+ *
+ * <p>The default JsonDecoder (as of Avro 1.8.2) does support ingestion of JSON
```
jklukas:
```suggestion
 * <p>The default JsonDecoder (as of Avro 1.8.2) doesn't support ingestion of JSON
```

ingestion-beam/src/main/java/com/mozilla/telemetry/avro/GuidedJsonDecoder.java
```diff
@@ -0,0 +1,410 @@
...
+ * points from integers and records from maps. To support BigQuery, the decoder
+ * uses unions to enforce required fields and will fail on variant-types.
+ * Additionally, field names are renamed to match the column-name rules for
+ * BigQuery tables. The source JSON schemas can be underspecified or overly
+ * complex. In such cases, the transpiler will cast fields into strings. This
+ * encoder supports encoding tree-nodes into strings when guided by the provided
+ * schema.
+ *
+ * <p>This class shares some code from the JsonDecoder class. The original class
+ * cannot be extended outside of the Avro package. The algorithm for buffering
```
jklukas: It's unsavory, but we've done it before: we can define this class inside the avro package so that it's able to inherit from avro's JsonDecoder. If that removes a significant amount of copied code here, I'd support it. As long as we document in the javadoc why we're doing what we're doing.

acmiyaguchi: Are there any pointers to defining a class within an external package? On another hand,I re-read through the variable scoping in the JsonDecoder class and found out that the JsonParser `in` that drives the decoding is a private variable. This means there's no way around this other than forking or re-implementing from the ParsingDecoder class.

Extending the class would reduce the code to`doAction`, `readString`, and `readIndex`. The `doAction` method needs to be re-implemented because of the custom renaming rule for records. One implementation outside of apache that extends the JsonDecoder is [PIMCO/avro](https://github.com/PIMCO/avro). They support a few custom extensions like decimal types. It looks like their fork changes the private variable into a protected one, which comes at the cost of maintaining a fork of the project.

jklukas: > Are there any pointers to defining a class within an external package?

It's rather silly that it works this way, but if you change the package declaration at the top to `org.apache.avro` (or whichever subpackage you need) and move the file to:

    ingestion-beam/src/main/java/org/apache/avro/GuidedJsonDecoder.java 

Then it gets all the privileges as if it had been defined in the avro source code.

jklukas: Assuming that tricks to inherit won't actually work due to the private `in`, can you document in the source which methods are modified? Perhaps every modified gets javadoc that specifies it's our code?

acmiyaguchi: I'll annotate the methods that contain mostly custom code. I did rewrite using the original as reference, so there are quite a few cosmetic changes to variable names and specific error handling.

acmiyaguchi:
Thanks for taking a look. I'm curious what the performance will look like across our data -- this is the first time I've used a streaming parser to manipulate JSON at such a low level.

PR #507, opened 2019-03-28
Use X-Pipeline-Proxy as submission_timestamp

### LINKED ISSUE #496: Preserve edge timestamp across the tee
If I understand correctly we aren't using the value of `X-Pipeline-Proxy` for anything, so in the tee we could set the value of that header to the `submission_timestamp` from the AWS edge (or maybe something equivalent) and in the decoder we could use that value to in place of the `submission_timestamp` set by ingestion-edge. Maybe the decoder would also monitor the lag between the two.

This should allow us to eliminate timestamp lag, which should make it easier to assess the gcp-ingestion pipeline and make a production data source.

cc @whd @jklukas

relud:
fixes #496

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
@@ -1,0 +1,99 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry.decoder;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.mozilla.telemetry.transforms.PubsubConstraints;
+import java.text.ParseException;
+import java.text.SimpleDateFormat;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.metrics.Counter;
+import org.apache.beam.sdk.metrics.Metrics;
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
```diff
@@ -0,0 +1,99 @@
...
+  private static final String X_FORWARDED_FOR = "x_forwarded_for";
+  private static final String X_PIPELINE_PROXY = "x_pipeline_proxy";
+
+  private static final Fn FN = new Fn();
+  private static final ParseProxy INSTANCE = new ParseProxy();
+
+  @VisibleForTesting
+  public static class Fn extends SimpleFunction<PubsubMessage, PubsubMessage> {
+
+    private final Counter countPipelineProxy = Metrics.counter(Fn.class, "count_pipeline_proxy");
```
jklukas:
```suggestion
    private final Counter countPipelineProxy = Metrics.counter(Fn.class, "pipeline_proxy");
```

I don't know why I included the "count_" prefix on this in the first place.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
```diff
@@ -0,0 +1,99 @@
...
+        boolean proxyIsTimestamp = true;
+        try {
+          TIMESTAMP_FORMAT.parse(xpp);
+        } catch (ParseException ignore) {
+          proxyIsTimestamp = false;
+        }
+
+        if (proxyIsTimestamp) {
+          // Rename submission timestamp to proxy timestamp
+          attributes.put(PROXY_TIMESTAMP, attributes.get(SUBMISSION_TIMESTAMP));
```
jklukas: Maintaining the proxy timestamp is comforting 👍

Issue #424, opened 2019-02-08
Implement namespace/docType aliases


fbertsch:
For Glean, we have multiple namespaces that are all really using the same schema. We don't want to duplicate this (exponentially - every application, ping, and version!) in mozilla-pipeline-schemas. Instead we'd like to alias namespaces and docTypes.

This would work by keeping a file in GCS of aliases. For example, `org-mozilla-reference-browser/baseline` would be an alias for `glean/glean`. When doing schema validation, the JSON schema would be found using `glean/glean`. Otherwise, the pipeline would use the other namespace and docType for other transformations.

We could additionally consider adding the aliased namespace and docType as metadata to the ping.

jklukas:
I assume we want to maintain different tables in BQ for each of these? It is only the schema validation step where we want to resolve the namespace/docType to a particular schema?

Let me push back a moment on this idea. Do we have a good sense of how many application/ping/version combinations will be introduced with Glean? Are we sure it warrants introducing a new concept (breaking the 1:1 mapping of namespace/docType to schema)?

With the existing structure of mozilla-pipeline-schemas, each of these namespace/docType/version combos need only be a trivial file with a reference to the `include` for the glean structure. It's straightforward to build a parser for that repository that has full knowledge of allowed docTypes and the associated schemas. I worry this would significantly complicate the ability for humans to comprehend and for computers to parse that repo.

fbertsch:
Thanks for the feedback @jklukas! I'll answer inline.

> I assume we want to maintain different tables in BQ for each of these?
Right, definitely. This mapping is solely for validation.

> Do we have a good sense of how many application/ping/version combinations will be introduced with Glean? Are we sure it warrants introducing a new concept (breaking the 1:1 mapping of namespace/docType to schema)?
So far, we have 3 applications that implement glean, and it doesn't even have any data yet! Given the 4 additional mobile browsers, lockbox, and VR; we're looking at 10+ applications.

For pings, we have: baseline, metrics, events, crash, and activation. Additionally each application can send custom pings, so e.g. a sync ping. In total, >60 pings, each with the same schema.

> With the existing structure of mozilla-pipeline-schemas, each of these namespace/docType/version combos need only be a trivial file with a reference to the `include` for the glean structure

[...]

PR #508, opened 2019-03-29
Add namespace/docType aliasing

akkomar:
Adds support for namespace/docType aliases (https://github.com/mozilla/gcp-ingestion/issues/424).

Aliasing configuration is stored in json config file, to be loaded from GCS. Decoder should be redeployed on configuration changes, likewise on schema changes.
If configuration file is present, schemas will be aliased in the `SchemaStore`, transparently to the rest of the pipeline.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -31,8 +31,9 @@ import org.json.JSONObject;
  */
 public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
 
-  public static ParsePayload of(ValueProvider<String> schemasLocation) {
-    return new ParsePayload(schemasLocation);
+  public static ParsePayload of(ValueProvider<String> schemasLocation,
+      ValueProvider<String> schemaAliasesLocation) {
+    return new ParsePayload(schemasLocation, schemaAliasesLocation);
   }
 
   ////////
@@ -43,12 +44,15 @@ public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<Pubs
       "json_validate_millis");
 
   private final ValueProvider<String> schemasLocation;
+  private final ValueProvider<String> schemaAliasesLocation;
 
   private transient Validator validator;
   private transient JSONSchemaStore schemaStore;
 
-  private ParsePayload(ValueProvider<String> schemasLocation) {
+  private ParsePayload(ValueProvider<String> schemasLocation,
+      ValueProvider<String> schemaAliasesLocation) {
     this.schemasLocation = schemasLocation;
+    this.schemaAliasesLocation = schemaAliasesLocation;
   }
 
   @Override
@@ -58,7 +62,7 @@ public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<Pubs
     Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
 
     if (schemaStore == null) {
-      schemaStore = JSONSchemaStore.of(schemasLocation);
+      schemaStore = JSONSchemaStore.of(schemasLocation, schemaAliasesLocation);
     }
 
     final int submissionBytes = message.getPayload().length;
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/JSONSchemaStore.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/JSONSchemaStore.java
@@ -16,13 +16,18 @@ import org.json.JSONObject;
 @SuppressWarnings("checkstyle:AbbreviationAsWordInName")
 public class JSONSchemaStore extends SchemaStore<Schema> {
 
-  /** Returns a SchemaStore based on the contents of the archive at schemasLocation. */
-  public static JSONSchemaStore of(ValueProvider<String> schemasLocation) {
-    return new JSONSchemaStore(schemasLocation);
+  /**
+   * Returns a SchemaStore based on the contents of the archive at schemasLocation
+   * with additional schemas aliased according to configuration.
+   */
+  public static JSONSchemaStore of(ValueProvider<String> schemasLocation,
+      ValueProvider<String> schemaAliasesLocation) {
+    return new JSONSchemaStore(schemasLocation, schemaAliasesLocation);
   }
 
-  protected JSONSchemaStore(ValueProvider<String> schemasLocation) {
-    super(schemasLocation);
+  protected JSONSchemaStore(ValueProvider<String> schemasLocation,
+      ValueProvider<String> schemaAliasesLocation) {
+    super(schemasLocation, schemaAliasesLocation);
   }
 
   @Override
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry.schemas;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class SchemaAlias {
+
+  @JsonCreator
+  static SchemaAlias create(@JsonProperty(value = "base", required = true) String base,
+      @JsonProperty(value = "alias", required = true) String alias) {
+    return new AutoValue_SchemaAlias(base, alias);
+  }
+
+  abstract String base();
+
+  abstract String alias();
+}
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAliasingConfiguration.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAliasingConfiguration.java
@@ -1,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry.schemas;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+import java.util.List;
+
+@AutoValue
+abstract class SchemaAliasingConfiguration {
+
+  @JsonCreator
+  static SchemaAliasingConfiguration create(
+      @JsonProperty(value = "aliases", required = true) List<SchemaAlias> aliases) {
+    return new AutoValue_SchemaAliasingConfiguration(aliases);
+  }
+
+  abstract List<SchemaAlias> aliases();
+}
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
@@ -5,6 +5,10 @@
 package com.mozilla.telemetry.schemas;
 
 import avro.shaded.com.google.common.annotations.VisibleForTesting;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.base.Strings;
+import com.google.common.collect.ArrayListMultimap;
+import com.google.common.collect.Multimap;
 import java.io.BufferedInputStream;
 import java.io.IOException;
 import java.io.InputStream;
@@ -17,6 +21,7 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
+import javax.annotation.Nullable;
 import org.apache.beam.sdk.io.FileSystems;
 import org.apache.beam.sdk.io.fs.MatchResult.Metadata;
 import org.apache.beam.sdk.options.ValueProvider;
@@ -70,8 +75,10 @@ public abstract class SchemaStore<T> implements Serializable {
 
   ////////
 
-  protected SchemaStore(ValueProvider<String> schemasLocation) {
+  protected SchemaStore(ValueProvider<String> schemasLocation,
+      ValueProvider<String> schemaAliasesLocation) {
     this.schemasLocation = schemasLocation;
+    this.schemaAliasesLocation = schemaAliasesLocation;
   }
 
   /* Returns true on a valid schema name e.g. `document.schema.json` */
@@ -85,6 +92,8 @@ public abstract class SchemaStore<T> implements Serializable {
   private static final org.slf4j.Logger LOG = org.slf4j.LoggerFactory.getLogger(SchemaStore.class);
 
   private final ValueProvider<String> schemasLocation;
+  @Nullable
+  private final ValueProvider<String> schemaAliasesLocation;
   private transient Map<String, T> schemas;
   private transient Set<String> dirs;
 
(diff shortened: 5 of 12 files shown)


ingestion-beam/src/test/resources/schemaAliasing/example-aliasing-config.json
```diff
@@ -0,0 +1,12 @@
+{
+  "aliases": [
+    {
+      "alias": "some-product/baseline",
+      "base": "glean/baseline"
+    },
```
akkomar: @fbertsch how do you like this way of configuring aliases?

mreid-moz: I thought we would want to alias entire namespaces, so we could have something like 
```
{
  "alias": "some-product",
  "base": "glean"
}
```

without having to specify each individual `docType`.  This would mean `some-product/*` would map to `glean/*`.

However, re-reading #424 seems to match the approach here.

An advantage to aliasing an entire namespace would be that when Glean adds a new doc type, all applications would be able to use it without adding more aliases. A disadvantage would be that we still need to manage the single glean schema being used by multiple doctypes in the mozilla-pipeline-schemas repo and build step.

fbertsch: I prefer this approach. The probe-scraper will keep track of default pings and new pings added per-product, so we won't have to take care of that in mozilla-pipeline-schemas, which seems appropriate since we will also keep track of new repositories in the probe-scraper. That way all of the information is contained to one place.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/JSONSchemaStore.java
```diff
@@ -18,11 +19,21 @@
...
   }
 
-  protected JSONSchemaStore(ValueProvider<String> schemasLocation) {
-    super(schemasLocation);
+  /**
+   * Returns a SchemaStore based on the contents of the archive at schemasLocation
+   * with additional schemas aliased according to configuration.
+   */
+  public static JSONSchemaStore of(ValueProvider<String> schemasLocation,
+      @Nullable ValueProvider<String> aliasingConfigurationLocation) {
```
jklukas: I have a slight preference for simplifying this by removing the single argument constructor and making this a required field rather than nullable, since the real use in the Decoder will always be passing in both value providers, even if the produced value ends up being empty.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
```diff
@@ -0,0 +1,37 @@
...
+
+package com.mozilla.telemetry.schemas;
+
+import static org.apache.commons.lang3.builder.ToStringStyle.SHORT_PREFIX_STYLE;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+class SchemaAlias {
```
jklukas: Slight preference for making this an AutoValue class to reduce some boilerplate and make this more consistent with AutoValue usage elsewhere in this codebase and in Beam.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAliasingConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+package com.mozilla.telemetry.schemas;
+
+import static org.apache.commons.lang3.builder.ToStringStyle.SHORT_PREFIX_STYLE;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.List;
+import org.apache.commons.lang3.builder.ToStringBuilder;
+
+class SchemaAliasingConfiguration {
```
jklukas: I'd also be interested in this being AutoValue. Unless hooking up an AutoValue class for Jackson is a pain.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
```diff
@@ -97,13 +105,15 @@ int numLoadedSchemas() {
...
+    Multimap<String, String> schemaAliasesMap;
     InputStream inputStream;
     try {
+      schemaAliasesMap = loadSchemaAliasesMap();
       Metadata metadata = FileSystems.matchSingleFileSpec(schemasLocation.get());
       ReadableByteChannel channel = FileSystems.open(metadata.resourceId());
       inputStream = Channels.newInputStream(channel);
     } catch (IOException e) {
-      throw new IOException("Exception thrown while fetching from configured schemasLocation", e);
+      throw new IOException("Exception thrown while loading schemas configuration", e);
```
jklukas: I'd prefer to keep schemaAliasesLocation as separated as possible from schemasLocation rather than nesting the loading of aliases inside loadAllSchemas.

Can we provide a loadSchemaAliases() method that gets called first and then either pass in the value to this method or have that method set a private variable to store the aliases?

That will separate concerns a bit more and allow us to keep these exception messages as specific as possible.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParsePayloadTest.java
```diff
@@ -36,7 +36,7 @@ public void testOutput() {
         .apply("AddAttributes", MapElements.into(TypeDescriptor.of(PubsubMessage.class))
             .via(element -> new PubsubMessage(element.getPayload(), ImmutableMap.of(
                 "document_namespace", "test", "document_type", "test", "document_version", "1"))))
-        .apply(ParsePayload.of(schemasLocation));
+        .apply(ParsePayload.of(schemasLocation, null));
```
jklukas: instead of nulls, I'd prefer we pass in a value provider via `pipeline.newProvider` with a null value, or if that doesn't work a `StaticValueProvider.of(null)`.

akkomar:
> Are the glean schemas distinctive enough that we can simply inspect the contents of a JSON payload and determine if it's a glean ping, and further infer which type of glean ping it is?

Interesting idea. Yes - I think that what `$schema` keyword is for, and we [should](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/validation/glean/baseline.1.all.pass.json#L2) be sending it.
`$schema` seems to be equivalent to @fbertsch's suggestion [here](https://github.com/mozilla/gcp-ingestion/issues/424#issue-408319725):
> We could additionally consider adding the aliased namespace and docType as metadata to the ping.


This has the benefit of making adding aliasing easier for clients other than Glean-based (in the current approach we'd need to generalize "aliasing pipeline" for them) at the expense of complicating decoder a little.
I don't see why this wouldn't work for Glean. I also don't see any huge drawbacks from the whole pipeline point of view (apart from complicating schema validation contract, but that's what aliasing does anyway). /cc @fbertsch @mreid-moz

jklukas:
I totally missed that @fbertsch already suggested adding schema to the ping. I wish I had realized the implications earlier in the discussion.

Having a required field in the general Glean schema to identify the ping type sounds like it would give us a robust basis for routing schemas appropriately, or maybe `$schema` already gives us what we need? I don't fully understand the implications of the discussion in https://github.com/mozilla/gcp-ingestion/pull/508#discussion_r270333335 above.

Being able to remove a dependency on the pipeline for scraping and publishing schema aliases would be a good thing for stability of the Decoder, in my opinion.

fbertsch:
Deciding which schemas to use at validation-time seems totally reasonable to me. In fact the Glean schemas already have that information: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/dev/schemas/glean/baseline/baseline.1.schema.json#L7

We could generalize that: `if ping contains "$schema" key then use that schema instead of namespace/doctype`. [In fact, that's what the ref docs indicate the field should be used for.](https://json-schema.org/latest/json-schema-core.html#rfc.section.7). They don't need to be valid URLs, and we could retrieve them from mps if it is prefixed with `moz://`.

mreid-moz:
My reservation is that this gets us back into having to parse the _contents_ of a document before we can determine which schema to apply. As things stand now we can determine it completely from the submission URL, and with the aliasing support, we can do it from a combination of the submission URL and the aliases list.

This also effectively changes the meaning of the `docType` in the submission URL to "target table", and leaves it to the contents of the ping itself to tell us its `docType`. This is not necessarily a bad thing, but it is a pretty fundamental change. Would we do the same for `docVersion`?

I agree that it would be nice to simplify the logic and tracking of this information, but I worry about pushing that logic inside the ping itself.

jklukas:
> This also effectively changes the meaning of the docType in the submission URL to "target table", and leaves it to the contents of the ping itself to tell us its docType

> I agree that it would be nice to simplify the logic and tracking of this information, but I worry about pushing that logic inside the ping itself.

To address those two concerns, maybe we could introduce a new concept of "scope" to namespace. What if we had the SomeProduct browser specify namespace as something like `glean.some-product` rather than as `some-product`:

    https://incoming.t.m.o/submit/glean.some-product/baseline/1/{uuid}
    https://incoming.t.m.o/submit/glean+some-product/baseline/1/{uuid}

If there's no `.` in the namespace, then we assume the "standard" scope and apply the given namespace to determine both the validation schema and the destination table. If a `.` appears then the validation scope namespace is `glean` and the destination table namespace is `some-product`.

relud:
if it's not too late to use `glean+` or `glean.` or even `https://incoming.t.m.o/glean/some-product/baseline/1/{uuid}`, i think that's an excellent solution

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
akkomar: Nice, I didn't know about AutoValue.
I thought about Lombok when generating this, but didn't want to introduce this library here.

mreid-moz:
How does everyone feel about taking this PR + implementation as a functional first step, then following up with alternate design / architecture proposal(s) separately?

It seems to address the immediate problem of handling reference browser and fenix schemas without a ton of manual schema-related work, and if we update the alias config "by hand" in the short term, the automation around keeping the mapping file up to date could wait until we've had a chance to have a round of design discussions.

akkomar:
> How does everyone feel about taking this PR + implementation as a functional first step, then following up with alternate design / architecture proposal(s) separately?
> 
I am for it - we need a solution sooner than later. Btw., regarding automation - Frank mentioned that there will be some Glean config generation anyway related to table schemas, so having "Glean ping pipeline" shouldn't be something we're tyring to avoind here.

Out of the ideas mentioned here I like Jeff's idea:
> What if we had the SomeProduct browser specify namespace as something like glean.some-product or glean+some-product rather than as some-product

the most. But since it involves changing public api's contract (so also requires minor changes in Glean client library) I'd argue to not introduce it hastily.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
akkomar: I went one step further and decoupled loading aliases from loading schemas. This way it's not mixed with archive-reading code and would be easy to change/remove in the future.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -31,8 +31,9 @@
  */
 public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
 
-  public static ParsePayload of(ValueProvider<String> schemasLocation) {
-    return new ParsePayload(schemasLocation);
+  public static ParsePayload of(ValueProvider<String> schemasLocation,
+      ValueProvider<String> aliasingConfigurationLocation) {
+    return new ParsePayload(schemasLocation, aliasingConfigurationLocation);
```
jklukas: I'd prefer the variable names follow the option names all the way through, so I'd like to see `aliasingConfigurationLocation` change to `schemaAliasesLocation` throughout this PR.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
```diff
@@ -0,0 +1,23 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.auto.value.AutoValue;
+
+@AutoValue
+abstract class SchemaAlias {
+
+  @JsonCreator
+  static SchemaAlias create(@JsonProperty(value = "base", required = true) String base,
+      @JsonProperty(value = "alias", required = true) String alias) {
```
jklukas: The verbosity needed for the jackson annotations is annoying and kind of ruins the brevity benefit of AutoValue classes, but I still thank you for making this change as AutoValue offers nice protection from some of the subtle bugs that can creep in when managing fields by hand.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/StorageIntegrationTest.java
```diff
@@ -125,12 +125,16 @@ public void testDecoderWithRemoteInputAndOutput() throws Exception {
   public void testCompileDataflowTemplate() throws Exception {
     String gcsPath = "gs://" + bucket;
 
+    String aliases = System.getProperty("user.dir")
```
jklukas: We don't build absolute paths with `"user.dir"` elsewhere in the tests; is there any reason we can't simply pass the relative path?

If we want to introduce more robust path handling for tests, I'd prefer that be a separate PR.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaAlias.java
akkomar: Still, that's a little bit less boilerplate.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/StorageIntegrationTest.java
akkomar: Thanks and sorry, that's probably leftover from some experiments.

whd:
> How does everyone feel about taking this PR + implementation as a functional first step, then following up with alternate design / architecture proposal(s) separately?

+1. I haven't been following this discussion but probably should have been, so a follow up architecture proposal sounds swell.

ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
```diff
@@ -131,10 +140,66 @@ private void loadAllSchemas() throws IOException {
...
+          String docTypeWithVersionAlias = docTypeAlias + versionAndSuffix;
+
+          tempSchemas.put(namespaceWithDocTypeAlias + "/" + docTypeWithVersionAlias, schema);
+        }
+      }
+    });
+
+    schemas = tempSchemas;
+    dirs = tempDirs;
+  }
```
jklukas: This is a pretty elegant approach to merging in aliased schemas to the existing schema map leaving the existing interface for looking up schemas unchanged. The one drawback I see to this is that it doesn't give ParsePayload any way of knowing that it was a glean schema being used for validation.

We had discussed previously adding "validation_namespace" and "validation_schema" attributes to the PubsubMessage in the case that an alias was followed to determine a schema to validate against. Those attributes could then be used by the Debug View pipeline to filter out any non-Glean messages. Does that still seem useful? If so, we may need to change the interface here so that we can return the name of the underlying schema returned.

akkomar: > We had discussed previously adding "validation_namespace" and "validation_schema" attributes to the PubsubMessage in the case that an alias was followed to determine a schema to validate against.

I'm sorry, I'm not sure I follow. If we go with `incoming.t.m.o/glean+product/glean+doc/1/{uuid}`, these attributes can be added to metadata earlier in the processing chain (say, `ParseUri`?). Then `SchemaStore` wouldn't need to be aware of aliasing and we could just revert its changes I'm introducing here.
Does this make sense?

jklukas: Yes, that makes sense. For the time being, then, the republisher topic for debug pings won't have any metadata that distinguishes glean pings from others and you'll have to look inside the payload to determine if something follows the glean schema. If that's okay for your purposes, then it's fine by me.

akkomar: > republisher topic for debug pings won't have any metadata that distinguishes glean pings from others

I think that's OK - I can look at `document_namespace`+`document_type` and validate against a list of products that I'm interested in (I could even reuse aliasing configuration introduced in this PR for that).

PR #511, opened 2019-04-02
Fix #504 - Add public methods for retrieving attribute map from DynamicPathTemplate

### LINKED ISSUE #504: Include attributes for fully qualified names to schemas into DynamicPathTemplate
A schema in mozilla-pipeline-schemas is fully qualified using the namespace, document type, and document version. The DynamicPathTemplate should always have access to the namespace, type, and version, since they form the fully qualified name to a schema. 

`FileIO.writeDynamic` takes a PCollection, groups by a destination type, and writes each partition to a sink using the Destination as context. The current File output type uses the `DynamicPathTemplate` to extract attributes from an incoming PubsubMessage into a list for partitioning and to replace placeholders in the naming template.

`AvroIO.sinkViaGenericRecords` is a sink that can be used with `FileIO.writeDynamic`. It requires a schema derived from the destination type. It is difficult to query the schema store unless the list contains the attributes for the fully qualified schema name in fixed positions. In the draft implementation of the Avro ingestion pipeline, I replaced the DynamicPathTemplate with a TreeMap using a custom coder.

https://github.com/mozilla/gcp-ingestion/blob/01b5955b205a54e2b7bb888059eea040def6758d/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java#L237-L240

However, this causes the File and Avro output types to diverge in implementation. We can abstract this detail behind the DynamicPathTemplate.

acmiyaguchi:
This fixes #504. We may want to include attributes in the placeholder values extracted by DynamicPathTemplate in order to fetch the schema for a document. This uses fixed locations at the head of the list to store the document namespace, type, and version.


ingestion-beam/src/test/java/com/mozilla/telemetry/util/DynamicPathTemplateTest.java
```diff
@@ -88,6 +92,20 @@ public void testExtractValuesFrom() {
...
+        "document_namespace", "x", "document_type", "y", "document_version", "z");
+    final ImmutableMap<String, String> partialQualifiedNameAttributes = ImmutableMap
+        .of("document_type", "z");
+
+    assertThat(path.extractValuesFrom(attributes), is(ImmutableList.of("x", "y", "z", "hi")));
+    assertThat(path.extractValuesFrom(null), is(ImmutableList.of("$", "$", "$", "b")));
+    assertThat(path.extractValuesFrom(partialQualifiedNameAttributes),
+        is(ImmutableList.of("$", "z", "$", "b")));
+  }
+
```
jklukas: Can you add a test case that more closely matches the reason the new methods are necessary for the Avro sink? If we wrote to a directory path that includes namespace, doctype, and version and thus placeholders like `${document_namespace:-foo}` appeared in the template, would these new methods be unnecessary? Could we document that paths must include those three attributes and raise an exception if that is violated?

It feels strange to have this use case-specific logic baked into this class. Maybe this would feel more natural as a subclass of DynamicPathTemplate rather than using a boolean constructor param to toggle on this functionality.

acmiyaguchi: Can do. The core bit is being able to reconstruct a Map<PlaceholderName, PlaceholderValue> from the List<PlaceholderValue>, which is only possible if the placeholder names are in scope and the names contain the three required attributes. I'll rework it to see if I can make it flow more naturally.

acmiyaguchi:
This is ready for another look. This adds two public methods to DynamicPathTemplate.

- `getPlaceholderNames` - Return a list of placeholders, which can be used to assert whether expected values exist shortly after construction
- `getPlaceholderAttributes` - Generate a map given a list of values, which will be used to pass into the SchemaStore.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/DynamicPathTemplate.java
```diff
@@ -58,25 +58,37 @@ public DynamicPathTemplate(String rawPath) {
 
   }
 
+  /** Return the placeholder names in the order of appearance in the original path. */
+  public List<String> getPlaceholderNames() {
+    return placeholderNames;
+  }
+
+  /** Return a mapping of placeholders to their original values. */
```
jklukas:
```suggestion
  /** Return a mapping of placeholder names to the passed values. */
```

PR #515, opened 2019-04-03
 Fix #512 - Insert default value of null for missing, nullable fields 

### LINKED ISSUE #512: GuidedJsonDecoder does not insert default values for nullable fields
When running the JsonDecoder, I've seen many messages in the following form:

```
{
  "attributeMap": {
    "error_message": "org.apache.avro.AvroTypeException: Expected field name not found: campaign",
    "error_type": "com.mozilla.telemetry.io.Write$AvroOutput$3@734fccac",
    "document_namespace": "telemetry",
    "exception_class": "org.apache.avro.AvroTypeException",
    "stack_trace": "[com.mozilla.telemetry.avro.GuidedJsonDecoder.doAction(GuidedJsonDecoder.java:204), org.apache.avro.io.parsing.Parser.advance(Parser.java:88), com.mozilla.telemetry.avro.GuidedJsonDecoder.readIndex(GuidedJsonDecoder.java:464), org.apache.avro.io.ResolvingDecoder.doAction(ResolvingDecoder.java:290), org.apache.avro.io.parsing.Parser.advance(Parser.java:88), org.apache.avro.io.ResolvingDecoder.readIndex(ResolvingDecoder.java:267), org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:179), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:153), org.apache.avro.generic.GenericDatumReader.readField(GenericDatumReader.java:232), org.apache.avro.generic.GenericDatumReader.readRecord(GenericDatumReader.java:222), org.apache.avro.generic.GenericDatumReader.readWithoutConversion(GenericDatumReader.java:175), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:153), org.apache.avro.generic.GenericDatumReader.read(GenericDatumReader.java:145...",
    "document_version": "10",
    "document_type": "core"
  },
  "payload": "..."
}
```

This particular field is union-ed with null, so the decoder should insert the appropriate value.

acmiyaguchi:
This fixes #512.

Fields may be specified in schemas, but missing in the actual documents. If the schema specifies the field as nullable i.e. `"type": [null, type]` in the definition, then the avro json-decoder should insert the default value into the document.

I've added several test cases to verify the behavior of the decoder around this particular issue:
* a missing, required field
* a null value for a field of type boolean with a boolean default
* a missing, nullable field of type [null, boolean] with a boolean default
* a missing, nullable field of type [null, boolean] with a null default

I think the values should be as follows:
* type exception
* boolean default
* boolean default
* null

However, the actual results after the patch are:
* type exception
* type exception
* null
* null

This patch only deals with the last two cases, ensuring that the first one stays constant. This decoder ignores the default value in the schema because the default information is not provided via the symbols in the underlying generating-grammar. However, this should still be consistent with BigQuery's behavior.


acmiyaguchi:
I filed [AVRO-2367](https://issues.apache.org/jira/projects/AVRO/issues/AVRO-2367) for extending the FieldAdjustAction symbol to include the default value, which would allow us to use defaults other than null.

acmiyaguchi:
This is not ready for review yet -- I've merged this into my development branch and found subtle bugs preventing ingestion of sampled documents.

acmiyaguchi:
I found the bug and tested this with the core ping. I need to write the appropriate test case to demonstrate the behavior though.

acmiyaguchi:
This is ready for another look. I added two failing test cases and another patch to fix it.

* An object containing an unused field with two nullable fields in the schema
* An empty object with two nullable fields in the schema

The first case was failing because the json parser would reach the end of the current object, while still needing to look at other nullable fields. The second would occur because the resolving decoder would start a field-adjust action, but the default value was not inserted.

These were both fixed by moving the `if (parser.topSymbol() == Symbol.UNION)` statement into the top level scope of the field-adjust action piece.

Issue #396, opened 2019-01-28
Need pub/sub topics with a random sample of data


trink:
To reduce pub/sub costs the sample streams should be broken out into separate topics.  For anomaly detection we only require a partition on normalized channel and then use the following sample sizes for each:
1% of release
10% of beta
100% of nightly

jklukas:
I assume this is talking about output from Decoder? Do we want to include all docTypes where `namespace=telemetry`?

mreid-moz:
cc @acmiyaguchi as he's been working on sampled datasets in AWS land.

jklukas:
@trink - This is being implemented in https://github.com/mozilla/gcp-ingestion/pull/517

Issue #513, opened 2019-04-02
java.lang.NumberFormatException with ParseProxy in Decoder


whd:
[Job](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-west1/jobs/2019-04-02_10_11_57-2698747101293808431?project=moz-fx-data-beam-nonprod-b5f4&folder&organizationId)

Sample stack trace:

     java.lang.NumberFormatException: For input string: ""
            java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
            java.lang.Long.parseLong(Long.java:601)
            java.lang.Long.parseLong(Long.java:631)
            java.text.DigitList.getLong(DigitList.java:195)
            java.text.DecimalFormat.parse(DecimalFormat.java:2051)
            java.text.SimpleDateFormat.subParse(SimpleDateFormat.java:2162)
            java.text.SimpleDateFormat.parse(SimpleDateFormat.java:1514)
            java.text.DateFormat.parse(DateFormat.java:364)
            com.mozilla.telemetry.decoder.ParseProxy$Fn.apply(ParseProxy.java:58)
            com.mozilla.telemetry.decoder.ParseProxy$Fn.apply(ParseProxy.java:40)
            org.apache.beam.sdk.transforms.Contextful.lambda$fn$36334a93$1(Contextful.java:112)
            org.apache.beam.sdk.transforms.MapElements$1.processElement(MapElements.java:135)

I checked some records in landfill e.g. gs://moz-fx-data-stage-landfill/telemetry/2019-04-02/20/output-2019-04-02T20:20:00.000Z-2019-04-02T20:30:00.000Z-0-00176-of-00200.ndjson.gz and `x_pipeline_proxy` looks correct.

whd:
More generally [this page](https://console.cloud.google.com/errors?time=P1D&order=COUNT_DESC&resolution=OPEN&resolution=ACKNOWLEDGED&project=moz-fx-data-beam-nonprod-b5f4&organizationId=442341870013) appears to be a useful resource for inspecting errors. For instance, there is another one "ArrayIndexOutOfBoundsException: 11946" in ParseProxy. The other ones are probably innocuous.

Issue #482, opened 2019-02-28
Add sample_id to BigQuery tables


jklukas:
Breakout from #477 

We will want to have `sample_id` available (`crc32(clientId) % 100`). BQ does not have a crc32 function, so we will need to populate that as part of the pipeline. Further, we may want this to be a top-level field so it can be used for clustering.

We will likely want to make this an `int` rather than `string` as it is in the parquet datasets.

jklukas:
Once this is in place, we can update the Republisher to produce monitoring topics based on sample_id rather than random numbers.

And per @whd:

> This sample id (and whether it matches the AWS pipeline's) is another easy comparison point across pipelines for validation.

relud:
thought: we could make this `crc32(clientId)`, and handle `% 100` downstream so that we can select even smaller datasets. This would be especially useful for using clustering in BigQuery to get very small samples.

jklukas:
> This would be especially useful for using clustering in BigQuery to get very small samples.

Tell me more. Would we cluster on the crc32 value directly?

relud:
yeah, and then we would do something like `SELECT ... WHERE crc32 < MAX_CRC32/100`

relud:
Now that I think about it, doing `SELECT ... WHERE MOD(crc32, 100) = 42` probably wouldn't actually reduce cost, and that's v different from the sample `crc32 < MAX_CRC32/100`, so IDK if this will actually be useful.

jklukas:
I think I'm inclined to target just the existing `sample_id` for simplicity. But allowing smaller samples would be a good thing to keep in mind for a `sample_id_v2` that uses a different hashing algorithm like MD5 with better avalanche behavior.

relud:
that sounds great

jklukas:
See https://github.com/mozilla/gcp-ingestion/pull/526

jklukas:
sample_id is now being produced in the Decoder and I've added a corresponding field to the core_v9 and core_v10 tables. Watching to see new data flow in.

jklukas:
sample_id is being populate in core tables. Closing.

PR #517, opened 2019-04-05
Republisher

### LINKED ISSUE #479: Implement Republisher
See the requirements in https://github.com/mozilla/gcp-ingestion/pull/471

jklukas:
Fixes #479 and #396

Adds the Republisher job as previously spec'd in `docs/architecture`.

Before this is deployed, we will need to create appropriate output topics
for the intended configuration.

Once this is deployed and stable, we can remove the decoded topic consumer
from Decoder. There will be a short overlap period where both the Decoder
and Republisher are marking messages as seen in Redis, but this shouldn't
cause any problems (other than the expense of two consumers).


docs/architecture/README.md
```diff
@@ -92,7 +92,7 @@ This document specifies the architecture for GCP Ingestion as a whole.
 - Must send messages rejected by transforms to a configurable error destination
    - Must allow error destinations in PubSub and Cloud Storage
 
-### Republisher (Not Yet Implemented)
+### Republisher
```
jklukas: For the Republisher on telemetry data, I'd anticipate the following configuration:

```
--input=projects/moz-fx-data-beam-nonprod-b5f4/subscriptions/telemetry-decoded.republisher

--perDocTypeEnabledList=event
--perDocTypeDestination=projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-doctype-${document_type}

--perChannelSampleRatios='{"nightly":1.0,"beta":0.1,"release":0.01}'
--perChannelDestination=projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-sampled-${channel}
```

@sunahsuh requested an event-ping-specific topic and the per-channel topics are as requested in https://github.com/mozilla/gcp-ingestion/issues/396

Which would require creation of a new subscription and of 4 new topics:

```
projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-doctype-event
projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-sampled-nightly
projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-sampled-beta
projects/moz-fx-data-shar-nonprod-efed/topics/telemetry-decoded-sampled-release
```

For the Republisher on structured data, I'd anticipate the following configuration:

```
--input=projects/moz-fx-data-beam-nonprod-b5f4/subscriptions/structured-decoded.republisher
--enableDebugDestination
--debugDestination=projects/moz-fx-data-shar-nonprod-efed/topics/structured-decoded-debug
```

Which would require creation of one new subscription and one new topic. At this point, I don't think it's necessary to break out any docTypes or sample channels for the low volume of structured ingestion data.

cc @whd on the above. Particularly for thoughts on naming.

jklukas:
cc @akkomar who will be taking advantage of the debug topic.

ingestion-beam/README.md
```diff
@@ -448,6 +455,99 @@ echo '{"payload":"dGVzdA==","attributeMap":{"remote_addr":"63.245.208.195"}}' >
...
+
+### Per-`docType` Republishing
+
+If `--perDocTypeEnabledList` is provided, a separate producer will be created
+for each `docType` specified in the given comma-separated list.
+See the `--help` output for details on format.
+
+### Per-Channel Sampled Republishing
+
+If `--perChannelSampleRatios` is provided, a separate producer will be created
```
mreid-moz: What happens to messages with values of `channel` outside the listed set?

ingestion-beam/src/main/java/com/mozilla/telemetry/Republisher.java
```diff
@@ -0,0 +1,143 @@
...
+      RepublisherOptions.Parsed opts = options.as(RepublisherOptions.Parsed.class);
+      opts.setOutput(NestedValueProvider.of(options.getPerChannelDestination(),
+          pattern -> pattern.replace("${channel}", targetChannel)));
+
+      decoded //
+          .apply("RandomlySample" + StringUtils.capitalize(targetChannel), Filter.by(message -> {
+            message = PubsubConstraints.ensureNonNull(message);
+            String channel = message.getAttribute("app_update_channel");
+            return targetChannel.equals(channel)
+                && ThreadLocalRandom.current().nextDouble() < ratio;
```
mreid-moz: I believe this should be randomly sampling by `clientId` rather than uniformly across all pings.

ingestion-beam/README.md
jklukas: If a channel is not explicitly listed, then there is no output topic for that channel and no messages of that channel type are republished via this perChannel mechanism.

ingestion-beam/src/main/java/com/mozilla/telemetry/Republisher.java
jklukas: Breaking this off to https://github.com/mozilla/gcp-ingestion/issues/519

docs/architecture/README.md
whd: A little late, but after having read https://github.com/mozilla/gcp-ingestion/issues/396 and the above it appears we're going to split the sample out by channel. I don't think we need this level of granularity, and the tradeoff of managing and reading per-channel sampled topics vs. reading the full sampled stream and selecting specific channels as needed seems to me in favor of a producing a single sampled topic. That said, having the sample size vary by channel within a single topic also seems like it could be very wonky from an analysis perspective, depending on how the sample stream is used (for the CEP it would fine). /CC @trink 

For now I'm going to implement this with per-channel topics but I would like to discuss whether this is something we really want.

jklukas: I imagine that the 1% sample of release is going to be a nice go-to for testing or for streaming applications that can suffice on a smaller stream.

That topic is more representative of the overall population than a combined stream of the sample topics would be, so I'm inclined to keep it split up, but I could be convinced otherwise.

whd: Ok, if @trink agrees we'll keep it as is. Likely only the CEP will need to read the three streams; the majority of applications will only read the release topic.

Another feature I would like that looks like it may be missing is the ability to generically sample an entire stream, kind of like how the GCS sink is agnostic as to the data it is syncing. With https://github.com/mozilla/gcp-ingestion/issues/482 it looks like we're going to move to sampling by client_id for decoded topics, but I was planning on running a republisher in production on `telemetry-raw` that republishes 1% (random sample as it is now) of the raw topic to stage for the CI dataflow jobs to read and catch issues like https://github.com/mozilla/gcp-ingestion/issues/513.

jklukas: > I was planning on running a republisher in production on telemetry-raw

How about we use "sample_id" if it's present in the message and fall back to a random integer if it's not?

jklukas: I forgot about the channel part. Maybe we just add an additional configuration option for a random sample on the whole stream.

whd: That's more or less what I want (since telemetry-raw wont have sample_id), but I would also need the ability to configure "1% of all messages" regardless of channel.

PR #518, opened 2019-04-05
Remove MarkAsSeen from Decoder

jklukas:
Once the Republisher (see #517) is deployed and stable, it will be marking IDs as seen in Redis, so the Decoder no longer needs to do so.

This PR removes all code related to Decoder publishing ids to Redis.


whd:
This is probably ok to merge now after a rebase, or perhaps wait until Monday to make sure the Republisher continues to work as expected.

jklukas:
> This is probably ok to merge now after a rebase, or perhaps wait until Monday to make sure the Republisher continues to work as expected.

Republisher looks happy after half a day of running, so I think we're good to merge this in.

PR #521, opened 2019-04-08
Generate `avro-schema-test.tar.gz` programmatically

acmiyaguchi:
This generates an avro schema artifact that's generated programmatically. This script will only update the artifact when the contents of the existing file changes, so it can be run without generating a completely new binary every time.

This is used for testing the AvroSchemaStore and the Avro sink.


jklukas:
Can you give me more context on how this fits into the overall strategy for avro schemas? Are we going to be publishing this file via mozilla-pipeline-schemas? If so, I'd rather all the logic for generating the file live there rather than in this repo.

acmiyaguchi:
Maybe the branch-name is a misnomer, but this generates unit-testing resources for `AvroSchemaStore` and `AvroSink`. The schema store generically accepts a `tar.gz` file for testing. I am planning to generate `ndjson` unit-test cases through this same file, since the pubsub message requires a base64 encoded payload.

acmiyaguchi:
https://github.com/mozilla/gcp-ingestion/blob/e40addff72acc2e3475ec4b680dd89666e5c1f8b/ingestion-beam/src/test/java/com/mozilla/telemetry/schemas/AvroSchemaStoreTest.java#L19-L34

This was originally generated by hand and contained empty files. This PR adds valid avro schemas that are easy to update.

ingestion-beam/bin/generate-avro-test-resources.py
```diff
@@ -0,0 +1,103 @@
...
+    logging.info("Done!")
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.DEBUG)
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--output-path",
+        type=str,
+        default=os.path.join(INGESTION_BEAM_ROOT, "src/test/resources/testdata"),
```
jklukas: My (undocumented) assumption was that the testdata directory would contain only json and ndjson files representing payloads; so "test data" in the sense that it's data that could be input or output of a beam pipeline.

I'd prefer we put this in a separate directory. "src/test/resources/tarballs"?  "src/test/resources/avro"?

ingestion-beam/bin/generate-avro-test-resources.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+
```
jklukas: Can you document in an extended comment or a module docstring here more about the context of this script and how to invoke it? In particular, it took me a while to realize that the output of this script is checked into the repo, so this would only need to be rerun if the developer was needing to change the test avro schemas.

ingestion-beam/bin/generate-avro-test-resources.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
```
jklukas: Can you remove the `.py` extension on this file to match the other executables in this directory?

acmiyaguchi:
There's a failing test related to the republisher -- but this PR shouldn't add any complications.

PR #526, opened 2019-04-09
Calculate sample_id in the Decoder

jklukas:
This calculates sample_id, adds to metadata, and pulls it out as a top-level field for BQ tables. It also leverages sample_id for stably sampling topics in Republisher.

The approach to top-level fields may change as the discussion evolves in the [ping metadata proposal](https://docs.google.com/document/d/1AsshL_QS1qRVXcHW5lz0jZeM-UKIWFl1oV3XJz7Xewo/edit#), but for now I'm sticking to the method that exists.


jklukas:
Closes https://github.com/mozilla/gcp-ingestion/issues/482

PR #531, opened 2019-04-15
Update download schemas

acmiyaguchi:
This modifies the `download-schemas` script to pull down the latest schemas and applies the patch for testing resources. I also did some refactoring of the `generate-avro-schemas` script after changing the name of the top-level directory in the archive.

```
$ ./bin/download-schemas 
patching file schemas/test/test/test.1.schema.json
patching file templates/test/test/test.1.schema.json
Running cleanup!

$ ./bin/generate-avro-schemas 
71/71 sucessfully converted
Running cleanup!
```


acmiyaguchi:
Aw, `patch` isn't installed on the docker container.

acmiyaguchi:
```
$ ./bin/download-schemas 
Checking patch schemas/test/test/test.1.schema.json...
Checking patch templates/test/test/test.1.schema.json...
Applied patch schemas/test/test/test.1.schema.json cleanly.
Applied patch templates/test/test/test.1.schema.json cleanly.
Running cleanup!
```

ingestion-beam/bin/download-schemas
```diff
@@ -1,12 +1,49 @@
...
+    }
+    trap cleanup EXIT
+    cd "$workdir" || exit
+}
+
+function insert_testing_schemas() {
+    # Patch the current schema archive with testing resources. This function
+    # should be called within the root directory of the extracted archive.
+    patch_url="https://github.com/jklukas/mozilla-pipeline-schemas/commit/9f8d4252a170ce9241a153b3b49eef235e1279a2.patch"
+    curl -sL ${patch_url} | git apply -v -p1
```
jklukas: This is some deep git that I would not have known to reach for. TIL.

For maintainability's sake, I think we may be better off avoiding patching in this commit and instead copy the contents directly into this script via heredocs.

Something like:

```
cat > schemas/test/test/test.1.schema.json << 'EOF'
{
  "$schema" : "http://json-schema.org/draft-04/schema#",
  "type" : "object",
  "title" : "default_schema",
  "properties" : {
  },
  "required" : []
}
EOF
```

I think that would let us get away from having to reference a commit from a random fork (which I again apologize for introducing as a test dependency here).

acmiyaguchi: The single quotes around EOF is a neat trick, I didn't know you could suppress parameter expansion that way.

Issue #478, opened 2019-02-27
Add "normalized" attributes (normalizedChannel, etc.)


jklukas:
@mreid requested that we get the following fields into BQ in https://github.com/mozilla/gcp-ingestion/issues/477:

- normalizedChannel
- normalizedOSVersion
- normalizedOs
- normalizedAppName

We don't currently produce those. I'll need to know where to look to find the old pipeline's logic for these.

whd:
Most likely https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/modules/moz_telemetry/normalize.lua

PR #532, opened 2019-04-16
Add normalized_channel attribute

jklukas:
Supports https://github.com/mozilla/gcp-ingestion/issues/478


ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
```diff
@@ -0,0 +1,47 @@
...
+    assertEquals("esr", Normalize.channel("esr"));
+    assertEquals("beta", Normalize.channel("beta"));
+    assertEquals("aurora", Normalize.channel("aurora"));
+    assertEquals("nightly", Normalize.channel("nightly"));
+
+    assertEquals("nightly", Normalize.channel("nightly-cck-ubuntu"));
+
+    assertEquals("Other", Normalize.channel("no"));
+    assertEquals("Other", Normalize.channel("esr52"));
+    assertEquals("Other", Normalize.channel("Release"));
```
jklukas: @trink @whd 

The existing lua implementation of channel name normalization doesn't force inputs to lower case. Do we want to preserve this behavior such that "Release" is normalized to "Other"?

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
```diff
@@ -0,0 +1,47 @@
...
+   * }</pre>
+   */
+  public void channel() {
+    assertEquals("release", Normalize.channel("release"));
+    assertEquals("esr", Normalize.channel("esr"));
+    assertEquals("beta", Normalize.channel("beta"));
+    assertEquals("aurora", Normalize.channel("aurora"));
+    assertEquals("nightly", Normalize.channel("nightly"));
+
+    assertEquals("nightly", Normalize.channel("nightly-cck-ubuntu"));
```
jklukas: @trink @whd 

There is a special case in the lua implementation to handle "nightly-cck-" prefix. Is that still needed? Do we have documentation of the significance of that case?

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+    assertEquals("Other", Normalize.channel("no"));
+    assertEquals("Other", Normalize.channel("esr52"));
+    assertEquals("Other", Normalize.channel("Release"));
+    assertEquals("Other", Normalize.channel("Prefeitura Municipal de Santos"));
+    assertEquals("Other", Normalize.channel("no"));
+    assertEquals("Other", Normalize.channel(null));
+
+    assertEquals("beta", Normalize.channel("beta-cdntest"));
+    assertEquals("beta", Normalize.channel("betastop"));
```
jklukas: @trink @whd 

The existing lua implementation accepts anything prefixed with "beta" (no dash required) as "beta" channel. Do we want to preserve that behavior? Do we have documentation for the history of why we have this case?

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
whd: Absent a compelling reason one way or the other I would say keep existing behavior. It would be better to make this decision by data (e.g. if there are a sufficiently large number of pings that we would consider "release" with "Release" as their channel then we might consider making this change). This is the kind of thing that could happen at the "validation work week" next week but it's probably relatively unimportant as I would guess there are few pings with this property.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
whd: I believe [CCK](https://mike.kaply.com/2012/04/13/customizing-firefox-extensions-and-the-cck-wizard/) is for pre-quantum and potentially ESR enterprise deployments of firefox. We might still care about these, though I believe ESR is quantum by now and so these builds wouldn't be supported anymore. @mreid-moz or similar can make a policy decision, but they are probably a sufficiently small number of pings as to be irrelevant how we bucket them (though again, data would be better).

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
whd: Ditto first comment i.e. absent a compelling reason one way or the other I would say keep existing behavior. I'm not aware of any documentation for the history of why we have this case. This is another instance where we could check with data but it may not be worth the effort.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
jklukas:
```
SELECT count(channel)
FROM `moz-fx-data-derived-datasets.telemetry.clients_daily_v6` 
WHERE channel = 'Release'
AND submission_date_s3 = DATE("2019-04-15")
```

1676 users in one day. Not particularly significant. I'll keep the existing behavior.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
jklukas: I'll plan to add a comment explaining what `-cck` means and keep the current behavior.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/NormalizeTest.java
jklukas:
```
SELECT countif(channel = 'beta')/countif(substr(channel, 0, 4) = 'beta') 
FROM `moz-fx-data-derived-datasets.telemetry.clients_daily_v6` 
where submission_date_s3 = DATE("2019-04-15")
```

Yields 0.9999954487218873

We have achieved 5 nines of reliability that a channel string starting with "beta" will be exactly equal to "beta".

Will leave existing behavior.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Normalize.java
```diff
@@ -0,0 +1,42 @@
...
+    if (CHANNELS.contains(name)) {
+      // We could lowercase before doing this check, but don't to maintain historical practice.
+      return name;
+    } else if (name == null) {
+      return OTHER;
+    } else if (name.startsWith("nightly-cck-")) {
+      // The cck suffix was used for various deployments before Firefox Quantum;
+      // cck refers to the "Client Customization Wizard", see
+      // https://mike.kaply.com/2012/04/13/customizing-firefox-extensions-and-the-cck-wizard/
+      return "nightly";
```
trink: return NIGHTLY;

Issue #450, opened 2019-02-21
Preserve non-schema'd telemetry as json


relud:
option 1: whole body is an uncompressed json string in bigquery, table schema matches pubsub message
option 2: fields in body missing from schema are moved to an extras column with uncompressed json string, table schema matches existing schema plus extras column

relud:
option 1 is the planned first-pass on this, because our first use-case is for telemetry doc types with no schema

mreid-moz:
See also #270

jklukas:
Capturing of non-schema'd json to an `additional_properties` field is included in this evolving PR: https://github.com/mozilla/gcp-ingestion/pull/536

jklukas:
The above is merged.

PR #536, opened 2019-04-23
Transform json map fields to arrays before output to BigQuery

jklukas:
In parallel to the work on Avro loading, we want to be able to test stability and performance of a JSON-based solution.

In this PR, we support a variety of transformations to account for limitations of BigQuery schemas. In particular, we support transforming maps into arrays of key/value pairs, serializing structures to json strings where a string is expected, and sanitizing field names that BQ cannot support.

We additionally build up a JSON string of items for which we have no JSON fields in the destination BigQuery table (#450).

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -11,19 +11,31 @@ import com.google.api.services.bigquery.model.TimePartitioning;
 import com.google.cloud.bigquery.BigQuery;
 import com.google.cloud.bigquery.BigQueryOptions;
 import com.google.cloud.bigquery.Dataset;
+import com.google.cloud.bigquery.Field;
+import com.google.cloud.bigquery.Field.Mode;
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.cloud.bigquery.Schema;
+import com.google.cloud.bigquery.Table;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
+import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.mozilla.telemetry.util.Json;
 import java.io.IOException;
 import java.time.Duration;
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.stream.Collectors;
 import org.apache.beam.sdk.coders.KvCoder;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryHelpers;
 import org.apache.beam.sdk.io.gcp.bigquery.TableDestination;
@@ -39,9 +51,8 @@ import org.apache.commons.text.StringSubstitutor;
  * Parses JSON payloads using Google's JSON API model library, emitting a BigQuery-specific
  * TableRow.
  *
- * <p>We also perform some light manipulation of the parsed JSON to match details of our table
- * schemas in BigQuery. In particular, we pull out submission_timestamp to a top level field so we
- * can use it as our partitioning field.
+ * <p>We also perform some manipulation of the parsed JSON to match details of our table
+ * schemas in BigQuery.
  */
 public class PubsubMessageToTableRow
     extends MapElementsWithErrors<PubsubMessage, KV<TableDestination, TableRow>> {
@@ -56,8 +67,9 @@ public class PubsubMessageToTableRow
 
   private final ValueProvider<String> tableSpecTemplate;
 
-  // We'll instantiate this on first use.
+  // We'll instantiate these on first use.
   private transient Cache<DatasetReference, Set<String>> tableListingCache;
+  private transient Cache<TableReference, Schema> tableSchemaCache;
 
   private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate) {
     this.tableSpecTemplate = tableSpecTemplate;
@@ -98,7 +110,9 @@ public class PubsubMessageToTableRow
             .getService();
         Dataset dataset = service.getDataset(ref.getDatasetId());
         if (dataset != null) {
-          dataset.list().iterateAll().forEach(t -> tableSet.add(t.getTableId().getTable()));
+          dataset.list().iterateAll().forEach(t -> {
+            tableSet.add(t.getTableId().getTable());
+          });
         }
         return tableSet;
       });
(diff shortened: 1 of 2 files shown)


acmiyaguchi:
I wrote some scripts to start benchmarking the unmapping transformation using the same methodology for the avro encoding.

* [`bin/generate-bq-schemas`](https://github.com/acmiyaguchi/gcp-ingestion/blob/bq-load/ingestion-beam/bin/generate-bq-schemas) - This generates a flat directory containing bigquery schemas in the form `{namespace}.{doctype}.{docversion}.bigquery.json`. The script will inject the `submission_timestamp` field and can be modified to inject metadata too. 
* [`bin/update-bq-table`](https://github.com/acmiyaguchi/gcp-ingestion/blob/bq-load/ingestion-beam/bin/update-bq-table) - This will create the dataset and tables using the schemas from `generate-bq-schemas` in the current project. Setting `SKIP_EXISTING=1` will skip over existing tables, while `REPLACE_EXISTING=1` will delete and create tables regardless of existing state. This should only be used in a sandbox project where you don't care about the state of the tables. See [this project](https://console.cloud.google.com/bigquery?project=amiyaguchi-unmap-json-bq-sink) for an example of the output.

I created a new project because my existing project already contains a few bigquery tables from the avro-ingestion benchmarking tests. 

* created a new sandbox project called `amiyaguchi-unmap-json-bq-sink` in the `dataops/sandbox` project. 
* created a new bucket with the project name
* copied `gs://avro-ingestion-test/data/raw/telemetry-decoded-2019-04-15-00/` to the bucket root (unfortunately couldn't figure out how to give the service account access to another sandbox project)
* created a `tmp` folder in the bucket root
* enabled the dataflow API

Then I ran the project with the following parameters:

```bash
export GOOGLE_APPLICATION_CREDENTIALS=bq-sink-key.json
BUCKET="gs://$PROJECT"
path="$BUCKET/telemetry-decoded-2019-04-15-00/*-00000-*"
    --runner=Dataflow \
    --autoscalingAlgorithm=NONE \
    --numWorkers=3 \
    --inputFileFormat=json \
    --inputType=file \
    --input=$path\
    --outputType=bigquery \
    --output=$PROJECT:\${document_namespace}.\${document_type}_v\${document_version} \
    --bqWriteMethod=file_loads \
    --errorOutputType=file \
"
```

[Here is the latest run](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-04-23_16_31_55-1961955981487757457?project=amiyaguchi-unmap-json-bq-sink) of the dataflow job on the telemetry-decoded sample.

The errors in the job contain the following errors:

* `Error while reading data, error message: JSON parsing error in row starting at position 0: Nested arrays not allowed.`
* `Error while reading data, error message: JSON parsing error in row starting at position 0: JSON object specified for non-record field: environment.settings.userPrefs`

The job only processed `focus-event` and `mobile-event`, so it'd probably be good to iterate in a tight loop to catch all of the edge cases. I also made a summary out of the error files (after manually concatenating them locally on my machine).

```
$ cat error_-290308-12-21T20_00_00.000Z--290308-12-21T20_10_00.000Z-0000*.ndjson | jq -c '.attributeMap | [.document_namespace, .document_type, .document_version, .error_message]'  | sort | uniq -c

 599 ["telemetry","focus-event","1","java.lang.NullPointerException: null value in entry: value=null"]
 167 ["telemetry","mobile-event","1","java.lang.NullPointerException: null value in entry: value=null"]
```

acmiyaguchi:
I think both of these cases are handled by casting the object into a string. Looking at the PoC table generated from the AWS sampled data [here](https://console.cloud.google.com/bigquery?organizationId=442341870013&project=test-avro-ingest&p=test-avro-ingest&d=telemetry&t=mobile_event_v1&page=table), the `mobile_event_v1` events column is a repeated field of strings, but an array of arrays in JSON. The second case is the generated schema containing `STRING` as the default type but receiving an object.

jklukas:
> an array of arrays in JSON

In parquet-to-bigquery, I believe nested arrays like this are handled by transforming into an array of records where the records have a single `list` field. I'm not sure if that conversion is done by BQ parquet loading or by our own code, but I think we should stick to existing convention.

acmiyaguchi:
That seems like a reasonable way to handle nested arrays. The `array<string>` is actually the result of translating a jsonschema `list<tuple>`. Since the default case for unsupported sub-schemas is a string, it ends up as it is now. A tuple should actually be handled like an anonymous struct (or a struct with names `_0`, `_1`, `_2`, etc), so defaulting to a record with a `list` field is a forward compatible way of handling this. On the other-hand, it might be easier to query the sub-list using `JSON_EXTRACT` if the schema is not specified. There was one other BQ related bug that I caught with the transpiler, so I'll file an issue there.

jklukas:
> A tuple should actually be handled like an anonymous struct (or a struct with names _0, _1, _2, etc), so defaulting to a record with a list field is a forward compatible way of handling this

I hadn't previously been aware of jsonschema "array" functioning as either a list or a tuple, but I've wrapped my head around it now and I agree that we either need to turn this into a struct with generated field names or stick it into a json string.

jklukas:
I'm working this morning on making a more general refactor that can support the transformation of nested arrays.

jklukas:
I just pushed a refactor that allows us to handle multiple types of transformations, using the BQ schema as a guide.

This now supports cleaning of field names (translating special characters to underscores, etc.), jsonifying objects when the BQ schema specifies a repeated string field, and accumulating additionalProperties that aren't present in the BQ schema. I have some tests in here now and am working on more.

The overall flow here is that jsonschema-transpiler has to make decisions about how to transform types that will be incompatible. The code here will need to handle walking the BQ schema in order to detect where payload transformations are expected, and then applying those payload transformations.

acmiyaguchi:
There are new exceptions on another subset of the data.

* optout_v4 - Error while reading data, error message: JSON parsing error in row starting at position 0: JSON object specified for non-record field: payload
* crash_v4 - Error while reading data, error message: JSON parsing error in row starting at position 0: JSON object specified for non-record field: environment.settings.userPrefs
* modules_v4 - Error while reading data, error message: JSON parsing error in row starting at position 0: JSON object specified for non-record field: environment.settings.userPrefs

[[source]](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-04-24_10_56_27-5660116657306023506?project=amiyaguchi-unmap-json-bq-sink)

This is caused by the general error-handling case of the transpiler -- all underspecified or incompatible subschemas are left as strings. There are two ways of handling this which are equally as valid, imo. The first is to stringify everything that the BQ expects as a string which comes with an overhead cost of serialization, but an equally useful ability to use `JSON_EXTRACT` in the correct place in the document. The second is to drop the field in the schema and rely on a mechanism to collect additional properties into a meta field. Both of these cases can handle an important edge-case of the empty schema, which can be used to register schemas into the system without specifying any additional structure.

jklukas:
> The first is to stringify everything that the BQ expects as a string

I've pushed a change to support this coercion.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -158,4 +195,111 @@ private static void promoteFields(TableRow tableRow) {
     }
   }
 
+  /**
+   * Recursively descend into the fields of the passed map, comparing to the passed BQ schema,
```
acmiyaguchi:
```suggestion
   * Recursively descend into the fields of the passed map and compare to the passed BQ schema,
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -158,4 +195,111 @@ private static void promoteFields(TableRow tableRow) {
...
+   * modifying the structure to avoid map types, nested arrays, etc.
```
acmiyaguchi:
```suggestion
   * while modifying the structure to accommodate map types, nested arrays, etc.
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -158,4 +195,111 @@ private static void promoteFields(TableRow tableRow) {
...
+   *
+   * @param parent the map object to inspect and transform
+   * @param bqFields the list of BQ fields expected inside this object
```
acmiyaguchi:
```suggestion
   * @param bqFields the list of expected BQ fields inside this object
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -158,4 +195,111 @@ private static void promoteFields(TableRow tableRow) {
...
+   * @param additionalProperties a map into which we place any fields absent from the BQ schema
```
acmiyaguchi:
```suggestion
   * @param additionalProperties a map for storing fields absent in the BQ schema
```

jklukas:
> it may be worth running a sanity check against bin/download-sampled-landfill to get a wider coverage of documents

Ran the script, uploaded to GCS, and [executed a batch job with the current code from this PR](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-04-25_12_35_17-12610237671121828251?project=klukas-dataflow-dev-2&authuser=0&organizationId=442341870013). No additional errors.

I'm going to go ahead and merge, watch this roll out, and then add the map-type `searches` field to the core ping tables to see if they get correctly populated.

PR #539, opened 2019-04-25
[PoC] Add Cloud Function Avro loader

akkomar:
This is a proof of concept configuration of a Cloud Function for loading Avro files from GCS to BigQuery.

Function itself is defined in  [gcs-bigquery-loader/app/main.py](gcs-bigquery-loader/app/main.py), configuration lives in [Terraform file](gcs-bigquery-loader/main.tf).

Function gets triggered when object in configured GCS bucket is finalized, files are loaded by asynchronous Load Jobs.

Notifications flow through PubSub, but in this setup this is a hidden implementation detail.
For production deployment it would be preferable to configure this as a two step process with GCS notifications published to custom PubSub topic and a function triggered on the topic. That would allow to temporarily disable/redeploy the function without losing GCS notifications.

For monitoring this process we would have to rely on Stackdriver logging and logs-based metrics (functions and load jobs are well integrated with Stackdriver).
For example, one of the most important checks would be to verify if a set of files was successfully loaded.
This can be implemented by querying Stackdriver load jobs' logs and comparing them to a list of relevant files. [README](README.md) contains some notes on how to query Stackdriver logs.
Additionally, custom [metrics](gcs-bigquery-loader/load_jobs_error_codes_metric.dat) can be leveraged to monitor load jobs and function invocations.

Stackdriver logs can also be automatically exported to BigQuery, which would offer more familiar interface for validating state of the system.

Limitations / risks:
* There is a [limit](https://cloud.google.com/bigquery/quotas#load_jobs) of 100k load jobs per day per project, which allows this system to handle 347 tables (assuming each one has a file incoming every 5 minutes). However this shouldn't be a problem, even with Glean pings incoming - right now we have 71 schemas in MPS; we also might be able to raise this limit.


gcs-bigquery-loader/app/main.py
```diff
@@ -0,0 +1,61 @@
...
+        context (google.cloud.functions.Context): Metadata of triggering event.
+    Returns:
+        None; asynchronous BigQuery load job is started, 
+        the output is written to Stackdriver Logging
+    """
+
+    dataset = os.environ.get("BIGQUERY_DATASET", "default_dataset")
+
+    bucket = data["bucket"]
+    file_path = data["name"]
```
jklukas: This looks like we're getting one invocation of the function per file uploaded to GCS. Is that correct?

I'd expect Beam's FileIO to output tens or hundreds of files (whatever numShards is set to) per table per 5 minutes, so I'd expect a load-per-file strategy to quickly run into the per-table daily load job limit.

akkomar: That's correct, it's called every time a file is finalized. If the job is going to dump hundreds of files per batch, that's not a good solution indeed.

akkomar:
As discussed - closing this, w can use this approach in the future for different workloads.

PR #541, opened 2019-04-26
Add scripts for testing bigquery table updates with schemas generated from mps

acmiyaguchi:
This adds two scripts that were used for reducing testing overhead in #536. 

`generate-bq-schemas` will generate BigQuery schemas from mps and insert the partitioning column and metadata. It also includes a fix for extracting the columns from the root record to conform to the `bq` cli tool.

`update-bq-table` will update the current project's BigQuery tables using schemas in the `bq-schemas` folder. This will create the datasets from the namespaces, the tables from the types and version, and the partitioning columns.

```bash
# install jsonschema-transpiler, jq, gcloud cli tools

# generates `schema.tar.gz`
$ ./bin/download-schemas

# generates `bq-schemas/`
$ ./bin/generate-bq-schemas

# updates BigQuery in current project `gcloud config get-value project` 
$ ./bin/update-bq-table
```

I ran the following command to update the tables with data:

```bash
export GOOGLE_APPLICATION_CREDENTIALS=bq-sink-key.json
BUCKET="gs://$PROJECT"
path="$BUCKET/telemetry-decoded-2019-04-15-00/*-00000-*"
    --runner=Dataflow \
    --autoscalingAlgorithm=NONE \
    --numWorkers=3 \
    --inputFileFormat=json \
    --inputType=file \
    --input=$path\
    --outputType=bigquery \
    --output=$PROJECT:\${document_namespace}.\${document_type}_v\${document_version} \
    --bqWriteMethod=file_loads \
    --errorOutputType=file \
"
```

Finally, I wrote a small query using the populated `telemetry.events_v4` table.

```sql
SELECT
  submission_timestamp,
  payload.events.parent,
  metadata
FROM
  telemetry.event_v4
```
[source](https://console.cloud.google.com/bigquery?project=amiyaguchi-unmap-json-bq-sink&j=bq:US:bquxjob_5ab6cad9_16a5af05c7e&page=queryresults)

### DIFF
--- a/ingestion-beam/bin/generate-bq-schemas
+++ b/ingestion-beam/bin/generate-bq-schemas
@@ -1,0 +1,96 @@
+#!/bin/bash
+
+set -e
+
+cd "$(dirname "$0")/.." || exit
+
+if [ ! -x  "$(command -v jsonschema-transpiler)" ]; then
+    echo "jsonschema-transpiler is not installed"
+    echo "Run 'cargo install --git https://github.com/acmiyaguchi/jsonschema-transpiler.git --branch dev'"
+    exit 1
+fi
+
+if [[ ! -f schemas.tar.gz ]]; then
+    echo "Run 'bin/download-schemas'"
+    exit 1
+fi
+
+function init_working_directory() {
+    # Create a temporary directory for working. The current directory is stored
+    # into $rootdir and the working directory is stored into $workdir. The
+    # working directory will be removed on exit.
+    rootdir=$(pwd)
+    workdir=$(mktemp -d -t tmp.XXXXXXXXXX)
+    function cleanup {
+        rm -rf "$workdir"
+        echo "Running cleanup!"
+    }
+    trap cleanup EXIT
+    cd "$workdir" || exit
+}
+
+total=0
+failed=0
+function generate_avro_schema() {
+    # From a relative path to a jsonschema, transpile an bigquery schema in a
+    # flat directory. This function requires the following global variables:
+    # $total, $failed
+    local src=$1
+    local root=$2
+    
+    local namespace
+    local name
+    # replace // with / and get the proper namespace
+    namespace=$(echo $src | sed 's/\/\//\//g' | cut -d/ -f3)
+    name="$namespace.$(basename "$src" .schema.json).bigquery.json"
+    local dst="$root/$name"
+    
+    # Create the folder to the new schema
+    mkdir -p "$root/$relpath"
+
+    if ! jsonschema-transpiler --type bigquery "$src" > "$dst" 2> /dev/null; then
+        echo "Unable to convert $(basename "$src")"
+        rm "$dst"
+        ((failed++))
+    fi
+    tmp="$dst.tmp"
+    
+    # remove top-level record and insert metadata and timestamp
+    submission='{"name": "submission_timestamp", "type": "TIMESTAMP", "mode": "REQUIRED"}'
+    metadata=$(cat << EOM
+{
+    "name": "metadata", 
+    "type": "RECORD", 
+    "fields": [
+        {"name": "key", "type": "STRING", "mode": "REQUIRED"},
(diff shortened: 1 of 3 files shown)


ingestion-beam/bin/generate-bq-schemas
```diff
@@ -0,0 +1,108 @@
+#!/bin/bash
+
+# Generate BigQuery schemas from a schema archive fetched by
+# `bin/download=schemas`. This script will write to a flat directory called
+# `bq-schemas` containing BQ schemas that have been injected with partitioning
```
jklukas: Can we add `bq-schemas` to `.gitignore`?

ingestion-beam/bin/generate-bq-schemas
```diff
@@ -0,0 +1,108 @@
...
+        echo "Running cleanup!"
+    }
+    trap cleanup EXIT
+    cd "$workdir" || exit
+}
+
+total=0
+failed=0
+function generate_bq_schema() {
+    # From a relative path to a jsonschema, transpile an bigquery schema in a
```
jklukas:
```suggestion
    # From a relative path to a jsonschema, transpile a bigquery schema in a
```

ingestion-beam/bin/generate-bq-schemas
```diff
@@ -0,0 +1,108 @@
...
+    metadata=$(cat << EOM
+{
+    "name": "metadata", 
+    "type": "RECORD",
+    "fields": [
+        {"name": "key", "type": "STRING", "mode": "REQUIRED"},
+        {"name": "value", "type": "STRING", "mode": "REQUIRED"}
+    ],
+    "mode": "REPEATED"
+}
```
jklukas: Instead of adding metadata here, we should merge in the appropriate metadata json schema as incubating in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/312 ; I expect to have that PR complete and merged on Monday.

jklukas: On that note, to what extent is this script redundant with the script that @fbertsch is working on for merging metadata schemas in with others. I don't want to fragment the logic with multiple implementations. Would it be possible to document in this repo how to pull down and use the other script for creating and updating bq schemas?

acmiyaguchi: `update-bq-table` should be treated similarly to `download-schemas`. Neither script should be used during operation because the resources are managed elsewhere. I see this script as a way to set up a minimal staging environment for beam-ingestion for development and validation. I left the metadata as a key-value pair to avoid having to implement the per-namespace logic. 

The tables here should be sandboxed and (generally) short-lived.

ingestion-beam/bin/generate-bq-schemas
```diff
@@ -0,0 +1,108 @@
...
+
+src="mozilla-pipeline-schemas"
+dst="bq-schemas"
+
+# Find all JSON schemas in the extracted archive. The top-level folder in the
+# archive is consistently named in `bin/download-schemas`
+tar -xf "$rootdir/schemas.tar.gz" -C "$workdir"
+schemas=$(find $src/schemas -type file -name "*.schema.json")
+
+# Transpile avro schema into a destination directory
```
acmiyaguchi:
```suggestion
# Transpile BigQuery schema into a destination directory
```

ingestion-beam/bin/generate-bq-schemas
jklukas: That makes some more sense. I'm okay with this as long as there's some more documentation at the top of the script to make its intended usage more clear.

jklukas: Also, do you think we should document this workflow in the ingestion-beam README?

acmiyaguchi: I'll write something short under `docs/` and point to it from the README.

acmiyaguchi:
I'm going to make a separate PR for documentation. I ended up making a diagram so it's a bit longer than I originally intended.

PR #542, opened 2019-04-26
Implement nested metadata

jklukas:
Per the accepted [Ping metadata in GCP proposal](https://docs.google.com/document/d/1AsshL_QS1qRVXcHW5lz0jZeM-UKIWFl1oV3XJz7Xewo/edit#).

A json schema for this metadata is being drafted in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/312


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -164,37 +163,6 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate) {
...
-    TableRow tableRow = Json.readTableRow(payload);
-    promoteFields(tableRow);
-    return tableRow;
-  }
-
-  /**
-   * BigQuery cannot partition or cluster tables by a nested field, so we promote some important
-   * attributes out of metadata to top-level fields.
-   */
-  private static void promoteFields(TableRow tableRow) {
```
jklukas: promoteFields is no longer needed, because AddMetadata is now responsible for adding top-level fields like submission_date.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -23,25 +20,12 @@
...
-  static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
-
   static final Field MAP_FIELD = Field //
       .newBuilder("mapfield", LegacySQLTypeName.RECORD, //
           Field.of("key", LegacySQLTypeName.STRING), //
           Field.of("value", LegacySQLTypeName.INTEGER)) //
       .setMode(Mode.REPEATED).build();
 
-  @Test
-  public void testBuildTableRow() throws Exception {
```
jklukas: This test goes away because the `buildTableRow` method is no longer needed.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -4,45 +4,173 @@
...
+        .filter(Map.class::isInstance) //
+        .map(m -> ((Map<String, Object>) m).entrySet()) //
+        .orElse(ImmutableSet.of()) //
+        .forEach(entry -> attributes.put("geo_" + entry.getKey(), entry.getValue().toString()));
+  }
+
+  static Map<String, Object> userAgentFromAttributes(Map<String, String> attributes) {
+    HashMap<String, Object> userAgent = new HashMap<>();
+    attributes.keySet().stream() //
+        .filter(k -> k.startsWith("user_agent_")) //
```
relud: `"user_agent"` and `"geo"` are referenced here enough times that it seems appropriate to pull them into constants so that the compiler can add some safety around spelling

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -4,45 +4,173 @@
...
     // Create an output stream for joining metadata with payload
     final ByteArrayOutputStream payloadWithMetadata = new ByteArrayOutputStream(
-        METADATA_PREFIX.length + metadata.length + payload.length);
-    // Write metadata prefix
-    payloadWithMetadata.write(METADATA_PREFIX);
-    // Write metadata
-    payloadWithMetadata.write(metadata);
+        metadata.length + payload.length);
+    // Write metadata without trailing `}`
+    payloadWithMetadata.write(metadata, 0, metadata.length - 1);
```
relud: probably out of scope, maybe completely unnecessary:

someone mentioned conflicting schemas overriding metadata values with payload values because of repeated keys (e.g. `{"metadata":4}` -> `{"metadata":{...},"metdata":4}`), and I don't know why it didn't occur to me before, but we could give attributes the highest priority by appending them instead of prepending them (e.g. `{"metadata":4}` -> `{"metdata":4,"metadata":{...}}`)

```diff
-    // Write metadata without trailing `}`
-    payloadWithMetadata.write(metadata, 0, metadata.length - 1);
+    // Write payload without trailing `}`
+    payloadWithMetadata.write(payload, 0, payload.length - 1);
    // Start next json field, unless object was empty
    if (payload.length > 2) {
      // Write comma to start the next field
      payloadWithMetadata.write(',');
    }
-    // Write payload without leading `{`
-    payloadWithMetadata.write(payload, 1, payload.length - 1);
+    // Write metadata without leading `{`
+    payloadWithMetadata.write(metadata, 1, metadata.length - 1);
```

jklukas: Either way, we end up with invalid json due to the duplicated keys, yes?

Given the assumption that AddMetadata will only ever happen in a pipeline after ParsePayload has been applied, we should never run into this situation. ParsePayload calls `stripPayloadMetadataToAttributes` which will overwrite any existing attributes with reserved names and remove all fields from the payload that are named like metadata fields. Nothing else alters the payload before hitting AddMetadata, so we should be guaranteed that none of the fields we're adding back in are already present in the payload.

But, having said this, I realize that I did not actually add the call to `stripPayloadMetadataToAttributes` into ParsePayload, so that's a major bug I will fix forthwith.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -4,45 +4,173 @@
...
+        .ifPresent(value -> payload.put(name, value)));
+    return payload;
+  }
+
+  static void stripPayloadMetadataToAttributes(Map<String, String> attributes,
+      Map<String, Object> payload) {
+    Optional.ofNullable(payload) //
+        .map(p -> p.remove("metadata")).filter(Map.class::isInstance) //
+        .ifPresent(m -> {
+          Map<String, Object> metadata = (Map<String, Object>) m;
```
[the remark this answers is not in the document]
jklukas: There's no reason we couldn't add a non-map field directly to the `metadata` object in the future. Likewise, there's no reason we couldn't add a non-string field into any of these maps.

So, it feels like it's happenstance that things aligned such that `metadata` consists of only `Map<String, String>`. I'm tempted to keep it general here to support future evolution.

PR #544, opened 2019-04-30
Remove duplicate schemaLocation config option

### LINKED ISSUE #538: Make Avro an option of outputFileFormat instead of outputType
As currently implemented, the avro sink requires a new dataflow template, since `outputType` is a compile-time parameter. We have `outputFileFormat`, a runtime parameter, which we should use instead, if possible. That way we can reuse the existing gcs sink template. Semantically I think this makes sense as avro is a file format analogous to json text etc.

If configuring avro exclusively with runtime parameters on top of the existing gcs sink template isn't possible, updating the README to this effect and with examples would be useful.

Perhaps unrelated, I'm seeing both `--schemasLocation` and `--schemaLocation` options in IRC conversations and the code; that should probably be cleaned up.

/CC @acmiyaguchi

jklukas:
Fixes #538


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -35,20 +35,6 @@
...
-
-  void setSchemasLocation(ValueProvider<String> value);
-
-  @Description("Path (local or gs://) to a .json file containing list of json schema aliases."
-      + " Example file: schemaAliasing/example-aliasing-config.json"
-      + " If not specified, no schemas will be aliased.")
-  ValueProvider<String> getSchemaAliasesLocation();
-
-  void setSchemaAliasesLocation(ValueProvider<String> value);
-
```
jklukas: DecoderOptions inherits from SinkOptions, so no longer necessary to declare them here.

PR #557, opened 2019-04-30
Save the uri attribute in nested metadata

jklukas:
After deploying #542, the BigQuery sink jobs were failing. The cause
was that we were writing a "uri" string to the metadata object as extra
identifying information, but that clobbered the "uri" struct of the same name,
leading to a string where BQ expected a record.

We now save the uri string inside the uri struct.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -40,13 +40,13 @@
 
   private static final String URI = "uri";
   private static final List<String> URI_ATTRIBUTES = ImmutableList //
-      .of("app_name", "app_version", "app_update_channel", "app_build_id");
+      .of("uri", "app_name", "app_version", "app_update_channel", "app_build_id");
```
jklukas: Had some discussion with @whd about whether it's safe to include the full uri in the payload. He's fine with this for the time being to get us unstuck, but I will open an issue to discuss the reprocessing paths we need and whether there's sufficient motivation to include uri in the payload.

PR #597, opened 2019-05-09
Update pain_points.md

relud:


docs/pain_points.md
```diff
@@ -37,13 +37,23 @@ Crashes the pipeline when the destination table does not exist.
...
+in batch mode.
+
+[bundles]: https://beam.apache.org/documentation/execution-model/#bundling-and-persistence
+[#380]: https://github.com/mozilla/gcp-ingestion/issues/380
 
 ## `PubsubIO.Write`
 
 Does not support dynamic destinations.
 
+Does not support [`NestedValueProvider`] for destinations, which is needed to
```
jklukas: AFAIK, this limitation only exists on Dataflow in streaming mode, where Google moves the implementation into the shuffler. Probably worth being explicit on that.

whd:
Since we just dealt with this, https://github.com/GoogleCloudPlatform/DataflowJavaSDK/issues/632 might be worth mentioning (compile-time option parsing works correctly which is our workaround).

PR #600, opened 2019-05-09
Skip tests for unmodified products

relud:


.circleci/config.yml
```diff
@@ -23,6 +23,20 @@ jobs:
...
     - checkout
+    - run: &skip_unmodified
+        name: Early return when dir not modified
+        command: |
+          if [ "$CIRCLE_BRANCH" = master ]; then
+              echo "Run tests because branch is master"
+          elif git log --format=%B --no-merges -n 1 | grep -qF '[run-tests]'; then
+              echo "Run tests because [run-tests] in commit message"
+          elif ! git diff --quiet origin/master... -- .; then
+              echo "Run tests because $(git rev-parse --show-prefix) was modified since branching off master"
```
jklukas: Should we add one more check that runs all tests if `.circleci` is modified, since that could have an effect on whether tests will succeed or fail?

PR #612, opened 2019-05-21
Refactor attribute normalization

jklukas:
This cleans up some technical debt, consolidating all field normalization
to a single transform rather than scattering normalizations at the points
where various fields are initially extracted.

We also refactor the way we look for attributes within the payload in order
to consolidate all "common ping"-style logic together.

We also now support parsing some attributes out of Glean pings as well.
This means that Glean tables will now have sample_id populated, which
will be helpful as we start to take advantage of clustering.


jklukas:
cc @fbertsch so you're aware of the new functionality here for Glean. It means Glean pings should now have top-level normalized_os, normalized_channel, and sample_id fields populated by the pipeline.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -158,44 +148,50 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+  private void addAttributesFromPayload(Map<String, String> attributes, JSONObject json) {
+
+    // Try to get glean-style client_info object.
+    Optional<JSONObject> gleanClientInfo = Optional.of(json) //
+        .map(j -> j.optJSONObject("client_info"));
+
+    // Try to get "common ping"-style os object.
+    Optional<JSONObject> commonPingOs = Optional.of(json) //
+        .map(j -> j.optJSONObject("environment")) //
+        .map(j -> j.getJSONObject("system")) //
```
relud: to prevent future headaches if a valid schema exists that doesn't match this shape I think we should make each nested key optional
```suggestion
        .map(j -> j.optJSONObject("system")) //
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -158,44 +148,50 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+        .map(j -> j.getJSONObject("os"));
```
relud:
```suggestion
        .map(j -> j.optJSONObject("os"));
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -158,44 +148,50 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+      ifNonEmpty(gleanClientInfo.get().optString(CLIENT_ID), v -> attributes.put(CLIENT_ID, v));
+    } else if (commonPingOs.isPresent()) {
+      // See common ping structure in:
+      // https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/common-ping.html
+      ifNonEmpty(commonPingOs.get().optString("name"), v -> attributes.put(OS, v));
+      ifNonEmpty(commonPingOs.get().optString("version"), v -> attributes.put(OS_VERSION, v));
+    } else {
+      // Try to extract "core ping"-style values; see
+      // https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/da4a1446efd948399eb9eade22f6fcbc5557f588/schemas/telemetry/core/core.10.schema.json
+      ifNonEmpty(json.optString("os"), v -> attributes.put(OS, v));
```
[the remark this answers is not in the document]
jklukas: I'm skipping the constant here because we have to use a literal for "osversion" which is different from the constant.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -158,44 +148,50 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+    // Try extracting variants of top-level client id.
+    ifNonEmpty(json.optString(CLIENT_ID), v -> attributes.put(CLIENT_ID, v));
+    ifNonEmpty(json.optString("clientId"), v -> attributes.put(CLIENT_ID, v));
 
-    if (!Strings.isNullOrEmpty(osName)) {
-      attributes.put("normalized_os", Normalize.os(osName));
-    }
+    // Add sample id.
+    Optional.ofNullable(attributes.get(CLIENT_ID)) //
+        .ifPresent(v -> attributes.put(SAMPLE_ID, Long.toString(calculateSampleId(v))));
```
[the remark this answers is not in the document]
jklukas: I'll use the filter solution.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
jklukas: That was my intention. Will change.

PR #616, opened 2019-05-23
Implement uri parsing for DSMO

### LINKED ISSUE #611: Support DSMO in the decoder
This is for stub installer urls, as per [legacy-systems in the edge spec](https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#legacy-systems):
> Accept [Stub Installer pings](https://firefox-source-docs.mozilla.org/browser/installer/windows/installer/StubPing.html) as GET to `/stub/[docVersion]/[dimensions]`, with no `docType` or `docId`, and over both HTTP and HTTPS. Use [POST/PUT Response codes](https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#postput-response-codes), even though this endpoint is for GET requests.

supports [bug 1357257](https://bugzilla.mozilla.org/show_bug.cgi?id=1357257)

this is high priority, it blocks an H1 goal for @javaun

relud:
fixes #611


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -54,13 +72,166 @@ private ParseUri() {
...
+      Matcher matcher = PING_VERSION_PATTERN.matcher(pingVersion);
+      if (!matcher.find()) {
+        throw new UnknownPingVersionException(pingVersion);
+      }
+      // Check length of elements is valid for this version
+      int unexpectedElements = elements.length - SUFFIX_LENGTH.get(matcher.group(VERSION));
+      if (unexpectedElements != 0) {
+        throw new UnexpectedPathElementsException(unexpectedElements);
+      }
+      // Initialize new payload with ping version and funnelcake
```
jklukas: Is there documentation we can link here for what funnelcake is?

relud: hmm, seems like https://wiki.mozilla.org/Funnelcake should do
```suggestion
      // Initialize new payload with ping version and funnelcake ID
      // https://wiki.mozilla.org/Funnelcake
```

relud:
> I'd love to inject more context here so that somebody looking at this code for the first time can easily understand what a "stub" is, what "DSMO" is/was, what a funnelcake is 🍰, and hopefully see a reference of the old logic that this is reimplementing.

Added a doc string to the stub uri class, and moved it to the end of ParseUri so that it interrupts the rest of the logic less.

> Is there any plan to move off of this logic and have stub installer send a normal structured ingestion payload, or will this need to be maintained for the long-term?

302 -> @matt-howell

matt-howell:
> Is there any plan to move off of this logic and have stub installer send a normal structured ingestion payload, or will this need to be maintained for the long-term?

There's no current plan. I don't have any objection to making that change, the problem is that old stub installers tend to stay in use basically indefinitely, so we wouldn't really be able to fully switch over for a long time without dropping some number of pings.

PR #619, opened 2019-05-24
Factor Republisher and support per-namespace topics

jklukas:
[Pocket has a need to ingest all doctypes associated with the `activity-stream` namespace](https://bugzilla.mozilla.org/show_bug.cgi?id=1553999); it seems efficient to be able to deliver one topic rather than multiple topics per docType.

We also take this chance to refactor the per-channel and per-doctype logic
out of Republisher proper into their own dedicated transforms.

As part of the refactor, we make the execution graph cleaner and a bit more
efficient. We now partition per channel or doctype first, rather than having
each configured channel and doctype branch directly off the initial input.
We also give the output transforms names specific to the channel or doctype.


jklukas:
cc @whd

whd:
Looking at the config in the bug I would slightly prefer a topic structure of `--perDocTypeEnabledList=activity-stream/*` map to the notion of namespace e.g.
`--perDocTypeDestination=projects/moz-fx-data-shar-nonprod-efed/topics/structured-decoded-namespace_${document_namespace}`. But using `/*` to create a psuedo-doctype of `all` seems reasonable enough. This also means we reserve the `all` doctype or rather that we can't differentiate between an `all` doctype topic and a namespace topic.

jklukas:
Given @whd's feedback, I think I'm going to add a separate transform and set of options to handle per-namespace topics rather than mashing it together with per-docType.

Issue #468, opened 2019-02-26
Review docs in this repo, update to match current state


jklukas:
Per request of @mreid-moz

whd:
One specific instance, https://github.com/mozilla/gcp-ingestion/blob/master/docs/decoder.md#decoded-message-metadata-schema which I was going to point at in https://bugzilla.mozilla.org/show_bug.cgi?id=1553999 is probably missing some newer fields like `normalized_country_code`.

PR #636, opened 2019-05-30
Document ParseProxy transform

jklukas:
A follow-up to #468 suggested in https://github.com/mozilla/gcp-ingestion/issues/496#issuecomment-497164953

### DIFF
--- a/docs/architecture/README.md
+++ b/docs/architecture/README.md
@@ -79,6 +79,9 @@ This document specifies the architecture for GCP Ingestion as a whole.
    1. Parse `uri` attribute into multiple attributes
    1. Gzip decompress `payload` if gzip compressed
    1. Validate `payload` using a JSON Schema determined by attributes
+   1. Parse `x_pipeline_proxy` attribute; if present and valid as a timestamp,
+      archive the value of `submission_timestamp` to `proxy_timestamp`,
+      replacing with the `x_pipeline_proxy` value
    1. Resolve GeoIP from `remote_addr` or `x_forwarded_for` attribute into
       `geo_*` attributes
    1. Parse `agent` attribute into `user_agent_*` attributes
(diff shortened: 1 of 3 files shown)


docs/architecture/README.md
```diff
@@ -60,6 +60,9 @@ This document specifies the architecture for GCP Ingestion as a whole.
    - Must use a 5XX error error code
 - Must accept configuration mapping `uri` to PubSub Topic
    - Expected initial topics are Structured Ingestion, Telemetry, and Pioneer
+- Must accept configuration defining HTTP headers to capture
+   - Must store configured headers as PubSub message attributes
+   - Must name attributes as the header name in lowercase with `-` converted to `_`
```
jklukas: Does this bit cover what you think needs to be communicated about headers, @whd?

whd: I'd put this in the `METADATA_HEADERS` section of https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-edge/README.md#configuration since it's an implementation detail and not really high-level architecture.  Though this does raise the point that the high-level architecture doc makes no statements on the configuration or ability of the edge to preserve headers, and the [edge specification](https://github.com/mozilla/gcp-ingestion/blob/master/docs/edge.md#edge-server-pubsub-message-schema) also mentions headers only in passing in the schema definition, and that notably `x-pipeline-proxy` is not present there.

With apologies for not doing it myself; as I stated, I think for the purpose for which this discussion arose simply conveying the explicit header name that the client should supply and the expected value of the header from this documentation is sufficient. I am actively trying not to bikeshed too much on documentation which is why I approved this as-is; for my purposes the original PR suffices.

docs/edge.md
```diff
@@ -74,6 +74,7 @@ required group attributes {
   optional string x_forwarded_for      // example: "10.98.132.74, 103.3.237.12"
   optional string x_pingsender_version // example: "1.0"
   optional string x_debug_id           // example: "my_debug_session_1"
+  optional string x_pipeline_proxy     // time that the AWS->GCP tee received the message, example: "2018-03-12T21:02:18.123456Z"
```
jklukas: Added `x_pipeline_proxy`

PR #645, opened 2019-06-11
Beam 2.13.0

jklukas:
Closes #639

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/RepublisherMainTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/RepublisherMainTest.java
@@ -37,10 +37,10 @@ public class RepublisherMainTest extends TestWithDeterministicJson {
     String input = inputPath + "/*.ndjson";
     String output = outputPath + "/out";
 
-    Republisher
-        .main(new String[] { "--inputFileFormat=json", "--inputType=file", "--input=" + input,
-            "--outputFileFormat=json", "--outputType=file", "--enableDebugDestination",
-            "--debugDestination=" + output, "--outputFileCompression=UNCOMPRESSED" });
+    Republisher.main(new String[] { "--inputFileFormat=json", "--inputType=file",
+        "--input=" + input, "--outputFileFormat=json", "--outputType=file",
+        "--enableDebugDestination", "--debugDestination=" + output,
+        "--outputFileCompression=UNCOMPRESSED", "--errorOutputType=stderr" });
 
     List<String> inputLines = Lines.files(inputPath + "/debug-messages.ndjson");
     List<String> outputLines = Lines.files(outputPath + "/out*.ndjson");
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/RepublisherMainTest.java
```diff
@@ -37,10 +37,10 @@ public void testDebugDestination() throws Exception {
...
     String output = outputPath + "/out";
 
-    Republisher
-        .main(new String[] { "--inputFileFormat=json", "--inputType=file", "--input=" + input,
-            "--outputFileFormat=json", "--outputType=file", "--enableDebugDestination",
-            "--debugDestination=" + output, "--outputFileCompression=UNCOMPRESSED" });
+    Republisher.main(new String[] { "--inputFileFormat=json", "--inputType=file",
+        "--input=" + input, "--outputFileFormat=json", "--outputType=file",
+        "--enableDebugDestination", "--debugDestination=" + output,
+        "--outputFileCompression=UNCOMPRESSED", "--errorOutputType=stderr" });
```
jklukas: `errorOutputType` defaults to `pubsub` and previously, these tests ran without actually calling any code for the error output, bit looks like 2.13.0 changes something internally in Beam such that the error output is accessed for display data, leading to exceptions (since this test isn't actually set up to provide everything a pubsub-based error output needs).

We now add `"--errorOutputType=stderr"` explicitly to many tests to avoid this scenario.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/PubsubIntegrationTest.java
```diff
@@ -130,7 +131,7 @@ public void canSendPubsubOutput() throws Exception {
 
     SinkOptions.Parsed sinkOptions = pipeline.getOptions().as(SinkOptions.Parsed.class);
     sinkOptions.setOutput(pipeline.newProvider(topicName.toString()));
-    sinkOptions.setOutputPubsubCompression(pipeline.newProvider(Compression.UNCOMPRESSED));
+    sinkOptions.setOutputPubsubCompression(StaticValueProvider.of(Compression.UNCOMPRESSED));
```
relud: I thought we changed this to use `pipeline.newProvider` because `StaticValueProvider.of` was causing issues? i mean, CI passes, so i'm fine with either, just confused.

jklukas: StaticValueProvider didn't cause problems that I'm aware of, but `pipeline.newProvider` should be more realistic since it will throw an error if you attempt to use a value before the pipeline is running.

There's something about this test that was failing with newProvider with Beam 2.13. I don't deeply understand the issue, but I'm assuming it's a bug in the DirectRunner or in the particular way we're manipulating options in the context of this test. I will add a comment to that effect.

Issue #651, opened 2019-06-12
Hitting API rate limit in BQ sink


jklukas:
We have at least one message going to error output due to:

> "error_message":"com.google.common.util.concurrent.UncheckedExecutionException: com.google.cloud.bigquery.BigQueryException: Exceeded rate limits: too many api requests per user per method for this user_method. For more information, see https://cloud.google.com/bigquery/troubleshooting-errors"

> "stack_trace":"[com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2051), com.google.common.cache.LocalCache.get(LocalCache.java:3953), com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4873), com.mozilla.telemetry.transforms.PubsubMessageToTableRow.processElement(PubsubMessageToTableRow.java:139), com.mozilla.telemetry.transforms.PubsubMessageToTableRow.processElement(PubsubMessageToTableRow.java:57), com.mozilla.telemetry.transforms.MapElementsWithErrors$DoFnWithErrors.processElementOrError(MapElementsWithErrors.java:88), com.mozilla.telemetry.transforms.MapElementsWithErrors$DoFnWithErrors$DoFnInvoker.invokeProcessElement(Unknown Source), org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.invokeProcessElement(SimpleDoFnRunner.java:213), org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.processElement(SimpleDoFnRunner.java:178), org.apache.beam.runners.dataflow.worker.SimpleParDoFn.processEleme..."

See gs://moz-fx-data-stage-data/telemetry-decoded_bq-sink/error/2019-06-08/13/error-2019-06-08T14:00:00.000Z-2019-06-08T14:10:00.000Z-0-00041-of-00060.ndjson.gz

I'm pretty sure this is due to us hitting the endpoint for describing tables, which we use to reference BQ schemas when transforming messages to TableRow objects. We cache the value of the call, but we may not be caching aggressively enough or this may be due to startup where we have to hit the same endpoint for a number of tables.

We perhaps need to add in some logic to detect the exception and hold off.

I will open a support case with Google to confirm this is the limit we're hitting.

jklukas:
I'm not going to open a case with Google, since I'm fairly certain I understand exactly what limit we're hitting now.

[...]

PR #653, opened 2019-06-12
Increase retries for calls to BQ API

jklukas:
Addresses #651

This does not eliminate the possibility of error due to retries, but it at least should significantly reduce the probability.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -4,10 +4,12 @@
 
 package com.mozilla.telemetry.transforms;
 
+import com.google.api.gax.retrying.RetrySettings;
 import com.google.api.services.bigquery.model.DatasetReference;
 import com.google.api.services.bigquery.model.TableReference;
 import com.google.api.services.bigquery.model.TableRow;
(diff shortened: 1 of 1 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -66,11 +68,20 @@ public static PubsubMessageToTableRow of(ValueProvider<String> tableSpecTemplate
   public static final TimePartitioning TIME_PARTITIONING = new TimePartitioning()
       .setField(SUBMISSION_TIMESTAMP);
 
+  // We have hit rate limiting issues that have sent valid data to error output, so we make the
+  // retry settings a bit more generous; see https://github.com/mozilla/gcp-ingestion/issues/651
+  private static final RetrySettings RETRY_SETTINGS = ServiceOptions //
+      .getDefaultRetrySettings().toBuilder() //
+      .setMaxAttempts(12) // Defaults to 6
+      .setTotalTimeout(org.threeten.bp.Duration.ofSeconds(120)) // Defaults to 50 seconds
```
whd: TIL about threeten.

jklukas: TIL as well. I was very confused at first.

PR #654, opened 2019-06-12
Add bqStrictSchemaDocTypes option

### LINKED ISSUE #644: Add option to not fill `additionalProperties` for specific schemas/docTypes
Specifically, we are looking at the main ping. To use table splitting, each table will be "missing" parts of the schema. We want these dropped instead of filled.

Short term, we can have a small main-ping schema, which includes the environment and common ping sections.  We would prefer this be used without additionalProperties as well, to reduce storage cost.

jklukas:
Closes #644

We should be able to set the following for the bq sink on both telemetry
and structured:

    bqStrictSchemaDocTypes=main

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
@@ -388,7 +388,8 @@ public abstract class Write
       streamingInput.ifPresent(messages -> {
         WriteResult writeResult = messages //
             .apply(PubsubMessageToTableRow.of(tableSpecTemplate, strictSchemaDocTypes))
-            .errorsTo(errorCollections).apply(baseWriteTransform //
+            .errorsTo(errorCollections) //
+            .apply(baseWriteTransform //
                 .withMethod(BigQueryWriteMethod.streaming.method)
                 .withFailedInsertRetryPolicy(InsertRetryPolicy.retryTransientErrors()) //
                 .skipInvalidRows() //
@@ -431,7 +432,8 @@ public abstract class Write
         }
         messages //
             .apply(PubsubMessageToTableRow.of(tableSpecTemplate, strictSchemaDocTypes))
-            .errorsTo(errorCollections).apply(fileLoadsWrite);
+            .errorsTo(errorCollections) //
+            .apply(fileLoadsWrite);
       });
 
       PCollection<PubsubMessage> errorCollection = PCollectionList.of(errorCollections)
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
@@ -317,17 +317,21 @@ public abstract class Write
     private final InputType inputType;
     private final int numShards;
     private final ValueProvider<List<String>> streamingDocTypes;
+    private final ValueProvider<List<String>> strictSchemaDocTypes;
 
     /** Public constructor. */
     public BigQueryOutput(ValueProvider<String> tableSpecTemplate, BigQueryWriteMethod writeMethod,
         Duration triggeringFrequency, InputType inputType, int numShards,
-        ValueProvider<List<String>> streamingDocTypes) {
+        ValueProvider<List<String>> streamingDocTypes,
+        ValueProvider<List<String>> strictSchemaDocTypes) {
       this.tableSpecTemplate = tableSpecTemplate;
       this.writeMethod = writeMethod;
       this.triggeringFrequency = triggeringFrequency;
       this.inputType = inputType;
       this.numShards = numShards;
       this.streamingDocTypes = NestedValueProvider.of(streamingDocTypes,
+          value -> Optional.ofNullable(value).orElse(Collections.emptyList()));
+      this.strictSchemaDocTypes = NestedValueProvider.of(strictSchemaDocTypes,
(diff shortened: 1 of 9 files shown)


jklukas:
One detail here is that I'm choosing to still remove fields from the payload if they don't exist in the BQ schema. That should lead to less data to schlep to BigQuery. We could instead choose to leave them in for docTypes we've opted into strict schema handling, and then let BQ ignore them.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -383,8 +387,8 @@ public BigQueryOutput(ValueProvider<String> tableSpecTemplate, BigQueryWriteMeth
 
       streamingInput.ifPresent(messages -> {
         WriteResult writeResult = messages //
-            .apply(PubsubMessageToTableRow.of(tableSpecTemplate)).errorsTo(errorCollections)
-            .apply(baseWriteTransform //
+            .apply(PubsubMessageToTableRow.of(tableSpecTemplate, strictSchemaDocTypes))
+            .errorsTo(errorCollections).apply(baseWriteTransform //
```
fbertsch: nit: should `errorsTo` be indented, and the following `apply` be uninedented?

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -167,12 +172,24 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate) {
     // Strip metadata so that it's not subject to transformation.
     Object metadata = tableRow.remove(AddMetadata.METADATA);
 
+    String namespace = message.getAttribute(ParseUri.DOCUMENT_NAMESPACE);
+    String docType = message.getAttribute(ParseUri.DOCUMENT_TYPE);
+    final boolean strictSchema;
+    if (("telemetry".equals(namespace) && strictSchemaDocTypes.get().contains(docType))
```
fbertsch: This special-casing for telemetry seems unnecessarily, we could remove it (which would simplify the documentation above and the implementation here).

jklukas: Perhaps in retrospect that would have been a good idea. We already support this format for other options that take in a comma separated list of docTypes, though, so I'm inclined to keep consistent.

jklukas:
> Makes sense for preserving existing behavior. We could consider removing as much special-casing of telemetry behavior as we can at some point in the future.

I have become sufficiently convinced that this is a good idea and that this will be easier to change now than later that I added in the change to this PR. All options that accept lists of docTypes now expect entries to be qualified with a namespace.

So the bqSink configuration will need to be:

    bqStrictSchemaDocTypes=telemetry/main

And the Republisher configuration will need to change the following:

    perDocTypeEnabledList=telemetry/event,telemetry/prio

No change is necessary for `bqStreamingDocTypes` since the values there are already fully qualified.

I will not merge this until I can confirm with @whd that we're ready with the updated configuration.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
jklukas: You're right. I've changed this.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -167,12 +172,20 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate) {
     // Strip metadata so that it's not subject to transformation.
     Object metadata = tableRow.remove(AddMetadata.METADATA);
 
+    String namespace = message.getAttribute(ParseUri.DOCUMENT_NAMESPACE);
+    String docType = message.getAttribute(ParseUri.DOCUMENT_TYPE);
+    final boolean strictSchema = (strictSchemaDocTypes.isAccessible()
+        && (("telemetry".equals(namespace) && strictSchemaDocTypes.get().contains(docType))
```
fbertsch: Can't this special-casing be removed now?

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -123,6 +123,19 @@ public void testAdditionalProperties() throws Exception {
     assertEquals(expectedAdditional, Json.asString(additionalProperties));
   }
 
+  @Test
+  public void testStrictSchema() throws Exception {
+    Map<String, Object> parent = new HashMap<>();
+    parent.put("outer", new HashMap<>(ImmutableMap.of("otherfield", 3, //
```
fbertsch: Might be good to add a telemetry test now.

whd:
I've updated the template builds to include fully qualified versions of the config options.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
jklukas: Added an integration test that exercises `bqStrictSchemaDocTypes`

PR #655, opened 2019-06-13
Setup maven for multiple modules

relud:


.circleci/config.yml
```diff
@@ -18,12 +18,16 @@ jobs:
...
+  ingestion-edge: &ingestion-edge
+    working_directory: /root/project/ingestion-edge
     docker:
     - image: python:3.7
     steps:
-    - checkout
-    - run: &skip_unmodified
+    - &checkout
+      checkout:
+        path: /root/project
```
jklukas: This is very nice to clean up the invocations. I did not know setting a checkout path was possible.

PR #658, opened 2019-06-14
Support republishing by random sample

### LINKED ISSUE #650: Support republishing by random sample
See https://github.com/mozilla/gcp-ingestion/pull/517#discussion_r272722475 and subsequent discussion. This would be useful for republishing raw streams to stage for testing purposes, and I had forgotten that this hasn't been (re)implemented yet.

jklukas:
Closes #650

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
@@ -53,11 +53,22 @@ public interface RepublisherOptions extends SinkOptions, PipelineOptions {
 
   void setDebugDestination(ValueProvider<String> value);
 
-  Map<String, Double> getPerChannelSampleRatios();
+  @Description("An output topic name (assuming --outputType=pubsub) for a random sample of"
+      + " incoming messages; messages are chosen based on sample_id if present or by a number"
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
```diff
@@ -53,11 +53,22 @@
...
+
+  @Description("A sampling ratio between 0.0 and 1.0; if not set, no random sample is produced")
+  Double getRandomSampleRatio();
+
+  void setRandomSampleRatio(Double value);
 
   @Description("A JSON-formatted map of channel name to sampling ratio; for example,"
       + " {\"nightly\":1.0,\"release\":0.01} would republish 100% of nightly pings to the"
       + " sampled nightly topic and 1% of release pings to the sampled release topic")
+  Map<String, Double> getPerChannelSampleRatios();
```
jklukas: This change is unrelated, but the description was in the wrong place here, so fixing that up.

whd:
It occurs to me now that I'm not sure what the behavior of the republisher is with respect to deduplication when documents don't have a document id and / or the redis URI is unspecified. Ideally we would simply turn off deduplication by making the redis URI option no longer required (or maybe it is currently optional, but I think it's required). If the behavior is to simply not write anything to redis while processing documents, that's also acceptable behavior, but a little bit confusing.

PR #672, opened 2019-06-24
Patch Beam to allow writing into clustered tables

jklukas:
This is a brittle change that could break if we don't correctly update our
patch when integrating new versions of Beam or if we end up needing to support
writing into both clustered and unclustered tables, so care is needed.

The purpose of this change is to allow us to create our production BigQuery
tables with clustering, which is currently blocked on Beam's lack of support
for configuring clustering. It will be far easier to make that change as
part of setting up production stacks rather than recreating the tables with
clustering after the fact, which will require careful handling to avoid data
loss.

If we integrate this change, we should commit to taking action on helping to
get clustering merged into upstream Beam in its next release.

This supports the currently open [clustering proposal](https://docs.google.com/document/d/1yjg0qj5Cowp41p-vhHqGx7ORnCFdz8QvsluafrFv8mU/edit?ts=5cffafb8#).

### DIFF
--- a/checkstyle/suppressions.xml
+++ b/checkstyle/suppressions.xml
@@ -12,6 +12,9 @@
     <suppress checks="LineLength"
               files="OutputType.java"/>
 
+    <!-- suppress all checks in files we have copied in from other projects -->
+    <suppress checks=".*" files="WriteTables.java" />
+
     <!-- suppress all checks in the generated directories -->
     <suppress checks=".*" files=".+[\\\/]generated[\\\/].+\.java" />
     <suppress checks=".*" files=".+[\\\/]generated-sources[\\\/].+\.java" />
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -21,9 +21,12 @@
         <!-- The classPath setting is a workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925 -->
         <argLine>-Xmx1024m -XX:MaxPermSize=256m -Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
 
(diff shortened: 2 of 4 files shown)


jklukas:
To test this, I created the following table representing a full day of main ping with several histograms parsed out to their own string fields:

```
CREATE OR REPLACE TABLE
  `moz-fx-data-derived-datasets.tmp.klukas_main_histo1`
PARTITION BY
  DATE(submission_timestamp)
CLUSTER BY
  submission_timestamp AS
SELECT
  submission_timestamp,
  JSON_EXTRACT(additional_properties,
    '$.payload.processes.content.keyedHistograms.MEDIA_CODEC_USED') AS keyed_MEDIA_CODEC_USED,
  JSON_EXTRACT(additional_properties,
    '$.payload.processes.content.histograms.HTTP_SUB_DNS_LOOKUP_TIME') AS HTTP_SUB_DNS_LOOKUP_TIME,
  JSON_EXTRACT(additional_properties,
    '$.payload.processes.content.histograms.CONTENT_PAINT_TIME') AS CONTENT_PAINT_TIME
FROM
  `moz-fx-data-shar-nonprod-efed.telemetry.main_v4`
WHERE
  DATE(submission_timestamp) = "2019-06-23"

-- Bytes processed: 5.55 TB
-- Duration: 4 min 34 seconds
-- Output size: 89 GB
```

I'll be posting results of performance tests in this issue.

jklukas:
Posted a Beam PR that hopefully addresses the lingering issues with getting clustering support merged: https://github.com/apache/beam/pull/8945

jklukas:
The upstream Beam PR is now merged (https://github.com/apache/beam/pull/8945) and clustering will be available in the 2.15 release, scheduled for late September.

PR #681, opened 2019-06-27
WIP Reimplement GCS Sink for Kubernetes

relud:
- [X] MVP
- [ ] make CI pass
- [ ] add tests

### DIFF
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
@@ -4,11 +4,18 @@
 
 package com.mozilla.telemetry.ingestion;
 
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
 import com.google.cloud.bigquery.BigQueryOptions;
 import com.mozilla.telemetry.ingestion.io.BigQuery;
+import com.mozilla.telemetry.ingestion.io.Gcs;
 import com.mozilla.telemetry.ingestion.io.Pubsub;
+import com.mozilla.telemetry.ingestion.transform.FileGroupByKey;
+import com.mozilla.telemetry.ingestion.transform.RouteMessage;
 import com.mozilla.telemetry.ingestion.transform.PubsubMessageToTableRow;
 import com.mozilla.telemetry.ingestion.util.Env;
+import com.mozilla.telemetry.ingestion.util.Json;
+import com.mozilla.telemetry.ingestion.util.KV;
 import java.util.concurrent.CompletableFuture;
 
 public class Sink {
@@ -17,19 +24,49 @@ public class Sink {
   }
 
   public static void main(String[] args) {
-    // route messages to OUTPUT_TABLE with OUTPUT_FORMAT
-    PubsubMessageToTableRow routeMessage = new PubsubMessageToTableRow(
-        Env.getString("OUTPUT_TABLE"),
-        PubsubMessageToTableRow.TableRowFormat.valueOf(Env.getString("OUTPUT_FORMAT", "raw")));
+    if (Env.optString("OUTPUT_BUCKET").isPresent()) {
+      // output messages to OUTPUT_BUCKET
+      Gcs.Write gcsWrite = new Gcs.Write(Env.getString("OUTPUT_BUCKET"));
 
-    // output messages to BigQuery
-    BigQuery.Write output = new BigQuery.Write(BigQueryOptions.getDefaultInstance().getService());
+      // output gcs object paths to OUTPUT_TOPIC
+      Pubsub.Write pubsubWrite = new Pubsub.Write(Env.getString("OUTPUT_TOPIC"),
+          Env.getInteger("EXECUTOR_THREADS", 1));
 
-    // read pubsub messages from INPUT_SUBSCRIPTION
-    new Pubsub.Read(Env.getString("INPUT_SUBSCRIPTION"),
-        message -> CompletableFuture.supplyAsync(() -> message) // start new future with message
-            .thenApplyAsync(routeMessage::apply) // determine output path of message
-            .thenComposeAsync(output)) // output message
-                .run(); // run pubsub consumer
+      // write newline delimited strings to files by key
+      FileGroupByKey<String, String> output = new FileGroupByKey<>(
+          Env.getInteger("OUTPUT_BATCH_MAX_BYTES", (int) 1e11), // default 100 GB
+          Env.getDuration("OUTPUT_BATCH_MAX_DELAY", "PT10m"), // default 10 min
+          Env.getInteger("OUTPUT_BATCH_MAX_MESSAGES", (int) 1e9), // default 1bn messages
+          kv -> CompletableFuture.supplyAsync(() -> kv) // start new future with key-file pair
+              .thenApplyAsync(gcsWrite) // upload file and return gcs object path
+              // encode gcs object path in pubsub message
+              .thenApplyAsync(s -> PubsubMessage.newBuilder() //
+                  .setData(ByteString.copyFrom(s.getBytes())).build()) //
+              .thenComposeAsync(pubsubWrite)); // publish to pubsub
+
+      // read pubsub messages from INPUT_SUBSCRIPTION
+      new Pubsub.Read(
+          Env.getString("INPUT_SUBSCRIPTION"),
+          message -> CompletableFuture.supplyAsync(() -> message) // start new future with message
+              .thenApplyAsync(RouteMessage::apply) // determine output path of message
+              .thenApplyAsync(kv -> new KV<>(kv.key, Json.asString(kv.value))) // encode message as json
+              .thenComposeAsync(output)) // output message
+          .run(); // run pubsub consumer
+    } else {
+      // route messages to OUTPUT_TABLE with OUTPUT_FORMAT
+      PubsubMessageToTableRow routeMessage = new PubsubMessageToTableRow(
+          Env.getString("OUTPUT_TABLE"),
+          PubsubMessageToTableRow.TableRowFormat.valueOf(Env.getString("OUTPUT_FORMAT", "raw")));
+
+      // output messages to BigQuery
+      BigQuery.Write output = new BigQuery.Write(BigQueryOptions.getDefaultInstance().getService());
+
+      // read pubsub messages from INPUT_SUBSCRIPTION
+      new Pubsub.Read(Env.getString("INPUT_SUBSCRIPTION"),
+          message -> CompletableFuture.supplyAsync(() -> message) // start new future with message
+              .thenApplyAsync(routeMessage::apply) // determine output path of message
+              .thenComposeAsync(output)) // output message
+          .run(); // run pubsub consumer
+    }
   }
 }
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/Pubsub.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/Pubsub.java
@@ -4,11 +4,20 @@
 
 package com.mozilla.telemetry.ingestion.io;
 
+import com.google.api.core.ApiFuture;
+import com.google.api.core.ApiFutureCallback;
+import com.google.api.core.ApiFutures;
+import com.google.cloud.pubsub.v1.Publisher;
 import com.google.cloud.pubsub.v1.Subscriber;
 import com.google.common.annotations.VisibleForTesting;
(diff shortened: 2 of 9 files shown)


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
```diff
@@ -0,0 +1,52 @@
...
+
+    // read pubsub messages from INPUT_SUBSCRIPTION
+    new Pubsub.Read(
+        Env.getString("INPUT_SUBSCRIPTION"),
+        message -> CompletableFuture.supplyAsync(() -> message) // start new future with message
+        .thenApplyAsync(RouteMessage::apply) // determine output path of message
+        .thenApplyAsync(kv -> new KV<>(kv.key, Json.asString(kv.value))) // encode message as json
+        .thenComposeAsync(output)) // output message
+            .run(); // run pubsub consumer
+  }
```
relud: r? on this part in particular as a way non-dataflow pipelines can be built on top of `CompletableFuture`

pom.xml
```diff
@@ -11,24 +11,62 @@
...
                 <configuration>
-                    <source>1.11</source>
-                    <target>1.11</target>
+                    <source>1.8</source>
+                    <target>1.8</target>
+                    <showWarnings>true</showWarnings>
                     <compilerArgument>-Xlint:all</compilerArgument>
                     <compilerArgument>-Xdoclint:all</compilerArgument>
-                    <compilerArgument>-Werror</compilerArgument>
+                    <!--<compilerArgument>-Werror</compilerArgument>-->
```
jklukas: Should this line still be commented out?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
jklukas: I have no previous experience with CompletableFuture, but this seems quite readable on first pass.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/util/KV.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Container class for a key-value pair
+ */
+public class KV<KeyType, ValueType> {
+
+  public KeyType key;
+  public ValueType value;
+
+  public KV(KeyType key, ValueType value) {
```
jklukas: This class seems well suited to a private constructor and a public factory method so that usage is `KV.of(key, value)`.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/Pubsub.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public Read(String subscriptionName, Function<PubsubMessage, CompletableFuture<?>> output) {
+      ProjectSubscriptionName subscription = ProjectSubscriptionName.parse(subscriptionName);
+      subscriber = Subscriber.newBuilder(subscription,
+          (message, consumer) -> CompletableFuture.supplyAsync(() -> message)
+              .thenComposeAsync(output::apply).whenCompleteAsync((result, exception) -> {
+                if (exception == null) {
+                  consumer.ack();
+                } else {
+                  consumer.nack();
```
jklukas: Should we log exceptions, or record a metric about exceptions?

On that note, do you have any thoughts on metrics for this application, or should we save that for future work? I'm most familiar with dropwizard-metrics as a metrics interface for standalone Java applications, which can emit metrics to Datadog or Stackdriver per configuration.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
```diff
@@ -0,0 +1,52 @@
...
+    FileGroupByKey<String, String> output = new FileGroupByKey<>(
+        Env.getInteger("OUTPUT_BATCH_MAX_BYTES", (int) 1e11), // default 100 GB
+        Env.getDuration("OUTPUT_BATCH_MAX_DELAY", "PT10m"), // default 10 min
+        Env.getInteger("OUTPUT_BATCH_MAX_MESSAGES", (int) 1e9), // default 1bn messages
+        kv -> CompletableFuture.supplyAsync(() -> kv) // start new future with key-file pair
+            .thenApplyAsync(gcsWrite) // upload file and return gcs object path
+            // encode gcs object path in pubsub message
+            .thenApplyAsync(s -> PubsubMessage.newBuilder() //
+                .setData(ByteString.copyFrom(s.getBytes())).build()) //
+            .thenComposeAsync(pubsubWrite)); // publish to pubsub
```
jklukas: Is there a reason we want to handle publishing the path to PubSub ourselves rather than relying on [GCS pubsub notifications](https://cloud.google.com/storage/docs/pubsub-notifications)?

It may be easier to handle all the configuration here rather than hiding the output topic config in the bucket configuration, but GCS may be able to provide better guarantees about delivering messages after a write succeeds.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
```diff
@@ -0,0 +1,101 @@
...
+    private final CompletableFuture<Void> full;
+    private final CompletableFuture<OutputType> result;
+
+    int size = 0;
+    int byteSize = 0;
+
+    private Batch(KeyType key, Function<KV<KeyType, File>, CompletableFuture<OutputType>> handler) {
+      this.key = key;
+      try {
+        file = File.createTempFile("", "");
```
jklukas: Perhaps this is a future improvement rather than something for the initial scope of this PR, but writing the message contents to disk may end up being the performance bottleneck here.

Instead, a `Batch` instance could initiate a multi-part upload to GCS, buffer incoming messages to an in-memory writer, periodically flush the buffer to GCS, and then pass on the mulit-part upload ID, and the next step of processing could finalize the object by making a completion request to GCS with the upload ID.

At least that's an approximation of how it would work on S3. It looks like GCS doesn't have multipart upload, but instead allows you to compose objects together. It's not immediately obvious to me how you would achieve the same workflow on GCS.

pom.xml
relud: no, that's a mistake. -wError makes jshell not work (because of java 11 compiler warnings), so I often have it commented out for testing my assumptions in jshell

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
```diff
@@ -0,0 +1,101 @@
...
+
+    synchronized private KV<KeyType, File> output(Void v, Throwable t) {
+      return new KV<>(key, file);
+    }
+
+    synchronized private Optional<CompletableFuture<OutputType>> add(String message) {
+      if (full.isDone()) {
+        return Optional.empty();
+      }
+      if (size > 0 && (size + 1 > maxMessages || byteSize + message.length() + 1 > maxBytes)) {
```
jklukas: We might need to keep track of these stats globally rather than per batch. We will have dozens of different destinations at once, so it could become hard to reason about how much overall disk/memory space we need if we can only tune max size per batch.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/Pubsub.java
relud: i'm only familiar with statsd and datadog's tagged statsd variant. we should definitely log exceptions, we might initially get away with having stackdriver extract metrics from logs.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
relud: > GCS may be able to provide better guarantees about delivering messages after a write succeeds.

Certainly seems better than failing the whole batch if Pub/Sub is down. I discussed this with @whd in whistler, and decided to write to pubsub here for ease of testing, but I think it's something we should probably make configurable.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
relud: sounds good, I'll attempt that approach if disk batching proves too slow.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
relud: > We might need to keep track of these stats globally rather than per batch

I think we would need both, so that we can complete the largest batch early when shared space passes a threshold and still complete batches early based on individual size.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
relud: I know hindsight queues everything to disk, sometimes more than once, and can generally handle all of telemetry on a single machine if it's not doing significant computation per-message. I was assuming that while an individual batch might fit in memory, all pending batches would not, but the multi-part method you specify should circumvent that.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
relud: Also, currently this depends on flow-control configuration in `Pubsub.Read` to watch global limits, so this shouldn't *break*, but uploading early will definitely have better behavior.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/util/KV.java
relud: Yeah, I was having issues with a static method returning variable type because I haven't done that and couldn't get the syntax right, so I went the simple way here.

Do you think we should follow the beam approach in general of having private constructors and public factory methods everywhere?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
jklukas: It feels a bit odd that basically the entire pipeline is defined as something we pass to Pubsub.Read. Perhaps this can be refactored to have a Pipeline class that's the thing we actually call `run` on? It's not clear how we would manage the startup and shutdown of the pubsub consumer in that case, though.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/FileGroupByKey.java
jklukas: Good point. Definitely sounds like something to delay until we get some experience with the simpler implementation and see if a more complex solution is warranted.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/util/KV.java
jklukas: > Do you think we should follow the beam approach in general of having private constructors and public factory methods everywhere?

Yes, or even better AutoValue all the things, which eliminates some code duplication and enforces that pattern.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/Sink.java
```diff
@@ -0,0 +1,52 @@
...
+
+    // output gcs object paths to OUTPUT_TOPIC
+    Pubsub.Write pubsubWrite = new Pubsub.Write(Env.getString("OUTPUT_TOPIC"),
+        Env.getInteger("EXECUTOR_THREADS", 1));
+
+    // write newline delimited strings to files by key
+    FileGroupByKey<String, String> output = new FileGroupByKey<>(
+        Env.getInteger("OUTPUT_BATCH_MAX_BYTES", (int) 1e11), // default 100 GB
+        Env.getDuration("OUTPUT_BATCH_MAX_DELAY", "PT10m"), // default 10 min
+        Env.getInteger("OUTPUT_BATCH_MAX_MESSAGES", (int) 1e9), // default 1bn messages
```
jklukas: I think I'd prefer that the Env class handle accessing the env var names and providing defaults, so that here we're making calls like `Env.outputBatchMaxBytes()` and perhaps rename the class to `Config`.

relud:
closing this in favor of starting a new PR based on the existing ingestion-sink.

Issue #671, opened 2019-06-21
Convert camel case bigquery column names to snake case


relud:
the direct to parquet datasets coerce camel case keys to snake case, but right now bigquery does not.

for migration convenience we should expose the columns as snake case by some means (e.g. bigquery views or coercion in the decoder)

cc @jklukas

PR #689, opened 2019-07-05
Coerce camelCase field names to snake_case in BQ sink

jklukas:
Fixes https://github.com/mozilla/gcp-ingestion/issues/671

We cannot merge this until corresponding changes are made in the schema transpiler (see https://github.com/mozilla/jsonschema-transpiler/issues/77) and the tables with the new structure are deployed. 

Deploying this change will likely happen at the same time as moving destination tables to `<namespace>_live` datasets in the shared-prod project. See https://bugzilla.mozilla.org/show_bug.cgi?id=1563740


acmiyaguchi:
Unicode segmentation was a red herring, it turns out that [CaseFormat is an ASCII utility](https://github.com/google/guava/blob/8000dc992d0124372972729b20df40a42b27a849/guava/src/com/google/common/base/CaseFormat.java#L24-L25) and finding breaks between unicode characters is mostly unrelated (just the usual separators). Instead, the `heck` create uses its [own definition of a word boundary](https://github.com/withoutboats/heck/blob/093d56fbf001e1506e56dbfa38631d99b1066df1/src/lib.rs#L9-L17) that considers consecutive uppercase characters. This is [purposely not implemented in CaseFormat](https://github.com/google/guava/issues/987) because it's not reversible.

We've already violated the assumption of reversibility by replacing periods and hyphens into underscores. Therefore, I suggest we implement heck's case formatting, because it produces names that seem more intuitive. For example, `clientID` and `clientId` both map to `client_id` instead of `client_i_d` and `client_id`. These also match the casing we see in `main_summary`.

whd:
For the previous implementation see https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/parquet/parquet.cpp#L141-L164, which was aiming at hive compatibility and is more similar to the rust implementation. Generally I agree with making the rust implementation the standard as it produces the more sane result in most cases.

It is unfortunate that there is no well-accepted standard or specification for performing this conversion that multiple libraries can be written against. Absent such a standard, a test suite including all of these special cases across the different language implementations would be ideal.

acmiyaguchi:
The following python snippet is an acceptance test that we could use.

```python
from itertools import product

for c in product("Aa7_", repeat=4):
    word = "".join(c)
    print(word)
```

This generates 256 (4**4) strings that covers most (if not all) of the renaming cases. [This](https://gist.github.com/acmiyaguchi/fbbea6cdb04b964336b8f75185ad68a0) is what it currently looks like.

acmiyaguchi:
I've been trying to find a regular expression to match `heck`'s word boundaries, but I'm not sure it's possible. I think numbers are treated to have the same case as the letter just prior to them (e.g. `a77Aa -> a77_aa` and `A77AAa -> a77a_aa`), which is troublesome for lookbehinds. I've put together 4 different test suites and compared `heck` against a regexp-based script, the direct2parquet logic, and the CaseFormat utility.

https://github.com/acmiyaguchi/test-casing/blob/master/RESULTS.md#rust-vs-python-cases

jklukas:
Perhaps we need to break this into two steps. First, we'd have a function to normalize runs of sequential capital letters such that only the first is capitalized (RAM -> Ram, ProductID -> ProductId), then pass to Guava's CaseFormat.

We could also memoize these conversions if they're getting expensive.

acmiyaguchi:
I found a good solution that matches the rust library by reversing the string to search for all the word boundaries. 

This passes all of the test cases:
* All strings of length 3 drawn with replacement from "aA7"
* All string of length 4 drawn with replacement from "aA7_"
* All column names from the diff against mozilla-pipeline-schemas using `--type bigquery --normalize-case` in [this PR](https://github.com/mozilla/jsonschema-transpiler/pull/79#issuecomment-509839572)
* Unit tests copied from the [`heck::snake` module](https://github.com/withoutboats/heck/blob/master/src/snake.rs#L54-L78)

We should be able to re-implement this in pretty much any language with a built in regex library. Updated test results can be found [here](https://github.com/acmiyaguchi/test-casing/blob/master/RESULTS.md#test-results).

```python
import re

# Search for all camelCase situations in reverse with arbitrary lookaheads.
REV_WORD_BOUND_PAT = re.compile(
    r"""
    \b                                  # standard word boundary
    |(?<=[A-Z])(?=[a-z])                # aA -> a|A boundary
    |(?<=[a-z][A-Z])(?=[A-Z])           # AAa -> A|Aa boundary
    |(?<=[a-z][A-Z])(?=\d*[A-Z])        # A7Aa -> A7|Aa boundary
    |(?<=[a-z][A-Z])(?=\d*[a-z])        # a7Aa -> a7|Aa boundary
    |(?<=[A-Z])(?=\d*[a-z])             # a7A -> a7|A boundary
    """,
    re.VERBOSE,
)


def snake_case(line: str) -> str:
    # replace non-alphanumeric characters with spaces in the reversed line
    subbed = re.sub(r"[^\w]|_", " ", line[::-1])

    # apply the regex on the reversed string
    words = REV_WORD_BOUND_PAT.split(subbed)

    # filter spaces between words and snake_case and reverse again
    return "_".join([w.lower() for w in words if w.strip()])[::-1]
```

acmiyaguchi:
Here are the test cases in csv in `reference,expected` format. A function implementing the above regex (reverse, sub, split, join, reverse) should pass all of the cases.

https://gist.github.com/acmiyaguchi/e377c4b6a53204d7d67e53a20dabc4a1

jklukas:
I've implemented the `SnakeCase` class with a regular expression, and it's passing the test cases. This is also now hooked into the BigQuery sink such that docTypes and namespaces will use the conversion when determining destination table names, as well as being applied to field names.

 Requesting another review from @acmiyaguchi.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -18,24 +18,46 @@
...
+    assertEquals("d2d_enabled", PubsubMessageToTableRow.convertNameForBq("D2DEnabled"));
+    assertEquals("gpu_active", PubsubMessageToTableRow.convertNameForBq("GPUActive"));
+    assertEquals("product_id", PubsubMessageToTableRow.convertNameForBq("ProductID"));
+    assertEquals("l3cache_kb", PubsubMessageToTableRow.convertNameForBq("l3cacheKB"));
+    assertEquals("_64bit", PubsubMessageToTableRow.convertNameForBq("64bit"));
+    assertEquals("hi_fi", PubsubMessageToTableRow.convertNameForBq("hi-fi"));
+    assertEquals("untrusted_modules", PubsubMessageToTableRow.convertNameForBq("untrustedModules"));
+    assertEquals("xul_load_duration_ms",
+        PubsubMessageToTableRow.convertNameForBq("xulLoadDurationMS"));
+    assertEquals("a11y_consumers", PubsubMessageToTableRow.convertNameForBq("A11Y_CONSUMERS"));
```
acmiyaguchi: `speedMHz` is a good one to add, it has an interesting look.

ingestion-beam/src/test/java/com/mozilla/telemetry/util/SnakeCaseTest.java
```diff
@@ -0,0 +1,67 @@
...
+public class SnakeCaseTest {
+
+  @Test
+  public void testSnakeCaseFormat() {
+    // one word
+    assertEquals(SnakeCase.format("Aa"), "aa");
+    // two words
+    assertEquals(SnakeCase.format("aA"), "a_a");
+    // underscores are word boundaries
+    assertEquals(SnakeCase.format("_a__a_"), "a_a");
```
acmiyaguchi: Food for thought:

```java
// underscores are word boundaries
assertEquals(SnakeCase.format("a__a"), "a_a");
// sentences can italicized
assertEquals(SnakeCase.format("_a__a_"), "_a_a_");
// sentences can bolded
assertEquals(SnakeCase.format("__a__a__"), "__a_a__");

PR #710, opened 2019-07-24
Bug 1562011 - Scrub RemoteType annotations from "crash" and "bhr" pings

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1562011)

This rejects crash and bhr pings with payloads starting with `webIsolated=` at the decoder.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -54,7 +54,6 @@ public class MessageScrubber {
       return true;
     } else if ("telemetry".equals(attributes.get(ParseUri.DOCUMENT_NAMESPACE))
         && "bhr".equals(attributes.get(ParseUri.DOCUMENT_TYPE))
-        && "nightly".equals(attributes.get(ParseUri.APP_UPDATE_CHANNEL))
         && (attributes.get(ParseUri.APP_VERSION).startsWith("68")
             || attributes.get(ParseUri.APP_VERSION).startsWith("69"))
         && Optional.of(json) // payload.hangs[].remoteType
@@ -66,6 +65,7 @@ public class MessageScrubber {
                 String s = arr.optJSONObject(i).optString("remoteType");
                 if (s != null && s.startsWith("webIsolated=")) {
                   isPresent = true;
+                  break;
                 }
               }
               return isPresent;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -34,6 +37,42 @@ public static boolean shouldScrub(Map<String, String> attributes, JSONObject jso
...
+            .map(j -> j.optJSONObject("payload")) //
+            .map(j -> j.optJSONObject("metadata")) //
+            .map(j -> j.optString("RemoteType")) //
+            .filter(s -> s.startsWith("webIsolated=")).isPresent()) {
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1562011
+      countScrubbedBug1562011.inc();
+      return true;
+    } else if ("telemetry".equals(attributes.get(ParseUri.DOCUMENT_NAMESPACE))
+        && "bhr".equals(attributes.get(ParseUri.DOCUMENT_TYPE))
+        && "nightly".equals(attributes.get(ParseUri.APP_UPDATE_CHANNEL))
```
whd: This match is less complicated than above because bhr is nightly-only? In which case, do we even care about filtering on nightly?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -34,6 +37,42 @@ public static boolean shouldScrub(Map<String, String> attributes, JSONObject jso
...
+            || attributes.get(ParseUri.APP_VERSION).startsWith("69"))
+        && Optional.of(json) // payload.hangs[].remoteType
+            .map(j -> j.optJSONObject("payload")) //
+            .map(j -> {
+              Boolean isPresent = false;
+              JSONArray arr = j.optJSONArray("hangs");
+              for (int i = 0; i < arr.length(); i++) {
+                String s = arr.optJSONObject(i).optString("remoteType");
+                if (s != null && s.startsWith("webIsolated=")) {
+                  isPresent = true;
```
whd: You could probably short circuit here but it's hardly a performance issue.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
acmiyaguchi: Yep, it's nightly only. Good point.

PR #713, opened 2019-07-26
Preserve original field names in additional_properties

### LINKED ISSUE #711: Field names in `additional_properties` are snake_cased
The field names of the additional properties are normalized by snake_casing when ingesting into BigQuery. While this makes the fields consistent with the rest of the table, this makes it difficult to update the source schemas. I expected the behavior to match the casing of the source data, so we can reduce the size of additional properties over time.

---

To motivate this, I'm looking into updating the main ping schema to include the `payload.info` section.

I wrote a query along the following lines:

```SQL
SELECT
  JSON_EXTRACT(additional_properties,
    "$.payload.info") AS info
FROM
  telemetry_live.main_v4
WHERE
  submission_timestamp >= timestamp '2019-07-23'
  AND submission_timestamp < timestamp '2019-07-24'
```

followed by:

```SQL
CREATE TEMP FUNCTION
  udf_types_js (input STRING)
  RETURNS ARRAY<STRUCT<key STRING,
  value STRING,
  type STRING>>
  LANGUAGE js AS """
    if (input == null) {
      return null;
    }
    result = [];
    let parsed = JSON.parse(input);
    for (let key in parsed) {
      result.push({"key": key, "value": parsed[key], "type": typeof parsed[key]});
    }
    return result;
    """;
WITH
  extracted AS (
  SELECT
    info.*
  FROM
    `amiyaguchi-dev.testing.payload_info`,
    UNNEST(udf_types_js(info)) AS info )
SELECT
  key, type, count(*) as freq
FROM
  extracted
group by 1, 2
order by 1, 2
```

and got results as follows:

Row | key | type | freq |  
-- | -- | -- | -- | --
1 | addons | boolean | 11 |  
2 | addons | string | 275656290 |  
3 | async_plugin_init | boolean | 24239158 |  
4 | flash_version | string | 154062600 |  
5 | previous_build_id | number | 3 |  
6 | previous_build_id | object | 245218834 |  
7 | previous_build_id | string | 31088744 |  
8 | previous_session_id | object | 1970803 |  
9 | previous_session_id | string | 274317059 |  
10 | previous_subsession_id | object | 1279342 |  
11 | previous_subsession_id | string | 275028239 |  
12 | profile_subsession_counter | number | 276307581 |  
13 | reason | string | 276307581 |  
14 | revision | string | 276307581 |  
15 | session_id | string | 276307581 |  
16 | session_length | number | 276254629 |  
17 | session_length | object | 135 |  
18 | session_start_date | string | 276307581 |  
19 | subsesshon_counter | number | 3 |  
20 | subsession_counter | number | 276307578 |  
21 | subsession_id | string | 276307575 |  
22 | subsession_start_date | string | 276307581 |  
23 | timezone_offset | number | 276307580 |  
24 | timezone_offset | object | 1 | 

However, matching this up the the [firefox source docs](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/telemetry/data/main-ping.html), I noticed that the names have been snake cased. In this particular case, it looks like `async_plugin_init` might be useful to add to the schema, but what is the original casing?

This affects other properties that we may want to add to the schemas, such as simple measures or user preferences.

jklukas:
Closes #711

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -22,9 +22,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Maps;
-import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.mozilla.telemetry.decoder.AddMetadata;
 import com.mozilla.telemetry.decoder.ParseUri;
@@ -41,6 +39,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import org.apache.beam.sdk.coders.KvCoder;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryHelpers;
@@ -250,27 +249,33 @@ public class PubsubMessageToTableRow
    */
   public void transformForBqSchema(Map<String, Object> parent, List<Field> bqFields,
       Map<String, Object> additionalProperties) {
+    Map<String, Field> bqFieldMap = bqFields.stream()
+        .collect(Collectors.toMap(Field::getName, Function.identity()));
 
-    // Clean the key names.
-    final Map<String, String> jsonFieldNames = new HashMap<>();
-    ImmutableSet.copyOf(parent.keySet()).forEach(key -> {
-      String transformedKey = getAndCacheBqName(key);
-      jsonFieldNames.put(transformedKey, key);
-      parent.put(transformedKey, parent.remove(key));
-    });
+    HashSet<String> jsonFieldNames = new HashSet<>(parent.keySet());
+    jsonFieldNames.forEach(jsonFieldName -> {
+      final String bqFieldName;
+      if (bqFieldMap.containsKey(jsonFieldName)) {
+        // The JSON field name already matches a BQ field.
+        bqFieldName = jsonFieldName;
+      } else {
+        // Try cleaning the name to match our BQ conventions.
+        bqFieldName = getAndCacheBqName(jsonFieldName);
 
-    // Strip out fields that do not appear in the BQ schema.
-    Set<String> fieldNames = bqFields.stream().map(Field::getName).collect(Collectors.toSet());
-    ImmutableSet.copyOf(Sets.difference(parent.keySet(), fieldNames)).forEach(k -> {
-      Object value = parent.remove(k);
-      if (additionalProperties != null) {
-        additionalProperties.put(jsonFieldNames.get(k), value);
+        // If the field name now matches a BQ field name, we rename the field within the payload,
+        // otherwise we move it to additionalProperties without renaming.
+        Object value = parent.remove(jsonFieldName);
+        if (bqFieldMap.containsKey(bqFieldName)) {
+          parent.put(bqFieldName, value);
+        } else if (additionalProperties != null) {
+          additionalProperties.put(jsonFieldName, value);
+        }
       }
-    });
 
-    // Special transformations for structures disallowed in BigQuery.
-    bqFields.forEach(field -> processField(jsonFieldNames.get(field.getName()), field,
-        parent.get(field.getName()), parent, additionalProperties));
+      Optional.ofNullable(bqFieldMap.get(bqFieldName))
+          .ifPresent(field -> processField(jsonFieldName, field, parent.get(bqFieldName), parent,
+              additionalProperties));
+    });
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jklukas:
Notably, this should not affect current prod data, since all docTypes being ingested to prod already use snake casing consistently (activity-stream and fenix pings).

If we were to copy existing nonprod telemetry pings to prod, then we'd have older data with snake-cased additional properties alongside new pings with casing preserved.

whd:
> If we were to copy existing nonprod telemetry pings to prod, then we'd have older data with snake-cased additional properties alongside new pings with casing preserved.

I imagine the prod migration not involving this, since we'll be backfilling data from AWS ala https://bugzilla.mozilla.org/show_bug.cgi?id=1568042 for data from before the final prod stacks validation occurs.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -252,21 +252,25 @@ public void transformForBqSchema(Map<String, Object> parent, List<Field> bqField
       Map<String, Object> additionalProperties) {
 
     // Clean the key names.
-    ImmutableSet.copyOf(parent.keySet())
-        .forEach(key -> parent.put(getAndCacheBqName(key), parent.remove(key)));
+    final Map<String, String> jsonFieldNames = new HashMap<>();
+    ImmutableSet.copyOf(parent.keySet()).forEach(key -> {
+      String transformedKey = getAndCacheBqName(key);
+      jsonFieldNames.put(transformedKey, key);
```
acmiyaguchi: It's technically possible for a name collision to occur in separate branches of a document e.g. `environment.someId` and  `payload.Some_ID`. I think the chances of seeing this in the wild are low, especially considering that this only affects additional_properties in a single document.

acmiyaguchi: A logging statement could be one way of warning about a name collision, although it only needs to be emitted as the field is inserted into additional properties. It might also be possible to push the renaming logic closer to insertion into the bq field to avoid this case.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -136,13 +136,13 @@ public void testAdditionalProperties() throws Exception {
     parent.put("outer", new HashMap<>(ImmutableMap.of("otherField", 3, //
         "mapField", new HashMap<>(ImmutableMap.of("foo", 3, "bar", 4)))));
     parent.put("clientId", "abc123");
-    parent.put("otherStrangeIdField", 3);
+    parent.put("otherStrangeIDField", 3);
```
relud: this change seems unrelated to the scope, just cleaning up along the way?

jklukas: I think I made this change to exercise another edge case of snake casing, but in this case the expected behavior is that the field should be left unchanged and put in additional_properties. I may go ahead and revert it.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -252,21 +252,25 @@ public void transformForBqSchema(Map<String, Object> parent, List<Field> bqField
...
+      parent.put(transformedKey, parent.remove(key));
```
relud: I think we should build additionalProperties while iterating through parent, both to only iterate once and to fix the issue from @acmiyaguchi's [comment](https://github.com/mozilla/gcp-ingestion/pull/713/files#r307911925) by using the original name in the first place.
```suggestion
      Object value = parent.remove(k);
      if (fieldNames.contains(transformedKey) {
        parent.put(transformedKey, value);
      } else if (additionalProperties != null) {
        additionalProperties.put(key, value);
      }
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: This shouldn't actually be an issue, because `jsonFieldNames` isn't passed recursively, so `environment.someId` and `payload.Some_ID` would have different local maps for `jsonFieldNames`.

Regardless, I think it would be better to build `additionalProperties` within this loop, instead of a separate loop, as per [this comment](https://github.com/mozilla/gcp-ingestion/pull/713/files#r308373476)

acmiyaguchi: You're right, it looks like name collisions across different namespaces is not an issue.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
jklukas: I'm going to pull this into a single loop as suggested.

jklukas:
Thanks for the review here. There was indeed a more elegant and less lossy way of expressing this by moving more logic inside a single loop. I expect this is in a good spot now, but will wait on merging until after tomorrow's data validation exercise.

PR #719, opened 2019-07-30
Bug: Use snake-cased docType in BQ sink

jklukas:
Discovered during validation exercise today.


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -126,7 +126,7 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
     // but some doc types and namespaces contain '-', so we convert to '_'; we don't pass all
     // values through getAndCacheBqName to avoid expensive regex operations and polluting the
     // cache of transformed field names.
-    attributes = Maps.transformValues(message.getAttributeMap(), v -> v.replaceAll("-", "_"));
+    attributes = Maps.transformValues(attributes, v -> v.replaceAll("-", "_"));
```
jklukas: `attributes` contains the snake-cased docType, but we were ignoring it here and using the unmodified attributes.

whd:
Looks like this needs a test fixup, otherwise LGTM.

PR #720, opened 2019-07-30
Add format option for sinking raw Pub/Sub messages BigQuery

relud:


relud:
> For the purposes of the POC do we want the destination output to match by interpolation as it currently does

I would prefer to see decoded topics output to tables named like `{namespace}_raw.{document_type}_v{document_version}`, e.g. `telemetry_raw.main_v4` 

for raw topics we could set non-templated names like `landfill_raw.telemetry_v1`, `landfill_raw.structured_v1`, `landfill_raw.tls_error_reports_v1`, `landfill_raw.stub_installer_v1`

> For clustering I imagine we would keep the same clustering we have now by timestamp.

sounds good to me

whd:
> I would prefer to see decoded topics output to tables named like `{namespace}_raw.{document_type}_v{document_version}`, e.g. `telemetry_raw.main_v4`

Ok, so I think we need a bq schema mapping the pubsub schema for the purposes of table management. I'll adjust the table management in stage to create these new datasets for "raw" data. Since it's UTC tomorrow now we should plan on landing this sometime tomorrow (before UTC Thursday). We can potentially also create the subscription and seek backwards so that we can process a full window of data more quickly.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -108,6 +109,13 @@
 
   void setOutputFileFormat(OutputFileFormat value);
 
+  @Description("File format for --outputType=bigquery; must be one of"
+      + " raw (each row contains payload[Bytes] and attributeMap[String,String]) or"
```
jklukas:
```suggestion
      + " raw (each row contains payload[Bytes] and selected attributes as additional fields) or"
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -139,6 +149,14 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
 
     final TableDestination tableDestination = new TableDestination(tableSpec, null,
         TIME_PARTITIONING);
+
+    if (tableRowFormat != null && tableRowFormat.get() == TableRowFormat.raw) {
+      TableRow tableRow = new TableRow();
+      AddMetadata.attributesToMetadataPayload(message.getAttributeMap()).forEach(tableRow::set);
```
jklukas: `AddMetadata.attributesToMetadataPayload` doesn't currently include namespace, doctype, or version, which will be probably the most interesting information for the landfill_raw cases. We may want to add those to `URI_ATTRIBUTES` so that they appear under `metadata.uri.document_type`, etc.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -139,6 +149,14 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
 
     final TableDestination tableDestination = new TableDestination(tableSpec, null,
         TIME_PARTITIONING);
+
+    if (tableRowFormat != null && tableRowFormat.get() == TableRowFormat.raw) {
```
jklukas:
```suggestion
    if (tableRowFormat.get() == TableRowFormat.raw) {
```

We should be guaranteed this is non-null after the check in the constructor, right?

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -139,6 +149,14 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
...
+      TableRow tableRow = new TableRow();
+      AddMetadata.attributesToMetadataPayload(message.getAttributeMap()).forEach(tableRow::set);
+      tableRow.set("payload", message.getPayload());
```
jklukas: I'm pretty sure the payload bytes here are unzipped. We should zip before setting the payload field on the TableRow.

@scholtzan wrote a JS UDF to unzip bytes, so we should be able to peek at these zipped payloads in BigQuery. I'll follow up with her about getting that UDF integrated into bigquery-etl.

jklukas: See https://github.com/mozilla/bigquery-etl/issues/270

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -139,6 +149,14 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
...
+      return KV.of(tableDestination, tableRow);
```
jklukas: The early return here makes me think we should factor out the rest of this method to a separate `processDecoded` method or the like. I think at least this should move down below the "Resolved destination table does not exist" check.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: we should set `--decompressInputPayloads=false` to preserve the compression we get from pubsub

relud: @whd ☝️

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: yeah, i agree. i just didn't like the size of diff in created 😭

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
whd: ACK

whd:
For the current GCS error sinks (decoder validation failures), replacing them with bq sinks to error tables seems fairly straightforward. I was planning to have a broader discussion about this yesterday (i.e. should we do this at all, should have first-class support for the errors schema in BQ instead of just using the pubsub schema etc.).

An additional consideration is what to do with the error streams from the beam jobs themselves (currently all output to GCS aside from the decoder), since I assume they use the same code path we're trying to replace. For prototyping it is irrelevant, and these sinks should be low volume and never the source of truth for a particular message, so leaving them as GCS-based is perhaps the way to go (esp. for the bq sinks themselves). We could also potentially create a pubsub topic for beam errors and a separate dataset/table to write them to.

whd:
We're potentially going to see a proliferation of datasets here. The current plan is to use `{namespace}_raw` for the decoded sinks. We'll also have a bq sink for actual landfill (pre-decoder), which should definitely go to its own dataset because it needs separate ACLs. With potentially also adding errors tables, we should consider once again naming of tables and datasets for these. It may be possible to consolidate under existing namespace/datasets but we should think carefully about this.

jklukas:
I initially thought that we'd best be served by having a single table per sink even for decoded messages, and encoding namespace, doctype, version as attributes. But there's real benefit in terms of querying efficiency by putting these in separate tables. We could also cluster by namespace, doctype, version, but that would require reprocessing daily (as we are going to do for live -> stable). By putting separate doctypes into different tables, we get the benefits of clustering without having to do that daily reprocessing.

I do worry about the layout of datasets/tables getting increasingly complex. Is it worth considering a separate project for this raw data? `shared-prod-raw` perhaps?

We could also choose to have a single `decoded_raw` dataset, and dump all tables there with names like `${document_namespace}__${document_type}_${document_version}` (using a double underscore between namespace and doctype). This would also have the benefit of allowing easy counts across all tables in all namespaces using wildcards (since you cannot use wildcards for datasets): `select _TABLE_PREFIX AS src, count(*) from decoded_raw.* GROUP BY 1`. That would at least reduce the number of datasets for these raw payload tables.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: it includes those in `IDENTIFYING_FIELDS`, which are put directly in `metadata`. I've added expanded the test to demonstrate this.

that said, `AddMetadata.attributesToMetadataPayload` won't work for landfill, so I've made a distinction between `raw`, `decoded`, and `payload` formats.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: turns out I wasn't setting the default properly in `SinkOptions`. I fixed that and we no longer need to check for null here or in the constructor.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: fixed

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
jklukas: Very good. I had totally forgotten about `IDENTIFYING_FIELDS`.

relud:
> We could also choose to have a single `decoded_raw` dataset

I would rather see three datasets: `raw`, `decoded`, `error`

> dump all tables there with names like `${document_namespace}__${document_type}_${document_version}`

+1 for that table naming convention.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -108,6 +109,14 @@
 
   void setOutputFileFormat(OutputFileFormat value);
 
+  @Description("Row format for --outputType=bigquery; must be one of"
+      + " raw (each row contains payload[Bytes] and attributes as top level fields) or"
+      + " decoded (each row contains payload[Bytes] and attributes as nested metadata fields) or"
+      + " payload (each row is extracted from payload); defaults to payload")
+  ValueProvider<TableRowFormat> getOutputTableRowFormat();
```
relud: reflagging for review because I added a format and changed the format names to be more like the pubsub topic names.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -235,4 +238,76 @@ public void testNestedRepeatedStructs() throws Exception {
...
+                    + "/release/20180206200532")
+            .put("protocol", "HTTP/1.1").put("method", "POST").put("args", "v=4")
+            .put("remote_addr", "172.31.32.5").put("content_length", "4722")
+            .put("date", "Mon, 12 Mar 2018 21:02:18 GMT").put("dnt", "1")
+            .put("host", "incoming.telemetry.mozilla.org").put("user_agent", "pingsender/1.0")
+            .put("x_forwarded_for", "10.98.132.74, 103.3.237.12").put("x_pingsender_version", "1.0")
+            .put("x_debug_id", "my_debug_session_1")
+            .put("x_pipeline_proxy", "2018-03-12T21:02:18.123456Z").build());
+    String expected = Json.asString(ImmutableMap.<String, String>builder()
+        .putAll(message.getAttributeMap()).put("payload", "dGVzdA==").build());
```
jklukas: The payload is showing up as base-64 encoded in the json. Do we understand where that encoding is happening? Are we sure that what we end up being inserted into BigQuery is the bytes rather than the encoded string? We should probably add a new case in BigQueryIntegrationTest to validate we get what we expect in a BYTES column.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -177,15 +183,54 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
       throw new IllegalArgumentException("Resolved destination table does not exist: " + tableSpec);
     }
 
+    TableRow tableRow;
```
jklukas:
```suggestion
    final TableRow tableRow;
```

I know Java supports checking that a `final` variable is set within all branches of an if statement; I assume we can get the same checking in a switch? Maybe that's not actually supported.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
relud: > Do we understand where that encoding is happening?

`Json.asString(tableRow)` will encoded a bytes field as base64, because bytes isn't a valid json type

> We should probably add a new case in BigQueryIntegrationTest to validate we get what we expect in a BYTES column.

good point

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -177,15 +183,54 @@ private PubsubMessageToTableRow(ValueProvider<String> tableSpecTemplate,
...
+        break;
+      case decoded:
+        tableRow = decodedTableRow(message);
+        break;
+      case payload:
+      default:
+        tableRow = payloadTableRow(message, ref, tableSpec, namespace, docType);
+        break;
+    }
+    return KV.of(tableDestination, tableRow);
```
jklukas: This is very nice as a switch statement 👍

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: looks like it works!

whd:
Current plan: 3 datasets (error, raw, decoded) prefixed by `payload_bytes_` (naming subject to change). Table name will also be prefixed with "pipeline family" (e.g. structured, stub_installer, telemetry) to map to current infrastructure topology.

Decoded datasets will have a table per docType and docVersion. Raw datasets will only be segmented by "pipeline family", so a single table per pipeline family. It seems like it would be useful to have error datasets segmented by document type and version, but since there are cases where the failure can happen before type and version are determined, we might want to stick these all in a single table per namespace or perhaps leverage UNKNOWN for missing attributed if that's a possibility.

All these tables will be clustered by submission_timestamp since they're being populated by ingestion infrastructure.

@relud will provide schemas for the various table archetypes (raw, decoded, error).

relud:
> @relud will provide schemas for the various table archetypes (raw, decoded, error).

https://gist.github.com/relud/453b96ee5795764a271ce048f0a7a9e3

PR #728, opened 2019-08-07
Add support for --bqStreamingDocTypes=*

relud:
cc @whd

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
@@ -374,13 +374,17 @@ public abstract class Write
             .apply("PartitionStreamingVsFileLoads", Partition.of(2, //
                 (message, numPartitions) -> {
                   message = PubsubConstraints.ensureNonNull(message);
-                  final String namespace = message.getAttribute("document_namespace");
-                  final String docType = message.getAttribute("document_type");
                   final boolean shouldStream;
-                  if (namespace == null || docType == null) {
-                    shouldStream = false;
+                  if (streamingDocTypes.get().contains("*")) {
+                    shouldStream = true;
                   } else {
-                    shouldStream = streamingDocTypes.get().contains(namespace + "/" + docType);
+                    final String namespace = message.getAttribute("document_namespace");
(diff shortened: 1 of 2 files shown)


relud:
@whd I'd like to request that all the bq sink jobs spend a full UTC day in nonprod with `--bqStreamingDocTypes=*` and with `--diskSizeGb` and `--workerDiskType` set the same as on the decoder, so that we can get an accurate cost estimate

whd:
> @whd I'd like to request that all the bq sink jobs spend a full UTC day in nonprod with `--bqStreamingDocTypes=*` and with `--diskSizeGb` and `--workerDiskType` set the same as on the decoder, so that we can get an accurate cost estimate for comparison

Ok, some of that is configured [here](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/bin/build-template) and some [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/beam/Jenkinsfile.build#L83). It's probably best to:

1. Put the parameters changes for the bq sink template build in a branch in this repo
2. Put the `bqStreamingDocTypes=*` change in a branch in the infra repo
3. Run the template build job and deploy jobs referencing those branches for nonprod

Likely we should aim for UTC Friday to be the full day (i.e. roll this out tomorrow).

relud:
1. okay, named it `deploy-all-streaming`
2. https://github.com/mozilla-services/cloudops-infra/pull/1261
3. waiting until tomorrow

jklukas:
> Likely we should aim for UTC Friday to be the full day (i.e. roll this out tomorrow).

I don't think we need to use midnight as the cutoff. If we get this deployed relatively early in the day today, we can choose any hour after it gets caught up as the start of the validation window.

relud:
> I don't think we need to use midnight as the cutoff.

I expect the gcp billing dashboard to be giving me UTC days, but i'll double-check

relud:
https://cloud.google.com/billing/docs/how-to/reports

> Time periods in the billing report use the Pacific Time Zone, and observe daylight saving time shifts.

apparently I need a PDT day

whd:
Welp, I guess the 9th PDT will be the day. I can start this on the earlier side anyway if desired so we can debug any issues.

Issue #729, opened 2019-08-08
mvn fails when run from the top level


jklukas:
@wlach reports that he ran into several different failures when trying to onboard on this project. We eventually realized that he was running the top-level `bin/mvn` script which we hadn't really exercised.

We should make sure that it's possible to run mvn at the top level and correctly build/test the sub-projects.

jklukas:
Also, it's not clear that you run some scripts to download files before being able to successfully execute tests.

relud:
running this from the git root works, but note that it requires `-pl ingestion-beam` to not run the exec plugin for all modules, including the parent:

```bash
echo '{"attributeMap":{"remote_addr":"63.245.208.195"},"payload":"dGVzdA=="}' > input.json
./bin/mvn -pl ingestion-beam exec:java -Dexec.args="--input=input.json --inputType=file --outputType=stdout --errorOutputType=stderr"
```

running `./bin/mvn test` from the root also works fine for me when i have the dependencies downloaded

relud:
I believe this is working now

PR #731, opened 2019-08-09
Fix an omission re: running ingestion-beam tests

wlach:
Add a note about downloading test data before running ingestion-beam
tests, see https://github.com/mozilla/gcp-ingestion/issues/729#issuecomment-519557808

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/heka/File.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/heka/File.java
@@ -1,0 +1,4 @@
+package com.mozilla.telemetry.heka;
+
+public class File {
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/heka/File.java
```diff
@@ -0,0 +1,4 @@
+package com.mozilla.telemetry.heka;
```
jklukas: I assume including this file in the PR is accidental?

wlach: argh, how did that get in there :) will remove

PR #732, opened 2019-08-09
Create an mkdocs-based site

wlach:
This substantially reorganizes the documentation as an [mkdocs](https://mkdocs.org)-based site. Main changes:

* All documentation is now browseable and searchable in a single site, with handy table of contents on the side of each section
* Top-level README significantly slimmed down (just pointing to docs site)
* READMEs inside individual components removed (moved to subdirectories inside docs/ folder, accessible via top-level in docs site)

I also made various editorial changes to the documentation to make it read a little better in the new format. You can see a preview of the changes here: 

https://wlach.github.io/gcp-ingestion/

No doubt further improvements are possible and perhaps some parts of this should move to dtmo as discussed, but I think this new format is substantially easier to navigate, especially for newcomers to the project.


jklukas:
Perhaps mkdocs can handle building tables of contents? As it is, looks like we need to run `./bin/update-toc` and there are some spelling issues. Now, I dive into a more careful reading :)

mkdocs.yml
```diff
@@ -0,0 +1,18 @@
+site_name: GCP Ingestion
+nav:
+  - Home: index.md
+  - Edge Ingestion: ingestion-edge/index.md
+  - Beam Ingestion:
```
jklukas: The terms "Edge Ingestion" and "Beam Ingestion" aren't used anywhere else. I think of `ingestion-edge` and `ingestion-beam` as the formal names for the codebases, so ideally those would be the section names. Or perhaps "Edge Service" and "Beam Pipelines" if the codebase names are ruinous to the aesthetics of the sitenav.

docs/architecture/decoder_service_specification.md
```diff
@@ -69,7 +53,7 @@ required group attributes {
 
 ### Raw message schema
 
-See [Edge Server PubSub Message Schema](edge.md#edge-server-pubsub-message-schema).
```
jklukas: I'd prefer "edge service" to "edge server" as the latter sounds like it's referring to a single machine, although if this is the terminology already used in the docs, I'm fine with keeping it.

wlach:
> The rendered docs look great. What would you think about using the same theme as docs.tmo for consistency?

docs.tmo uses the rust-based mdBook for generating documentation, so we'd have to switch to using that (unless we wanted to port the theme over, which I suspect we don't). 

I guess we could? I'm not sure if I see a big advantage to using its theme for this particular use case. I feel like the mkdocs-based site is a little easier to understand/navigate for a document like this one, where there are only a few high-level topics.

> Only blocker for approval I see is getting CI into shape. It looks like we need to rip out the `update-toc` script and remove that check from the .circleci config and we need to integrate running of mkdocs and publishing the result. Or does GitHub Pages magically do that for us?

Yeah, I don't think we need the `update-toc` script anymore, as the rendered documentation will have table of contents generated automatically. Definitely we will need to update circle to support this, but I expect that'll be relatively straightforward. Iodide already does something like this, so we can likely just copy its circle tasks over.

jklukas:
> docs.tmo uses the rust-based mdBook for generating documentation, so we'd have to switch to using that (unless we wanted to port the theme over, which I suspect we don't).

I got confused and forget that mdbook and mkdocs are different things. Carry on. 😄

mkdocs.yml
wlach: wfm, this is mainly developer focused so this makes sense

docs/architecture/decoder_service_specification.md
wlach: Sure thing, this would be a change from how things were originally written, but no reason not to do that cleanup now.

jezdez:
If you'd like to keep it close to the DTMO theme, there is also a custom theme that mimics GitBook: https://lramage.gitlab.io/mkdocs-gitbook-theme/

Other than that I encourage you to use the Material theme given its better typography options: https://squidfunk.github.io/mkdocs-material/

wlach:
> If you'd like to keep it close to the DTMO theme, there is also a custom theme that mimics GitBook: https://lramage.gitlab.io/mkdocs-gitbook-theme/
> 
> Other than that I encourage you to use the Material theme given its better typography options: https://squidfunk.github.io/mkdocs-material/

I have to say, having just tried the gitbook plugin and looked at examples of the material ui one, I think I kinda prefer the default mkdocs theme for this project. There's a few reasons:

1. It's easy to navigate to any section of the document via the top-level navbar (instead of having to "reveal" a navigation bar via a hamburger menu, or look through a very long table of contents on the sidebar).
2. Code and script chunks are formatted to be easy to read, as opposed to on gitbook where there is minimal contrast.

That said, if people really think it's important we be consistent with dtmo (I don't personally think that's a big deal), I can switch it over. Here's what it would look like:

image

jklukas:
> That said, if people really think it's important we be consistent with dtmo (I don't personally think that's a big deal), I can switch it over.

I don't have strong feelings on the matter. If you think the navigation experience will be better with the current theme, that feels like solid rationale for a decision.

README.md
```diff
@@ -1,19 +1,16 @@
...
-under the various `ingestion-*` service directories:
+There are currently two components:
 
 - [ingestion-edge](ingestion-edge): a simple Python service for accepting HTTP
   messages and delivering to Google Cloud Pub/Sub
 - [ingestion-beam](ingestion-beam): a Java module defining
   [Apache Beam](https://beam.apache.org/) jobs for streaming and batch
   transformations of ingested messages
+
+For more information, see [mozilla.github.io/gcp-ingestion](the documentation).
```
relud:
```suggestion
For more information, see [the documentation](https://mozilla.github.io/gcp-ingestion).
```

PR #735, opened 2019-08-12
Add support for reading from BigQuery

jklukas:
cc @whd


ingestion-beam/src/main/java/com/mozilla/telemetry/options/InputType.java
```diff
@@ -24,6 +25,26 @@ public Read read(SinkOptions.Parsed options) {
...
+
+  },
+
+  bigquery_query {
+
+    /** Return a PTransform that reads the results of a BigQuery query. */
+    public Read read(SinkOptions.Parsed options) {
+      return new BigQueryInput(options.getInput(), options.getBqReadMethod(),
+          BigQueryInput.Source.QUERY, options.getBqRowRestriction(), options.getBqSelectedFields());
+    }
```
jklukas: I chose to implement 2 new input types rather than a single `bigquery` input type, allowing us to distinguish between reading from a table vs. a query without introducing an additional compile-time parameter.

ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -66,6 +66,33 @@
...
+
+  void setBqRowRestriction(String value);
+
+  @Description("When --bqReadMethod=storageapi, all fields of the input table are read by default,"
+      + " but this option can take a comma-separated list of field names, in which case only the"
+      + " listed fields will be read, saving costs; when reading decoded payload_bytes, none of the"
+      + " metadata fields are needed, so setting --bqSelectedFields=payload is recommended")
+  List<String> getBqSelectedFields();
+
+  void setBqSelectedFields(List<String> value);
```
jklukas: Note that all three of these new options are compile-time only. I suspect that the latter two could be specified at runtime, but no `ValueProvider` interface exists. We could consider filing an upstream PR; see #736

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
```diff
@@ -368,7 +368,7 @@ public void canLoadRawFormat() throws Exception {
   }
 
   @Test
-  public void canLoadDecodedFormat() throws Exception {
+  public void canLoadAndReadDecodedFormat() throws Exception {
```
jklukas: I gave this existing test more responsibility rather than factoring out a separate test, since this already provides the necessary logic for putting realistic decoded bytes rows into BigQuery, which is exactly the format we need to exercise the new code.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
```diff
@@ -450,6 +451,21 @@ public void canLoadDecodedFormat() throws Exception {
             .build()),
         String.join("\n",
             stringValuesQueryWithRetries("SELECT TO_JSON_STRING(t) FROM " + tableSpec + " AS t")));
+
+    String fileOutput = outputPath + "/out";
+    PipelineResult readResult = Sink.run(new String[] { "--inputType=bigquery_table",
+        "--input=" + fullyQualifiedTableSpec, "--project=" + projectId, "--bqReadMethod=storageapi",
+        "--outputFileCompression=UNCOMPRESSED", "--bqSelectedFields=payload",
+        "--bqRowRestriction=CAST(submission_timestamp AS DATE)"
+            + " BETWEEN CAST('2020-01-10' AS DATE) AND CAST('2020-01-14' AS DATE)",
```
jklukas: The storage API is quite picky about how the row restriction is expressed. It looks like the explicit cast syntax is necessary rather than the simpler `DATE(expr)` syntax.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -56,4 +67,69 @@ public FileInput(ValueProvider<String> fileSpec, InputFileFormat fileFormat) {
...
+      BigQueryIO.TypedRead<PubsubMessage> read = BigQueryIO
+          .read((SchemaAndRecord schemaAndRecord) -> {
+            TableSchema tableSchema = schemaAndRecord.getTableSchema();
+            GenericRecord record = schemaAndRecord.getRecord();
+            byte[] payload = ((ByteBuffer) record.get("payload")).array();
+
+            // We copy only string-type fields to attributes, which is complete for raw and
+            // error tables; decoded payload tables have the nested metadata object also encoded in
+            // the payload, so we can safely drop the metadata object here and rely on ParsePayload
+            // to parse attributes from it.
```
relud: > rely on ParsePayload to parse attributes from it.

but the decoder runs `ParseUri` before `ParsePayload` and throws `NullUriException` if the `uri` attribute is missing. should we attempt to extract just `uri` from the `metadata` object?

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
relud: I could have sworn that the spark connector worked when I used date literals (e.g. both `submission_date = DATE '2019-01-01'` and `submission_date = '2019-01-01'` worked), even if `CAST(submission_timestamp AS DATE)` is required

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
jklukas: Good catch!

We still have the issue of BQ streaming inserts failing and getting put to error output, in which case we lose all attributes, and we only have the metadata to refer to. So I think we need to solve that case as well. It's not immediately clear to me how best to handle that.

jklukas: Perhaps we should remove the `NullUriException` and allow message to pass through `ParseUri` unchanged in case the "uri" attribute is missing. Then, in ParsePayload, after the payload is parsed and we call `AddMetadata.stripPayloadMetadataToAttributes`, we should be good.  We could add a sanity check at that point to make sure that doctype and namespace are present and throw an exception if missing.

relud: > We still have the issue of BQ streaming inserts failing and getting put to error output, in which case we lose all attributes, and we only have the metadata to refer to. So I think we need to solve that case as well. It's not immediately clear to me how best to handle that.

for streaming inserts we get errors, could we restore `uri` when we [turn the row back into a pubsub message](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java#L410-L433)?

for file loads do we just...not get errors? I don't see error handling after [insert](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java#L446-L450), so I assume they crash the job?

jklukas: > for file loads do we just...not get errors? I don't see error handling after insert, so I assume they crash the job?

Yes, it's my expectation that they would crash the job and bubble up.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
jklukas: It looks like you're right. I tested using `DATE('2019-01-01')`, which didn't work, but looks like the even simpler string literal form does work for the date literals.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
relud: > Perhaps we should remove the `NullUriException` ... Then, in `ParsePayload`, after the payload is parsed and we call `AddMetadata.stripPayloadMetadataToAttributes`, we should be good.

the reason we throw `NullUriException` is we don't want to gunzip & json parse if we don't know what schema to validate.

relud: but actually, i think if `uri` is missing, this is fine, because we know we did that

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -89,15 +89,15 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+
     if (MessageScrubber.shouldScrub(attributes, json)) {
       // Prevent the message from going to success or error output.
       throw new MessageShouldBeDroppedException();
     }
 
-    // In case this message is being replayed from an error output where AddMetadata has already
-    // been applied, we strip out any existing metadata fields and put them into attributes.
-    AddMetadata.stripPayloadMetadataToAttributes(attributes, json);
-
```
jklukas: This is a semi-related change as I'm scrutinizing order of operations. We may want to be able to scrub messages when reading from previously decoded output, in which case we probably want attributes to be available.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -94,7 +91,10 @@ protected PubsubMessage processElement(PubsubMessage message) throws InvalidUriE
     // parse uri based on prefix
     final String uri = attributes.get("uri");
     if (uri == null) {
-      throw new NullUriException();
```
jklukas: We now allow null uris.

jklukas:
Requesting review again to make sure the changes for null URI handling look good.

PR #738, opened 2019-08-13
Implement Streaming BQ Sink for Kubernetes

relud:
TODO:
- [x] `ingestion-sink` job in CI needs to be fixed
- [x] `PubsubReadTest` needs credentials or PUBSUB_EMULATOR_HOST in CI
- [x] Integration test for `Sink`
- [x] Unit test for `BigQuery.Write` with mocked `BigQuery`
- [x] `mozilla/ingestion-sink` docker container needs to be published

@whd the config options rn are `INPUT_SUBSCRIPTION` (fully qualified e.g. `projects/test/subscriptions/test`), `OUTPUT_TABLE` (templated just like in beam) and `OUTPUT_FORMAT` (optional, `raw` is the only format implemented right now)

Closes #730


whd:
[WIP infra branch](https://github.com/mozilla-services/cloudops-infra/compare/data_kube_sink?expand=1)

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+public class BigQuery {
+
+  public static class Write implements Function<Write.TableRow, CompletableFuture<Write.TableRow>> {
+
+    private final com.google.cloud.bigquery.BigQuery bigquery;
+    private final Map<TableId, Batch> batches = new HashMap<>();
+    private final int maxBytes = 10000000; // HTTP request size limit: 10 MB
```
jklukas:
```suggestion
    private final int maxBytes = 10_000_000; // HTTP request size limit: 10 MB
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+    private final int maxMessages = 10000; // Maximum rows per request: 10,000
```
jklukas:
```suggestion
    private final int maxMessages = 10_000; // Maximum rows per request
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+    private final long maxDelayMillis = 1000; // default to 1 second
```
jklukas:
```suggestion
    private final long maxDelayMillis = 1000; // Default to 1 second
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.Optional;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+
+public class BigQuery {
+
+  public static class Write implements Function<Write.TableRow, CompletableFuture<Write.TableRow>> {
+
+    private final com.google.cloud.bigquery.BigQuery bigquery;
+    private final Map<TableId, Batch> batches = new HashMap<>();
```
jklukas: This should probably be a ConcurrentHashMap if it's being accessed and mutated by multiple threads.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+    private final int maxMessages = 10000; // Maximum rows per request: 10,000
+    private final long maxDelayMillis = 1000; // default to 1 second
+
+    public Write(com.google.cloud.bigquery.BigQuery bigquery) {
+      this.bigquery = bigquery;
+    }
+
+    @Override
+    public CompletableFuture<TableRow> apply(TableRow row) {
+      return Optional.ofNullable(batches.get(row.tableId)).flatMap(batch -> batch.add(row))
```
jklukas: If we use a ConcurrentHashMap, this could be replaced with a call to `batches.computeIfAbsent(row.tableId, tableId -> {create new batch})` to check and create the batch atomically.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+
+    private class Batch {
+
+      private final com.google.cloud.bigquery.BigQuery bigquery;
+      private final InsertAllRequest.Builder builder;
+      private final CompletableFuture<Void> full;
+      private final CompletableFuture<InsertAllResponse> result;
+
+      int size = 0;
+      int byteSize = 0;
```
jklukas: Getting synchronized access right is always a little scary, but it looks like you're doing a reasonable thing here and only accessing or mutating these from synchronized methods.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+      private void timeout() {
+        try {
+          Thread.sleep(maxDelayMillis);
+        } catch (InterruptedException e) {
+          // this is fine
+        }
+      }
+
+      private synchronized InsertAllResponse insertAll(Void v, Throwable t) {
+        return bigquery.insertAll(builder.build());
```
jklukas: It's unclear to me what happens after we call insertAll. I expected that we'd either reset state on this Batch and reuse the object for more rows, or we'd remove this Batch from BigQuery's batches map, so that the next row for this table creates a new batch. But I don't see either of those happening.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
relud: TIL

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
relud: in `BigQuery.Write.apply` I use `flatMap` to start a new batch when `batch.add(row)` returns `Optional.empty()`:
```
    public CompletableFuture<TableRow> apply(TableRow row) {
      return Optional.ofNullable(batches.get(row.tableId)).flatMap(batch -> batch.add(row))
          .orElseGet(() -> {
            final Batch batch = new Batch(bigquery, row.tableId);
            final CompletableFuture<TableRow> response = batch.add(row)
                .orElseThrow(() -> new IllegalArgumentException("Empty batch rejected message"));
            batches.put(row.tableId, batch);
            return response;
          });
    }
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+      private synchronized Optional<CompletableFuture<TableRow>> add(TableRow row) {
+        if (full.isDone()) {
+          return Optional.empty();
+        }
+        if (size > 0 && (size + 1 > maxMessages || byteSize + row.byteSize + 1 > maxBytes)) {
+          this.full.complete(null);
+          return Optional.empty();
+        }
+        size += 1;
+        byteSize += row.byteSize + 1;
```
jklukas: Can you add a comment to explain why we add the extra `1` byte here?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+
+      private synchronized InsertAllResponse insertAll(Void v, Throwable t) {
+        return bigquery.insertAll(builder.build());
+      }
+
+      private synchronized Optional<CompletableFuture<TableRow>> add(TableRow row) {
+        if (full.isDone()) {
+          return Optional.empty();
+        }
+        if (size > 0 && (size + 1 > maxMessages || byteSize + row.byteSize + 1 > maxBytes)) {
```
jklukas:
```suggestion
        if (size > 0 && (size >= maxMessages || (byteSize + row.byteSize) >= maxBytes)) {
```

jklukas: Actually, how about we delay this check until we've already incremented these variables so we avoid having to write out the additions twice?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,124 @@
...
+          this.full.complete(null);
+          return Optional.empty();
+        }
+        size += 1;
+        byteSize += row.byteSize + 1;
+        builder.addRow(row.content);
```
jklukas: I'd advocate that we pull out `document_id` if available and set it as the message ID here so that the BigQuery streaming service can deduplicate if we double-publish.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
jklukas: Interesting. That's subtle enough that I'd recommend making sure the `flatMap` is on its own line, and have a comment above the line to explain that Batch::add will return Optional.empty() as a special flag to indicate it's done and a new batch should be created.

relud: it's different now that i'm using `ConcurrentMap`, so it's more obvious

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
relud: I'm against having the values be incorrect by one message when `insertAll` runs, so I'll put the calculated values in variables and reuse those.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -0,0 +1,130 @@
...
+    private final int maxMessages = 10_000; // Maximum rows per request: 10,000
+    private final long maxDelayMillis = 1000; // Default to 1 second
+
+    public Write(com.google.cloud.bigquery.BigQuery bigquery) {
+      this.bigquery = bigquery;
+    }
+
+    @Override
+    public CompletableFuture<TableRow> apply(TableRow row) {
+      CompletableFuture<CompletableFuture<TableRow>> future = new CompletableFuture<>();
```
jklukas: AFAICT, the outer future here is just a wrapper that allows us to modify a value inside the lambda below. This is a bit difficult to parse due to the nested futures, so perhaps using an AtomicReference would be better:

```
      AtomicReference<CompletableFuture<TableRow>> futureRef = new AtomicReference<>();
      batches.compute(row.tableId, (tableId, batch) -> {
        if (batch != null) {
          batch.add(row).ifPresent(futureRef::set);
        }
        if (futureRef.get() == null) {
          batch = new Batch(bigquery, row.tableId);
          batch.add(row).ifPresent(futureRef::set);
        }
        return batch;
      });
      return Optional.ofNullable(futureRef.get())
          .orElseThrow(() -> new IllegalArgumentException("Empty batch rejected message"));
```

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/SinkIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    });
+
+    environmentVariables.set("INPUT_SUBSCRIPTION", subscriptionName.toString());
+    environmentVariables.set("OUTPUT_TABLE",
+        project + "." + dataset + ".${document_type}_v${document_version}");
+
+    CompletableFuture<Void> main = CompletableFuture.runAsync(() -> Sink.main(new String[] {}));
+
+    String query = "SELECT * REPLACE (FORMAT_TIMESTAMP('%FT%R:%E*SZ', submission_timestamp) AS "
+        + "submission_timestamp), _TABLE_SUFFIX AS table FROM `" + project + "." + dataset + ".*`";
```
jklukas:
```suggestion
        + String.format("submission_timestamp), _TABLE_SUFFIX AS table FROM `%s.%s.*`", project, dataset);
```

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/SinkIntegrationTest.java
```diff
@@ -0,0 +1,146 @@
...
+    AtomicReference<List<List<String>>> actual = new AtomicReference<>();
```
jklukas: I don't think there's any need for an AtomicReference here. We're mutating `actual` within the lambda expression, but not assigning a new value, so there shouldn't be any compiler errors to worry about.

relud: it doesn't compile if i unnest it because two lines down I make a new list (`actual.set(new LinkedList<>());`), and variables referenced in a lambda "have to be final or effectively final".

jklukas: Missed that 👍

PR #747, opened 2019-08-14
Allow BQ API rate limit exceptions to bubble up

### LINKED ISSUE #745: Pipeline should throw exception on BQ API rate limiting
[As discussed in the GCP migration validation doc](https://docs.google.com/a/mozilla.com/document/d/14UoZGAI2CgMQW4qeXIX5Bj9nq2kvfCUaONiNknXXxYE/edit?disco=AAAADV_ZM-A), we are still occasionally hitting BQ API rate limits, causing records to go to error output. We want the exception to bubble up instead, causing the batch to be retried.

jklukas:
Closes #745

Introduces the concept of a `BubbleUpException` as a way of breaking out of MapElementsWithErrors catching transient exceptions and forwarding messages to error output.


jklukas:
> i'm concerned that we will lose data if this happens, due to the pipeline failing after checkpointing and without draining to any destination

Fair enough, though we're already vulnerable to that situation for OOM exceptions and other resource constraints.

PR #749, opened 2019-08-14
Implement decoded format for ingestion-sink

relud:
cc @whd (configured by setting `OUTPUT_FORMAT=decoded`)


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/util/Attribute.java
```diff
@@ -0,0 +1,34 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package com.mozilla.telemetry.ingestion.util;
+
+public class Attribute {
```
jklukas: I like where this is going in bringing all these names together in one class. I think we should go further in an upcoming PR and create a new module called something like `ingestion-java-common` to hold these constants and whatever logic needs to be shared between ingestion-sink and ingestion-beam.

PR #750, opened 2019-08-15
Fix ingestion-sink errors and performance

relud:


relud:
for the record, it looks like we handle well with the following pod resources in kubernetes:
```yaml
resources:
  limits:
    cpu: 800m
    memory: 2Gi
  requests:
    cpu: 400m
    memory: 1Gi
```

pom.xml
```diff
@@ -27,7 +27,7 @@
 
         <!-- Keep these dependency versions in sync with those pulled in by beam;
              check https://mvnrepository.com/artifact/org.apache.beam -->
-        <google-cloud.version>1.49.0</google-cloud.version>
+        <google-cloud.version>1.61.0</google-cloud.version>
```
jklukas: I'm surprised this doesn't throw all sorts of errors. I think the ingestion-beam tests are being skipped, but this change could have significant impact on ingestion-beam.

relud: additional info: beam 2.13.0 depends on google-cloud-core `1.61.0`, so this is picking up bug fixes for ingestion-sink, but it's also putting the dependency back in sync with beam.

relud: I've updated the circleci config to run tests if the root pom.xml changed and there's a child pom.xml in the working dir

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/io/BigQuery.java
```diff
@@ -35,16 +36,28 @@ private WriteErrors(List<BigQueryError> errors) {
...
+    private final int maxMessages;
+    private final long maxDelayMillis;
 
     @VisibleForTesting
     final ConcurrentMap<TableId, Batch> batches = new ConcurrentHashMap<>();
 
     public Write(com.google.cloud.bigquery.BigQuery bigquery) {
+      this(bigquery, 10_000_000, // HTTP request size limit: 10 MB
+          10_000, // Maximum rows per request: 10,000
+          100); // Default 0.1 second
```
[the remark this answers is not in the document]
relud: I've removed the redundant constructor

PR #752, opened 2019-08-16
Normalize paths in SchemaStore to support casing variants

### LINKED ISSUE #744: Loading to BQ fails for doctypes with underscores
[As mentioned in the GCP migration validation document](https://docs.google.com/a/mozilla.com/document/d/14UoZGAI2CgMQW4qeXIX5Bj9nq2kvfCUaONiNknXXxYE/edit?disco=AAAADV_ZM9w), we are seeing errors that fenix bookmarks_sync pings are being dropped with errors that show no appropriate schema exists in the schema store.

The table does exist, but our logic for the schema store assumes that pings have hyphenated names rather than underscores.

We may want to build validation into mozilla-pipeline-schemas at this point to prevent deviation from naming conventions. To solve the immediate problem, we could alias the existing schemas to new places, or we could update the gcp-ingestion code.

I am inclined to pursue an m-p-s-based solution.

cc @relud @whd

jklukas:
Closes #744

Also see https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/387 for discussion of longer-term solutions to avoid this complexity.

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/schemas/JSONSchemaStoreTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/schemas/JSONSchemaStoreTest.java
@@ -39,6 +39,8 @@ public class JSONSchemaStoreTest {
     JSONSchemaStore store = JSONSchemaStore.of(SCHEMAS_LOCATION, EMPTY_ALIASING_CONFIG_LOCATION);
     assertTrue(store.docTypeExists("telemetry", "main"));
     assertTrue(store.docTypeExists("telemetry", "update"));
+    assertTrue(store.docTypeExists("telemetry", "untrustedModules"));
+    assertTrue(store.docTypeExists("telemetry", "untrusted_modules"));
   }
 
   @Test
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/schemas/JSONSchemaStoreTest.java
```diff
@@ -39,6 +39,8 @@ public void testDocTypeExists() {
     JSONSchemaStore store = JSONSchemaStore.of(SCHEMAS_LOCATION, EMPTY_ALIASING_CONFIG_LOCATION);
     assertTrue(store.docTypeExists("telemetry", "main"));
     assertTrue(store.docTypeExists("telemetry", "update"));
+    assertTrue(store.docTypeExists("telemetry", "untrustedModules"));
```
whd: Shouldn't we test BigQuerySchemaStore? The new case we're resolving is with e.g. `bookmarks_sync` on BQ output. For test purposes they amount to the same thing I suppose (in the current implementation, but I have thoughts on that in the MPS PR).

jklukas: The BigQueryIntegrationTest does effectively test this behavior for the BigQuerySchemaStore, but I can understand pushback for more unit testing.

PR #754, opened 2019-08-16
Refactor java common

jklukas:
cc @relud 

I am starting a refactor here so that beam and sink can share code. I will plan to factor our the decoded BQ transformation as I build this PR.

### DIFF
--- a/docs/ingestion-beam/index.md
+++ b/docs/ingestion-beam/index.md
@@ -528,10 +528,10 @@ If you wish to just run a single test class or a single test case, try something
 
 ```bash
 # Run all tests in a single class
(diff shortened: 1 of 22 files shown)


jklukas:
Welp. Of course this gets complicated. I had forgotten that the SchemaStore interface uses beam libs under the hood to access files on local filesystem or GCS, so the common module would be pulling in all of the beam deps. At that point, modules aren't really giving us much in terms of isolating dependencies and it would be easier to pull sink and beam together into a single module. I'll keep thinking on this, but will switch to focusing on refactoring the beam sink to delay decompression.

PR #760, opened 2019-08-20
Bug: Cannot serialize PipelineOptions

jklukas:
Fixup for #759 where we accessed a PipelineOptions instance from
a function body that has to execute on workers, thus it tried to
serialize the options object and failed.

This change fixes the bug and adds an integration test case to make
sure we exercise pubsub error output.


jklukas:
Integration test is failing due to order of attributes in the output apparently being non-deterministic. Looking into it.

PR #762, opened 2019-08-21
Share code between ingestion-beam and ingestion-sink

### LINKED ISSUE #761: Share code between ingestion-beam and ingestion-sink
Summary
---

This repo is configured as a maven multi-module project, so we can create an `ingestion-core` module with shared code that may be used by both `ingestion-beam` and `ingestion-sink`, but it will break running maven in module directories instead of the git root.

Naming
---

I prefer naming this shared module `core` because it's 4 letters, which is nice because so are `edge`, `beam`, and `sink`.

Maven
---

Maven will not automatically resolve sibling dependencies when run from within a module, e.g. running `cd ingestion-sink; mvn compile` would fail because `ingestion-core` would not be available.

A suboptimal option stack overflow commonly recommends is to install `ingestion-core` to `~/.m2` where it may quickly become stale, e.g. `cd ingestion-core; mvn install; cd ../ingestion-sink; mvn compile`.

A slightly better option is run maven with a projects list from the git root, e.g. `mvn compile -am -pl ingestion-sink`, where `-am` causes maven automatically build and include sibling dependencies. Unfortunately this requires all module configs to be present, which causes issues when building the docker image for `ingestion-sink`, because we don't want to include unrelated files like `ingestion-beam/pom.xml`.

An option that adds some repetition of sibling dependencies, but doesn't require all module configs to be present, is to use maven profiles in the top-level `pom.xml` that specify a subset of modules. This makes running maven look like `mvn compile -P ingestion-sink` after adding profiles to the top-level `pom.xml` like this:

```diff
diff --git a/pom.xml b/pom.xml
index 1451aac..3c5ab56 100644
--- a/pom.xml
+++ b/pom.xml
@@ -9,11 +9,22 @@
     <version>0.1-SNAPSHOT</version>
     <packaging>pom</packaging>
 
-    <modules>
-        <module>ingestion-beam</module>
-        <module>ingestion-core</module>
-        <module>ingestion-sink</module>
-    </modules>
+    <profiles>
+        <profile>
+            <id>ingestion-sink</id>
+            <modules>
+                <module>ingestion-core</module>
+                <module>ingestion-sink</module>
+            </modules>
+        </profile>
+        <profile>
+            <id>ingestion-beam</id>
+            <modules>
+                <module>ingestion-core</module>
+                <module>ingestion-beam</module>
+            </modules>
+        </profile>
+    </profiles>

     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
```

relud:
fixes #761 

backwards incompatible with calling `mvn` in module directories, but `./bin/mvn` should still do the right thing, see #761 for why.


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/transform/PubsubMessageToTableRow.java
```diff
@@ -4,11 +4,10 @@
 
 package com.mozilla.telemetry.ingestion.transform;
 
-import static com.mozilla.telemetry.ingestion.util.Attribute.CLIENT_ID;
-
 import com.google.cloud.bigquery.TableId;
 import com.google.pubsub.v1.PubsubMessage;
 import com.mozilla.telemetry.ingestion.io.BigQuery.Write.TableRow;
+import com.mozilla.telemetry.ingestion.util.Attribute;
```
jklukas:
```suggestion
import com.mozilla.telemetry.ingestion.core.Attribute;
```

Now that we're sharing code between modules, I think we should try to keep the package naming in line with the directory naming, so everything in ingestion-core should be under subpackage `...intestion.core`. I don't feel strongly about whether we have an additional `util` subpackage beneath that or not. The `Attribute` class in particular feels important enough that it could live at the root of `core`.

relud: Sounds good, I'll also move sink classes to `ingestion.sink.*`, but leave ingestion-beam as-is because now (right before prod deploy) is not the time to introduce that change

PR #766, opened 2019-08-21
Fix exec:java

relud:


pom.xml
```diff
@@ -114,6 +117,7 @@
                     </execution>
                 </executions>
                 <configuration>
+                    <skip>true</skip>
```
jklukas:
```suggestion
                  <skip>true</skip>
```

relud: i think we should instead indent the next line

PR #769, opened 2019-08-21
Fix template build

jklukas:
@whd reported that the template builds were failing, and it looks like that's due to the script not being updated to run maven at the top level.

I take this opportunity to make `bin/mvn` usable within the template build script.

### DIFF
--- a/bin/mvn
+++ b/bin/mvn
@@ -2,7 +2,7 @@
 
 # Wrapper script for executing maven via docker, but interacting with
 # the local filesystem. Useful for local development without installing
-# Java and mvn.
+# Java and mvn and also used by Jenkins.
 
 set -eo pipefail
 
@@ -17,9 +17,14 @@ else
     PROFILE=()
 fi
 
+INTERACTIVE_FLAGS=""
+if [ -z "$NONINTERACTIVE" ]; then
+    INTERACTIVE_FLAGS="-it"
+fi
+
 # Run mvn with a non-root user id
 # https://docs.docker.com/samples/library/maven/#Running-as-non-root
-docker run -it --rm \
+docker run $INTERACTIVE_FLAGS --rm \
     -u $UID \
     -v "$(git rev-parse --show-toplevel)":/var/maven/project \
     -w /var/maven/project \
--- a/bin/mvn
+++ b/bin/mvn
@@ -17,10 +17,14 @@ else
     PROFILE=()
 fi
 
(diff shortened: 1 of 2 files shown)


relud:
> Seeing `the input device is not a TTY` failures now on Jenkins

it's probably the `-it`. we might be able to work around that by adding `< /dev/tty`

jklukas:
I'll find a way to avoid the `--it`

bin/mvn
```diff
@@ -17,10 +17,14 @@ else
...
-    -v ~/.m2:/root/.m2 \
-    -v "$(git rev-parse --show-toplevel)":/root/project \
-    -w /root/project \
+    -u $UID \
+    -v "$(git rev-parse --show-toplevel)":/var/maven/project \
+    -w /var/maven/project \
+    -v ~/.m2:/var/maven/.m2 \
+    -e MAVEN_CONFIG=/var/maven/.m2 \
     -e GOOGLE_APPLICATION_CREDENTIALS \
     maven \
```
relud:
```suggestion
     maven:3-jdk-8 \
```

just pulled maven:latest and it can't compile because it has jdk 11 now and it throws:

```
[WARNING] COMPILATION WARNING : 
[WARNING] bootstrap class path not set in conjunction with -source 8
[ERROR] COMPILATION ERROR : 
[ERROR] warnings found and -Werror specified
```

jklukas:
@whd This should avoid a tty when running `build-template` now.

whd:
I'm running this now. So far I see `Unable to find image 'maven:3-jdk-8' locally` and it's pulling down dependencies, so it looks like the build at least starts. I don't know what version was being used before but hopefully it doesn't matter.

bin/mvn
```diff
@@ -17,10 +17,19 @@ else
...
+# https://docs.docker.com/samples/library/maven/#Running-as-non-root
+docker run $INTERACTIVE_FLAGS --rm \
+    -u $UID \
+    -v "$(git rev-parse --show-toplevel)":/var/maven/project \
+    -w /var/maven/project \
+    -v ~/.m2:/var/maven/.m2 \
+    -e MAVEN_CONFIG=/var/maven/.m2 \
     -e GOOGLE_APPLICATION_CREDENTIALS \
-    maven \
+     maven:3-jdk-8 \
```
whd:
```suggestion
    maven:3-jdk-8 \
```

PR #771, opened 2019-08-21
Add more spotless rules

relud:
to automatically fix some checkstyle checks

@jklukas we discussed this earlier, but it turns out these configs in spotless are separate from the beam codestyle we are pulling in.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -5,7 +5,6 @@
 package com.mozilla.telemetry.decoder;
 
 import com.google.common.collect.ImmutableList;
-import com.mozilla.telemetry.decoder.ParseUri;
```
relud: all removed imports in this PR are due to being in the same package as the current file.

pom.xml
```diff
@@ -174,10 +174,23 @@
...
+                            <!-- end header with extra trailing newline to separate from package statement -->
+                            <file>${maven.multiModuleProjectDirectory}/checkstyle/java.header</file>
+                        </licenseHeader>
                         <eclipse>
                             <file>https://raw.githubusercontent.com/apache/beam/v${beam.version}/sdks/java/build-tools/src/main/resources/beam/beam-codestyle.xml</file>
                             <version>4.7.1</version>
                         </eclipse>
+                        <importOrder>
+                            <!-- default import order -->
+                            <order>#</order>
```
relud: this is actually a single-line config, and that line is a comment that gets ignored

ingestion-beam/src/test/java/com/mozilla/telemetry/avro/PubsubMessageRecordFormatterTest.java
```diff
@@ -9,17 +9,14 @@
...
 import org.apache.avro.AvroTypeException;
 import org.apache.avro.Schema;
 import org.apache.avro.SchemaBuilder;
 import org.apache.avro.generic.GenericArray;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
-
 import org.json.JSONArray;
 import org.json.JSONObject;
-
```
relud: r? @acmiyaguchi because this removes the package separation you used in the avro stuff, bringing it more in line with the rest of the code base

checkstyle/java.header
```diff
@@ -1,3 +1,4 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
relud: from pom.xml:

> end license header with extra trailing newline to separate from package statement

pom.xml
```diff
@@ -174,10 +174,23 @@
...
                 <configuration>
                     <java>
+                        <excludes>
+                            <!-- exclude beam patch from formatting -->
+                            <exclude>src/main/java/org/apache/beam/sdk/io/gcp/bigquery/WriteTables.java</exclude>
+                        </excludes>
+                        <licenseHeader>
+                            <!-- end license header with extra trailing newline to separate from package statement -->
+                            <file>${maven.multiModuleProjectDirectory}/checkstyle/java.header</file>
+                        </licenseHeader>
```
jklukas: Does this mean spotless is going to add our header for us automatically?!!!!!

cc @mreid-moz this is worth at least 10 points toward H2 IAP bonus.

ingestion-beam/src/main/java/com/mozilla/telemetry/avro/BinaryRecordFormatter.java
```diff
@@ -8,7 +8,6 @@
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.UncheckedIOException;
-
```
jklukas: I have gotten used to having this space between Java imports and others, so I was going to complain. But apparently this change is in line with Google Java Style. I'm generally in favor of sticking to whatever they recommend by default to avoid having to think too hard about it.

https://google.github.io/styleguide/javaguide.html#s3.3.3-import-ordering-and-spacing

PR #779, opened 2019-08-28
Remove clustering patch in favor of beam 2.15 clustering support

### LINKED ISSUE #704: Remove clustering patch
Once Beam 2.15 is released (approximately late September), we will have official support for BigQuery clustering and we should remove our custom patch and adapt to the official interface as part of the upgrade to 2.15.

jklukas:
Closes #704


jklukas:
> we should check that we are updating any beam-synchronized dependencies if needed

I spent a good chunk of time trying newer versions. It looks like we no longer have to worry about grpc, since beam now vendors that dependency. But we still do get conflicts on other packages. In particular, `com.google.apis:google-api-services-bigquery`, `com.google.apis:google-api-services-storage`, etc. are pulled in by both the beam dataflow lib and by the google-cloud client libraries. If these don't match, we have problems.

1.61 appears to still be the most recent version of google-cloud that doesn't confuse the dataflow dependencies.

jklukas:
Jackson is also up to date with what beam is using here.

PR #781, opened 2019-08-28
Bug 1568042 - Allow ingesting heka-framed blobs

wlach:
Work in progress. This adds heka blob parsing, some tests for that, and the very beginnings of a heka->pubsub sink. Still needed:

* Code to traverse heka blobs in a filesystem store (equivalent of Dataset from scala moztelemetry)
* Make the json -> pubsub integration bits actually work. It seems like the existing text->pubsub and json->pubsub classes assume a 1:1 mapping, which might not be what we want here. There's a bunch I still need to figure out about how to make this work...


wlach:
Gotcha, could you point me where we can suppress spotless checks?

relud:
> Gotcha, could you point me where we can suppress spotless checks?

[pom.xml#L189](https://github.com/mozilla/gcp-ingestion/blob/master/pom.xml#L189)

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -64,11 +68,33 @@ public FileInput(ValueProvider<String> fileSpec, InputFileFormat fileFormat) {
...
+
+    private final ValueProvider<String> fileSpec;
+
+    public HekaInput(ValueProvider<String> fileSpec) {
+      this.fileSpec = fileSpec;
+    }
+
+    @Override
+    public Result<PCollection<PubsubMessage>> expand(PBegin input) {
+      return input.apply(FileIO.match().filepattern(fileSpec)).apply(FileIO.readMatches()).apply(new HekaTransform())
```
jklukas: `.apply(new HekaTransform())` wraps its output in Result, so it looks like the type here would be incorrect for the flatmap. We need to pass just the successful output into the flatmap. I'm playing with a solution now.

jklukas:
```
      Result<PCollection<List<PubsubMessage>>> result = input
          .apply(FileIO.match().filepattern(fileSpec)).apply(FileIO.readMatches())
          .apply(new HekaTransform());
      return Result.of(result.output().apply(Flatten.iterables()), result.errors());
```

wlach: Makes perfect sense! It's difficult to keep track of all these types 😢

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaTransform.java
```diff
@@ -0,0 +1,33 @@
...
+import org.apache.beam.sdk.io.FileIO.ReadableFile;
+import org.apache.beam.sdk.io.fs.MatchResult;
+import org.json.JSONObject;
+import com.mozilla.telemetry.transforms.WithErrors;
+import com.mozilla.telemetry.transforms.MapElementsWithErrors;
+
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.ParDo;
+
+public class HekaTransform extends MapElementsWithErrors<ReadableFile, List<PubsubMessage>> {
```
jklukas: Beam provides some naming guidelines for transforms: https://beam.apache.org/contribute/ptransform-style-guide/#naming

In that sense, the best name for this would probably be `HekaIO.read()`, which would mean the outer `Heka` class doesn't extend anything, but would have a single static `read()` method that returns an instance of `HekaIO.Read()` which would be the thing extending MapElementsWithErrors.

I don't think it's worth worrying about giving this a pretty name right now. I'm mostly just trying to continue giving context on Beam as we go here.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -92,32 +118,33 @@ public BigQueryInput(ValueProvider<String> tableSpec, BigQueryReadMethod method,
...
-          break;
-        default:
-        case QUERY:
-          read = read.fromQuery(tableSpec).usingStandardSql();
+      case TABLE:
+        read = read.from(tableSpec);
+        break;
+      default:
+      case QUERY:
+        read = read.fromQuery(tableSpec).usingStandardSql();
```
jklukas: It looks like you have editor settings that are fighting with what exists here. In the end, you can call `mvn spotless:apply` and it'll auto-format everything back to what google java style wants anyway.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
```diff
@@ -0,0 +1,118 @@
...
+      // get message data
+      byte[] messageBuffer = new byte[header.getMessageLength()];
+      is.read(messageBuffer);
+
+      int uncompressedLength = Snappy.uncompressedLength(messageBuffer);
+      byte[] uncompressedMessage = new byte[uncompressedLength];
+      Snappy.uncompress(messageBuffer, 0, header.getMessageLength(), uncompressedMessage, 0);
+      Heka.Message message = Heka.Message.parseFrom(uncompressedMessage);
+
+      JSONObject payload = new JSONObject(new String(message.getPayload().getBytes()));
```
jklukas: We just in the last week decided to isolate usage of `org.json.JSONObject` as much as possible, so you're going to have some errors after rebasing. I didn't anticipate that conflicting with your work, so I apologize for that. I'm happy to hop in and do refactoring to have this use jackson `ObjectNode` instead (using our `Json.asObjectNode(message.getPayload())`).

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaTransform.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    protected List<PubsubMessage> processElement(ReadableFile readableFile) throws IOException {
```
jklukas: We will also need to define a `processError` override method, since `MapElementsWithErrors` doesn't have a default implementation of how to process a `ReadableFile` into an error ouput pubsubmessage. We might want to turn this into a PusubMessage with no payload, but make sure we add a field for the file path of the ReadableFile, so we can go back and look at that file to see what's problematic.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
jklukas: Looks like Beam can't automatically infer the type here, so this could become:

```
      Result<PCollection<List<PubsubMessage>>> result = input
          .apply(FileIO.match().filepattern(fileSpec)).apply(FileIO.readMatches())
          .apply(new HekaTransform());
      result.output().setCoder(ListCoder.of(PubsubMessageWithAttributesCoder.of()));
      return Result.of(result.output().apply(Flatten.iterables()), result.errors());
```

where we have an explicit `setCoder` call.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
wlach: No problem, I don't think this should be too difficult to deal with. I think I will try to get things working at least locally with the current structure, then do a rebase.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaTransform.java
wlach: can you go into more detail with what you mean by a "field for the file path"?

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
wlach: Thanks! That did it

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaTransform.java
jklukas: See `FailureMessage.java`. That has various methods for taking a caught exception and the associated record, and creating a PubsubMessage out of it. In this case, we'd want something like:

```
  public static PubsubMessage of(Object caller, ReadableFile readableFile, Throwable e) {
    Map<String, String> attributes = errorAttributes(caller, e);
    attributes.put("readable_file", readableFile.toString());
    return new PubsubMessage(null, errorAttributes(caller, e));
  }
```

I expect `readableFile.toString()` will give a full description of the object, including the path in GCS.

jklukas: Looks like it's failing on the `Heka.Header.parseFrom(headerBuffer)` call. `headerBuffer` is a zero-length byte array, and this raises `com.google.protobuf.InvalidProtocolBufferException: Message missing required fields: message_length`.

So that gets caught by the MapElementsWithErrors machinery, invoking `processError` and it looks like I steered you wrong there. You call `FailureMessage.of` which ends up creating a PubsubMessage with a null payload. I had forgotten that the null payload would cause encoding errors. So perhaps it would be best to make the payload `"{}".getBytes(StandardCharsets.UTF_8)`.

wlach: Yup, this was the problem, plus a small mistake in your suggested error message code (not actually using the `attributes` map).

wlach: Here's what an error payload looks like now: 

```
{"attributeMap":{"error_message":"com.google.protobuf.InvalidProtocolBufferException: Message missing required fields: message_length","exception_class":"com.google.protobuf.InvalidProtocolBufferException","stack_trace":"[com.google.protobuf.UninitializedMessageException.asInvalidProtocolBufferException(UninitializedMessageException.java:79), com.google.protobuf.AbstractParser.checkMessageInitialized(AbstractParser.java:68), com.google.protobuf.AbstractParser.parseFrom(AbstractParser.java:191), com.google.protobuf.AbstractParser.parseFrom(AbstractParser.java:203), com.google.protobuf.AbstractParser.parseFrom(AbstractParser.java:208), com.google.protobuf.AbstractParser.parseFrom(AbstractParser.java:48), com.mozilla.telemetry.heka.Heka$Header.parseFrom(Heka.java:575), com.mozilla.telemetry.heka.HekaReader.readHekaMessage(HekaReader.java:48), com.mozilla.telemetry.heka.HekaReader.readHekaStream(HekaReader.java:106), com.mozilla.telemetry.heka.HekaTransform.processElement(HekaTransform.java:23), com.mozilla.telemetry.heka.HekaTransform.processElement(HekaTransform.java:1), com.mozilla.telemetry.transforms.MapElementsWithErrors$DoFnWithErrors.processElementOrError(MapElementsWithErrors.java:89), com.mozilla.telemetry.transforms...","readable_file":"ReadableFile{metadata=Metadata{resourceId=/var/maven/project/ingestion-beam/tmp/cheezburger.bad.heka, sizeBytes=4, isReadSeekEfficient=true, lastModifiedMillis=0}, compression=UNCOMPRESSED}","error_type":"HekaTransform"},"payload":"e30="}
```

wlach:
@jklukas I think this is ready for another look. The ingestion-beam tests are now passing, the  ingestion-sink-integration tests aren't passing, but I don't think that has to do with this PR.

AFAICT there are 3 issues remaining, but I'm not sure if there's anything blocking merging. Might be easier to get this in and then followup, thoughts?

jklukas:
This needs a rebase on master before being ready to merge, which will require refactoring to use jackson ObjectNode rather than org.json JSONObject.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
```diff
@@ -0,0 +1,114 @@
...
+    // get message data
+    byte[] messageBuffer = new byte[header.getMessageLength()];
+    is.read(messageBuffer);
+
+    int uncompressedLength = Snappy.uncompressedLength(messageBuffer);
+    byte[] uncompressedMessage = new byte[uncompressedLength];
+    Snappy.uncompress(messageBuffer, 0, header.getMessageLength(), uncompressedMessage, 0);
+    Heka.Message message = Heka.Message.parseFrom(uncompressedMessage);
+
+    JSONObject payload = new JSONObject(new String(message.getPayload().getBytes()));
```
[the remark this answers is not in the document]
wlach: 😢  Do you know if this is the case in the production data we care about? It looks like we tested these cases in the python version by creating specially crafted messages:

https://github.com/mozilla/python_moztelemetry/blob/09ddf1ec7d953a4308dfdcb0ed968f27bd5921bb/tests/heka/test_message_parser.py#L152

I don't think it's worth doing any more than we absolutely need to for our backfilling use case.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
```diff
@@ -0,0 +1,114 @@
...
+    Object field = message.getField(fieldDescriptor);
+    if (field instanceof List) {
+      for (Object i : (List) field) {
+        Heka.Field f = (Heka.Field) i;
+        String key = (String) f.getName();
+        List<String> path = key.contains(".") ? Arrays.asList(key.split("\\."))
+            : Arrays.asList("meta", key);
+        if (f.getValueType() == Heka.Field.ValueType.STRING) {
+          String value = f.getValueString(0);
+          if (value.charAt(0) == '{') {
```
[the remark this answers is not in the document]
wlach: Hmm, yes, this might need a bit more refinement. It looks like the scala version attempts to parse the value as json, falling back to just setting the value to a string if that fails:

https://github.com/mozilla/moztelemetry/blob/63e2e57d098f923e5e005e478d447b8f481b64bc/src/main/scala/com/mozilla/telemetry/heka/package.scala#L116

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
```diff
@@ -0,0 +1,145 @@
...
+      Optional.ofNullable(meta.path("appName").textValue())
+          .ifPresent(s -> attributes.put(Attribute.APP_NAME, s));
+      Optional.ofNullable(meta.path("appUpdateChannel").textValue())
+          .ifPresent(s -> attributes.put(Attribute.APP_UPDATE_CHANNEL, s));
+      Optional.ofNullable(meta.path("appVersion").textValue())
+          .ifPresent(s -> attributes.put(Attribute.APP_VERSION, s));
+      Optional.ofNullable(meta.path("documentId").textValue())
+          .ifPresent(s -> attributes.put(Attribute.DOCUMENT_ID, s));
+      Optional.ofNullable(meta.path("sourceVersion").textValue())
+          .ifPresent(s -> attributes.put(Attribute.DOCUMENT_VERSION, s));
```
jklukas: @mreid-moz Is sourceVersion the doc version?

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
```diff
@@ -0,0 +1,145 @@
...
+      Optional.ofNullable(meta.path("geoCity").textValue())
+          .ifPresent(s -> attributes.put(Attribute.GEO_CITY, s));
+      Optional.ofNullable(meta.path("geoCountry").textValue())
+          .ifPresent(s -> attributes.put(Attribute.GEO_COUNTRY, s));
+      Optional.ofNullable(meta.path("geoSubdivision1").textValue())
+          .ifPresent(s -> attributes.put(Attribute.GEO_SUBDIVISION1, s));
+      Optional.ofNullable(meta.path("geoSubdivision2").textValue())
+          .ifPresent(s -> attributes.put(Attribute.GEO_SUBDIVISION2, s));
+      // TODO: Do heka messages contain parsed user agent info? DNT? Method? Protocol?
```
jklukas: @mreid-moz Do heka messages contain parsed user agent info? Other headers?

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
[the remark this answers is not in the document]
mreid-moz: data being in the `submission` field is definitely a real thing in the production data. I don't think we need to handle the `content` field, IIRC that is only a thing in data in Landfill.

+1 for doing exactly what's needed.

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
mreid-moz: [Yes](https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L223)

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
mreid-moz: I don't think we parse/store user agent info in the telemetry data on AWS.

cc @trink can you confirm?

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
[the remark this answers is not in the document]
wlach: Attempted to address this here: https://github.com/mozilla/gcp-ingestion/pull/781/commits/78b9e27825785bbf8ae080f233c815c6e1f01b4b

ingestion-beam/src/main/java/com/mozilla/telemetry/heka/HekaReader.java
whd: We do not store/parse user agent info for telemetry data in AWS (we do for structured). We do store these headers: https://github.com/mozilla-services/lua_sandbox_extensions/blob/master/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L321-L324.

ingestion-beam/src/test/resources/testdata/heka/crashping.heka.json
```diff
@@ -10,7 +10,7 @@
         "version": "69.0",
         "xpcomAbi": "x86-msvc"
     },
-    "clientId": "a136763f-3b9a-492d-8964-bfe783b68dc6",
+    "clientId": "0e8e41d2-d46f-4818-b071-c6605257e6aa",
```
jklukas: This test was failing before the commit. I'm assuming the json is just not up to date with the heka data for the test, but correct me if I'm wrong on that, @wlach

wlach:
Let's merge this! It seems to work reasonable well for the data we've thrown at it so far. Can always address any issues in followups.

PR #788, opened 2019-08-30
Ensure we use Jackson for serializing to JSON

### LINKED ISSUE #780: Non-ASCII text being redacted in the pipeline
@tdsmith found that description values for Mozilla Online addons in BQ often consist of a lot of `?` characters, presumably non-ASCII text that's getting chewed up:
```
SELECT
  addons.value.description
FROM
  `moz-fx-data-shared-prod.telemetry_stable.event_v4`,
  UNNEST(environment.addons.active_addons) AS addons
WHERE
  DATE(submission_timestamp) = "2019-08-22"
LIMIT
  10
```

Row | description |  
-- | -- | --
1 | Erstellen Sie Bildschirmfotos von vollst?ndigen Webseiten oder Ausschnitten und speichern Sie diese |  
2 | Upozorn? na nov? e-maily, p?elo?? slova, p?evede m?ny a pejsek Krasty d? v?d?t, kdy? zjist? n?co nav |  
3 | Roz???en? pro pr?ci s elektronick?mi formul??i Software602 a.s. |  
4 | Po?izujte sn?mky webov?ch str?nek a ukl?dejte je do?asn? nebo natrvalo. |  
5 | Nejrychlej?? cesta na Seznam. |  
6 | ??????? ??????? ? ????????? ?? ????????? ? ?????????? ?? ???????? ??? ????????. |  
7 | McAfee? WebAdvisor |  
8 | Tw?rz wycinki i?zrzuty stron internetowych i?zapisuj je tymczasowo lub trwale. |  
9 | A very lightweight add-on that allows you to watch YouTube? videos using Flash? Player. |  
10 | Effectuez des captures d’?cran sur le Web et sauvegardez-les de mani?re temporaire ou permanente.

jklukas:
Closes #780

I wasn't able to reproduce encoding problems on my Mac using the Java local runner, but I _was_ able to reproduce issues by submitting a job to Dataflow. The change in this PR led to correct output without corrupted characters replaced by `?`.

I don't see a way to write a regression test for this, since I still don't understand what exactly about the Dataflow execution environment differs from local testing.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -140,7 +140,11 @@ public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<Pubs
 
     addAttributesFromPayload(attributes, json);
 
-    byte[] normalizedPayload = json.toString().getBytes();
+    // https://github.com/mozilla/gcp-ingestion/issues/780
+    // We need to be careful to consistently use our util methods (which use Jackson) for
+    // serializing and deserializing JSON to reduce the possibility of introducing encoding
+    // issues. We previously called json.toString().getBytes() here without specifying a charset.
+    byte[] normalizedPayload = Json.asBytes(json);
 
     PerDocTypeCounter.inc(attributes, "valid_submission");
     PerDocTypeCounter.inc(attributes, "valid_submission_bytes", submissionBytes);


mreid-moz:
Can / should we add a check to warn on use of `JSONObject$toString`?

jklukas:
> Can / should we add a check to warn on use of JSONObject$toString?

Probably yes.  I don't know how to do so, but suspect there are ways. Will file an issue to follow up.

relud:
regression test: we should be able to run an integration test using actual GCP resources *and dataflow* to verify that we aren't reproducing the original problem.

jklukas:
> regression test: we should be able to run an integration test using actual GCP resources _and dataflow_ to verify that we aren't reproducing the original problem.

I attempted last year to write an automated test that submitted a Dataflow job. It seemed prohibitively slow; we have no guarantee on how long it takes for a submitted pipeline to start, and empirically it can be well over five minutes. I'm reticent to slow down our CI to allow for this.

relud:
> empirically it can be well over five minutes

😭

> I'm reticent to slow down our CI to allow for this.

I would say this is the kind of test that should only run on master, because it shouldn't block anything so much as notify us that we have regressed.

jklukas:
> I would say this is the kind of test that should only run on master, or on a nightly basis, because it shouldn't block anything so much as notify us that we have regressed.

That doesn't sound too terrible. So we wouldn't let it block deploys of docker images?

relud:
> So we wouldn't let it block deploys of docker images?

correct

PR #801, opened 2019-09-03
Support multiple topics in TestWithPubsubResources

relud:
blocks to-be-PR'd `Pubsub.Write` implementation

### DIFF
--- a/ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
+++ b/ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
@@ -8,26 +8,26 @@ import static org.junit.Assert.assertEquals;
 
 import com.google.protobuf.ByteString;
 import com.google.pubsub.v1.PubsubMessage;
-import com.mozilla.telemetry.ingestion.sink.util.TestWithPubsubResources;
+import com.mozilla.telemetry.ingestion.sink.util.TestWithSinglePubsubTopic;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import org.junit.Test;
 
-public class PubsubReadIntegrationTest extends TestWithPubsubResources {
+public class PubsubReadIntegrationTest extends TestWithSinglePubsubTopic {
+
+  private static final PubsubMessage TEST_MESSAGE = PubsubMessage.newBuilder()
+      .setData(ByteString.copyFrom("test".getBytes())).build();
 
   @Test
-  public void canReadOneMessage() throws Exception {
-    publisher
-        .publish(PubsubMessage.newBuilder().setData(ByteString.copyFrom("test".getBytes())).build())
-        .get(10, TimeUnit.SECONDS);
+  public void canReadOneMessage() {
+    publish(TEST_MESSAGE);
 
     AtomicReference<PubsubMessage> received = new AtomicReference<>();
     AtomicReference<Pubsub.Read> input = new AtomicReference<>();
 
-    input.set(new Pubsub.Read(subscriptionName.toString(),
+    input.set(new Pubsub.Read(getSubscription(),
         // handler
         message -> CompletableFuture.supplyAsync(() -> message) // create a future with message
             .thenAccept(received::set) // add message to received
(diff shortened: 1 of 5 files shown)


jklukas:
What's our use case for `Pubsub.Write`?

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
```diff
@@ -12,22 +12,22 @@
...
-        .publish(PubsubMessage.newBuilder().setData(ByteString.copyFrom("test".getBytes())).build())
-        .get(10, TimeUnit.SECONDS);
+  public void canReadOneMessage() {
+    publish(0, TEST_MESSAGE);
 
     AtomicReference<PubsubMessage> received = new AtomicReference<>();
     AtomicReference<Pubsub.Read> input = new AtomicReference<>();
 
-    input.set(new Pubsub.Read(subscriptionName.toString(),
+    input.set(new Pubsub.Read(subscriptions.get(0).getName(),
```
jklukas: This may be too much magic. It took me a while to realize `subscriptions` was defined in `TestWithPubsubResources`. This sort of thing is normal in Scala, but feels non-idiomatic in Java.

Could we perhaps rename this method to `generatedSubscriptions` or `testSubscriptions`? That might make it a little more obvious that they're coming from the base class? Or maybe we should access it as `TestWithPubsubResources.this.subscriptions`?

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
```diff
@@ -12,22 +12,22 @@
...
+  private static final PubsubMessage TEST_MESSAGE = PubsubMessage.newBuilder()
+      .setData(ByteString.copyFrom("test".getBytes())).build();
+
   @Test
-  public void canReadOneMessage() throws Exception {
-    publisher
-        .publish(PubsubMessage.newBuilder().setData(ByteString.copyFrom("test".getBytes())).build())
-        .get(10, TimeUnit.SECONDS);
+  public void canReadOneMessage() {
+    publish(0, TEST_MESSAGE);
```
jklukas: Should we have a convenience method `publish(TEST_MESSAGE)` that assumes topic 0 for the simple case where the test only uses one topic?

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/util/TestWithPubsubResources.java
```diff
@@ -14,32 +15,45 @@
...
 
-  public String projectId;
-  public String topicId;
-  public String subscriptionId;
-  public ProjectTopicName topicName;
-  public ProjectSubscriptionName subscriptionName;
-  public TopicAdminClient topicAdminClient;
-  public SubscriptionAdminClient subscriptionAdminClient;
+  private String projectId;
+  protected int numTopics = 1;
```
jklukas: Maybe we should make `numTopics` abstract so that clients are forced to explicitly set it. That would perhaps make it easier to understand why we're indexing methods with `0` in various places.

jklukas: Or maybe we should have a `TestWithSinglePubsubTopic` subclass that sets numTopics = 1 and provides simpler methods that avoid having to pass a `0` index.

relud: I like this option, will do

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
relud: how about just `getSubscriptions()`?

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
relud: added for `TestWithSinglePubsubTopic`

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
jklukas: Yes, I like `getSubscriptions` much better.

relud:
> What's our use case for `Pubsub.Write`

I'm going to use it in my initial implementation of BigQuery file loads

PR #803, opened 2019-09-03
Implement Pubsub.Write

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/Sink.java
```diff
@@ -53,30 +55,35 @@ public static void main(String[] args) {
...
-      output = new BigQuery.Write(BigQueryOptions.getDefaultInstance().getService(),
-          // BigQuery.Write.Batch.getByteSize reports protobuf size, which can be ~1/3rd more
-          // efficient than the JSON that actually gets sent over HTTP, so we use to 60% of the 10MB
-          // API limit by default.
-          Env.getLong(BATCH_MAX_BYTES, 6_000_000L), // default 6MB
-          // BigQuery Streaming API Limits maximum rows per request to 10,000
-          Env.getInt(BATCH_MAX_MESSAGES, 10_000), // default 10K messages
-          Env.getDuration(BATCH_MAX_DELAY, "1s"), // default 1 second
-          Env.getString(OUTPUT_TABLE), format);
+      if (Env.optString(OUTPUT_TOPIC).isPresent()) {
```
jklukas: I think this should be an else/if rather than nested. 

But perhaps we should really have a bit of a refactor here. We should validate that only one of (OUTPUT_BUCKET, OUTPUT_TOPIC, OUTPUT_TABLE) is specified in the environment. Perhaps this warrants a new method in `Env` and factoring out a `buildOutput` method:

```
// Throws IllegalArgumentException if zero or more than one of these variables is set.
String outputVarName = Env.getMutuallyExclusiveOption(ImmutableList.of(OUTPUT_BUCKET, OUTPUT_TOPIC, OUTPUT_TABLE));

output = buildOutput(outputVarName);

...

public Function<PubsubMessage, CompletableFuture<?>> buildOutput(String outputVarName) {

switch (outputVarName) {
  case OUTPUT_BUCKET:
    return ...;
  ...
}
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/Sink.java
```diff
@@ -53,30 +55,35 @@ public static void main(String[] args) {
           Env.getLong(BATCH_MAX_BYTES, 100_000_000L), // default 100MB
           Env.getInt(BATCH_MAX_MESSAGES, 1_000_000), // default 1M messages
           Env.getDuration(BATCH_MAX_DELAY, "10m"), // default 10 minutes
-          gcsPrefix, format);
+          gcsPrefix, format)::apply;
```
jklukas: I don't understand why we're now calling `apply` here.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToTemplatedString.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Transform a {@link PubsubMessage} into a {@link String} using {@link StringSubstitutor}.
+ *
+ * <p>All message attributes, plus derived attributes provided by {@link DerivedAttributesMap},
+ * are available for substitution.
+ *
+ * @throws IllegalArgumentException if template is not fully resolved for a given message.
+ */
+public class PubsubMessageToTemplatedString implements Function<PubsubMessage, String> {
```
jklukas: This is a nice improvement 👍

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/Sink.java
relud: because `Function<PubsubMessage, CompletableFuture<?>>` is incompatible with `Function<PubsubMessage, CompletableFuture<Void>>` and `Function<PubsubMessage, CompletableFuture<String>>`, due to the nesting levels, but I just realized that I can wrap `Pubsub.Write` so that it is `Function<PubsubMessage, CompletableFuture<Void>>`.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/Sink.java
relud: I'll file a follow up PR to refactor this

PR #804, opened 2019-09-03
Use unique codecov flags per CI job

relud:


relud:
> What's the intention of separating them out?

codecov is failing to merge reports that share the same flag, which i thought was because it was deduplicating reports by flag, but now i see the docs say it should merge multiple uploads:

https://docs.codecov.io/docs/merging-reports
> Codecov does not override report data for multiple uploads. We always merge the data.

closing invalid.

PR #805, opened 2019-09-03
Make junit test dependency top-level

relud:


relud:
> We probably want to remove the dep from ingestion-beam too.

We don't explicitly depend on it in ingestion-beam >_<

PR #813, opened 2019-09-04
Move ingestion-sink config to dedicated class

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -0,0 +1,126 @@
...
+    return PubsubMessageToJSONObject.Format.valueOf(Env.getString(OUTPUT_FORMAT, "raw"));
+  }
+
+  public static Pubsub.Read getInput() {
+    OutputType outputType = OutputType.get();
+    // read pubsub messages from INPUT_SUBSCRIPTION
+    return new Pubsub.Read(Env.getString(INPUT_SUBSCRIPTION), outputType.getOutput(),
+        builder -> builder.setFlowControlSettings(FlowControlSettings.newBuilder()
+            .setMaxOutstandingElementCount(Env.getLong(MAX_OUTSTANDING_ELEMENT_COUNT,
+                outputType.defaultMaxOutstandingElementCount()))
```
jklukas: I think we should have a public final method to do this:

```
public final long getMaxOutstandingElementCount() {
  return Env.getLong(MAX_OUTSTANDING_ELEMENT_COUNT, getDefaultMaxOutstandingElementCount())
}
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -0,0 +1,126 @@
...
+            // 10MB API limit by default.
+            Env.getLong(BATCH_MAX_BYTES, 6_000_000L), // default 6MB
+            // BigQuery Streaming API Limits maximum rows per request to 10,000
+            Env.getInt(BATCH_MAX_MESSAGES, 10_000), // default 10K messages
+            Env.getDuration(BATCH_MAX_DELAY, "1s"), // default 1 second
+            Env.getString(OUTPUT_TABLE), getFormat());
+      }
+    };
+
+    long defaultMaxOutstandingElementCount() {
```
jklukas: I think these should have a `get` prefix and be documented as intended to be overridable.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -0,0 +1,126 @@
...
+
+    long defaultMaxOutstandingRequestBytes() {
+      return 100_000_000L; // 100MB
+    }
+
+    // Each case in the enum must implement this method to define how to write out messages.
+    abstract Function<PubsubMessage, CompletableFuture<Void>> getOutput();
+
+    static OutputType get() {
+      if (Env.optString(OUTPUT_BUCKET).isPresent()) {
```
jklukas: I still think we should check that only one of these is set. Perhaps:

```
List<OutputType> outputTypes = new ArrayList<>();
if (Env.optString(OUTPUT_BUCKET).isPresent()) {
  outputTypes.put(OutputType.gcs);
}
if (...) {

}
if (outputTypes.size() == 1) { return outputTypes.get(0) }
else { throw ... }
```

relud: I decided to generalize this and make it check that all environment variables that can be used for config must be used if they are set in the environment.

PR #815, opened 2019-09-04
Setup Gcs output for BigQuery file loads

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -50,15 +52,29 @@
       @Override
       Function<PubsubMessage, CompletableFuture<Void>> getOutput(Env env) {
         String gcsPrefix = env.getString(OUTPUT_BUCKET);
-        // Append - to output prefix if necessary to separate each blob's UUID from other parts
-        if (gcsPrefix.contains("/") && !gcsPrefix.endsWith("/")) {
-          gcsPrefix = gcsPrefix + "-";
+        // Append / to GCS prefix to enforce that it will be a directory
+        if (!gcsPrefix.endsWith("/")) {
+          gcsPrefix += "/";
```
relud: because if we append `OUTPUT_TABLE` it should be after a directory

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
```diff
@@ -71,9 +71,9 @@ public BatchWrite(long maxBytes, int maxMessages, Duration maxDelay, String batc
 
     public Batch() {
       // wait for init then setup full indicator by timeout
-      full = init.thenRunAsync(this::timeout);
+      full = init.thenRunAsync(this::timeout).exceptionally(ignore -> null);
       // wait for full then close
-      result = full.handleAsync(this::close);
+      result = full.thenComposeAsync(this::close);
```
relud: `this::close` can't ignore exceptions on `full` *and* return a `CompletableFuture`, so `full` got an `exceptionally` clause to handle ignoring exceptions.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -50,15 +52,29 @@
...
+        // Append OUTPUT_TABLE to GCS prefix if present
+        if (env.containsKey(OUTPUT_TABLE)) {
+          gcsPrefix += "output_table=" + env.getString(OUTPUT_TABLE) + "/";
+        }
+
+        // if OUTPUT_TOPIC is present send notification for each gcs blob written
+        final Function<BlobInfo, CompletableFuture<Void>> batchCloseHook;
+        if (env.containsKey(OUTPUT_TOPIC)) {
+          Function<PubsubMessage, CompletableFuture<Void>> pubsubWrite = pubsub.getOutput(env);
+          batchCloseHook = blobInfo -> pubsubWrite.apply(BlobInfoToPubsubMessage.apply(blobInfo));
```
jklukas: I am rather worried about the integrity of this hook. It seems inevitable that there will be situations in which we write an object, but then fail to write to Pub/Sub.

Have we evaluated using GCS [https://cloud.google.com/storage/docs/pubsub-notifications](PubSub notifications)? It seems like Google should be able to implement this with greater integrity than we can in our application code.

relud: > It seems inevitable that there will be situations in which we write an object, but then fail to write to Pub/Sub.

true, but then we can simply fail the batch and retry, and we would only load to bigquery if the pubsub notification succeeded.

> Have we evaluated using GCS [PubSub notifications](https://cloud.google.com/storage/docs/pubsub-notifications)?

I talked to @whd about this before. IIRC we liked having one less thing to deploy, and I like the idea that this could send notifications to a pubsub emulator or non-pubsub system, e.g. for testing, which gcs won't do.

I'm hesitant to trust Google to have a service set up for at least once delivery vs just failing after a backoff period. Preserving that guarantee is IMO a minimum req of this system.

jklukas: Thanks for the thoughts. Consider me convinced. This feels a bit like we're double-publishing, but really the GCS write is non-binding. It's the write to pubsub that makes an object available for loading into BQ, so we're good as long as we make sure we only ack after the pubsub message is delivered.

jklukas:
Will we have a cleanup policy in place for the bucket we're writing into? I want to make sure that if we're writing objects to GCS and then potentially losing track of them if the pubsub message never gets sent, that we'll eventually clean up the files on GCS.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Gcs.java
```diff
@@ -43,58 +48,60 @@ public Ndjson(Storage storage, long maxBytes, int maxMessages, Duration maxDelay
...
+    private static final Pattern blobIdPattern = Pattern
+        .compile("(gs://)?(?<" + BUCKET + ">[^/]+)(/(?<" + NAME + ">.*))?");
+
     @Override
     protected Batch getBatch(String gcsPrefix) {
-      return new Batch(storage, gcsPrefix);
+      final Matcher gcsPrefixMatcher = blobIdPattern.matcher(gcsPrefix);
+      if (!gcsPrefixMatcher.matches()) {
+        throw new IllegalArgumentException("Gcs prefix must match \"" + blobIdPattern.pattern()
+            + "\" but got \"" + gcsPrefix + "\" from: " + batchKeyTemplate.template);
```
jklukas: This might be easier to read with `String.format`

```java
         throw new IllegalArgumentException(String.format("Gcs prefix must match \"%s\" but got \"%s\" from: %s",
           blobIdPattern.pattern(), gcsPrefix, batchKeyTemplate.template));
```

relud:
> Will we have a cleanup policy in place for the bucket we're writing into?

yes we will need one, ideally set at 7 days, since that's how long the pubsub notifications can stick around.

Issue #811, opened 2019-09-04
Remove use of JSONObject#toString from ingestion-sink


relud:
the gcs output uses this method, and needs to be switched to jackson.

@jklukas how do you feel about using jackson's ObjectNode instead of org.json's JSONObject, and converting to JSONObject on an as-needed basis?

jklukas:
I like this a lot. Jackson certainly seems like a more highly-used project than org.json and we're only using org.json because that's what the schema validator requires. We'll have to refactor some code, but should be reasonably straight-forward.

We can encapsulate all of our use of org.json to a single class and do some mvn setup to forbid importing org.json elsewhere.

PR #816, opened 2019-09-05
Confine use of JSONObject

### LINKED ISSUE #789: Warn on use of JSONObject#toString
As a follow-up to https://github.com/mozilla/gcp-ingestion/pull/788, we should make sure we don't accidentally use JSONObject#toString for serializing JSON in the future. There are likely ways to enforce such things in a Java project, but I don't know specifically how to do so.

jklukas:
Closes #789 

I am leaving the change to ingestion-sink (#811) as a separate effort.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -46,10 +50,10 @@ public static boolean shouldScrub(Map<String, String> attributes, JSONObject jso
...
                 && attributes.get(Attribute.APP_VERSION).startsWith("68")))
         && Optional.of(json) // payload.metadata.RemoteType
-            .map(j -> j.optJSONObject("payload")) //
-            .map(j -> j.optJSONObject("metadata")) //
-            .map(j -> j.optString("RemoteType")) //
-            .filter(s -> s.startsWith("webIsolated=")).isPresent()) {
+            .map(j -> j.get("payload")) //
+            .map(j -> j.get("metadata")) //
+            .map(j -> j.get("RemoteType")) //
+            .map(JsonNode::textValue).filter(s -> s.startsWith("webIsolated=")).isPresent()) {
```
relud: `JsonNode::get` throws a runtime error if the key doesn't exist, and `JsonNode::textValue` will return `null` if the node isn't a string.
I think we should use `JsonNode::path` and `Optional::ofNullable` instead:
```java
Optional.ofNullable(json.path("payload").path("metadata").path("RemoteType").textValue())
    .filter(s -> s.startsWith("webIsolated=")).isPresent()
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -58,20 +62,14 @@ public static boolean shouldScrub(Map<String, String> attributes, JSONObject jso
...
-                }
-              }
-              return isPresent;
-            }) //
-            .filter(b -> b).isPresent()) {
+            .map(j -> j.get("payload")) //
+            .map(j -> j.get("hangs")) //
+            .filter(JsonNode::isArray).map(ArrayNode.class::cast) //
+            .map(ArrayNode::elements) //
+            .map(Streams::stream).orElseGet(Stream::empty) //
```
relud: `Streams.stream(json.path("payload").path("hangs").elements())` should work even if `payload.hangs` isn't an array, or isn't present

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -58,20 +62,14 @@ public static boolean shouldScrub(Map<String, String> attributes, JSONObject jso
...
+            .map(j -> j.get("remoteType")) //
+            .map(JsonNode::textValue) //
+            .anyMatch(s -> s.startsWith("webIsolated="))) {
```
relud: calling `JsonNode::textValue` can return null, so this should probably be wrapped in an `Optional`, e.g.
```java
            .anyMatch(j -> Optional.ofNullable(j.path("remoteType").textValue())
                .map(s -> s.startsWith("webIsolated=").isPresent()) //
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
jklukas: I was under the impression that `JsonNode::get` would return null in the case of a missing element according to docs for that method.

But I did not know about `path` and indeed that looks easier for chained calls.

I was relying on `JsonNode::textValue` to return null in the case it's not a string, which will return `Optional.empty` for a `map` call on an `Optional` instance.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
jklukas: `.map(JsonNode::textValue)` will return `Optional.empty` rather than `Optional.of(null)` if the cast that textValue returns null, so it's not necessary to make another explicit `Optional.ofNullable` call.

jklukas:
@relud Do you feel like your PR is in a better state than this one? I have not read through yours yet and I'll hold off on making any additional changes here until I hear word from you about which one you think is a better jumping-off point.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
relud: mind=blown

relud: but it isn't an optional at that point, it's a stream.

relud:
> @relud Do you feel like your PR is in a better state than this one? I have not read through yours yet and I'll hold off on making any additional changes here until I hear word from you about which one you think is a better jumping-off point.

let's stick with this one, and i'll rebase out the uses of JSONObject and make it focus on handling issues with sorting json, since that's a semi-unrelated issue that this also revealed.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
jklukas: Darn. You're totally right. Maybe humans shouldn't be allowed to program.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/util/Json.java
```diff
@@ -39,7 +39,7 @@ public static ObjectNode asObjectNode(Map<String, ?> map) {
    * Use {@code MAPPER} to convert {@link ObjectNode} to {@code Map<String, Object>}.
    */
   public static Map<String, Object> asMap(ObjectNode objectNode) {
-    return MAPPER.convertValue(objectNode, new TypeReference<Object>() {
+    return MAPPER.convertValue(objectNode, new TypeReference<Map<String, Object>>() {
```
jklukas: @relud I missed this in the last review, but `TypeReference<Object>()` looks like a mistake here. Does this change make sense to you?

jklukas:
Ready for another review.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/util/Json.java
relud: lgtm

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -140,6 +165,11 @@ public static PubsubMessage readPubsubMessage(String data) throws IOException {
     return output;
   }
 
+  /** Convert a tree to a map of strings to strings, as appropriate for attributes. */
+  public static <T> T convertValue(ObjectNode root, Class<T> klass) {
+    return MAPPER.convertValue(root, klass);
```
relud: `MAPPER::convertValue` doesn't take advantage of the type of `root`, but `MAPPER::treeToValue` does:
```suggestion
    return MAPPER.treeToValue(root, klass);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -105,24 +106,48 @@ public static Schema readBigQuerySchema(byte[] data) throws IOException {
...
+    if (root == null || !root.isObject()) {
+      throw new IOException("json value is not an object");
+    }
+    return (ObjectNode) root;
+  }
+
+  /**
+   * Return a new, empty {@link ObjectNode}.
+   */
+  public static ObjectNode emptyTree() {
```
relud: I would rather these be named `createObjectNode` and `createArrayNode`, like the `MAPPER` methods they wrap
```suggestion
  public static ObjectNode createObjectNode() {
```

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -105,24 +106,48 @@ public static Schema readBigQuerySchema(byte[] data) throws IOException {
...
+    return MAPPER.createObjectNode();
+  }
+
+  /**
+   * Return a new, empty {@link ArrayNode}.
+   */
+  public static ArrayNode emptyArray() {
```
relud:
```suggestion
  public static ArrayNode createArrayNode() {
```

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/AddMetadataTest.java
```diff
@@ -70,142 +77,92 @@ public void testOutput() {
...
-    Map<String, String> expected = ImmutableMap.<String, String>builder().put("app_name", "Firefox")
-        .put("sample_id", "18").put("geo_country", "CA").put("x_debug_id", "mysession")
-        .put("normalized_channel", "release").build();
-    Map<String, String> attributes = new HashMap<>();
-    AddMetadata.stripPayloadMetadataToAttributes(attributes, payload);
-    assertEquals(expected, attributes);
-    assertEquals(ImmutableMap.of("field1", 99), payload.toMap());
-  }
+  /*
+   * @Test public void testUserAgentFromAttributes() { Map<String, String> attributes =
```
relud: the commented out tests should get fixed before we merge

PR #819, opened 2019-09-06
Replace JSONObject with ObjectNode in ingestion-beam

relud:
builds on #818 

todo:
- [ ] find and fix `com.fasterxml.jackson.databind.exc.MismatchedInputException` in ParsePayload
- [ ] fix PubsubMessageRecordFormatterTest to correctly have null values

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -114,6 +114,13 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>${project.groupId}</groupId>
+            <artifactId>ingestion-core</artifactId>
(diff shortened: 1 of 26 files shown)


relud:
@jklukas i got carried away when implementing #818 and applied it everywhere. I split it up into separate PRs, just in case it's not too late to avoid duplicating the work. otherwise this can be closed invalid.

relud:
this also addresses the issue that `sortJSON` was doing no such thing, because the sorting option only applies to `java.util.Map`, not `JSONObject` >_<

relud:
ah, i see that you already filed an implementation

PR #823, opened 2019-09-11
Add DeduplicateByDocumentId transform

### LINKED ISSUE #821: Add option to deduplicate based on document_id
For the backfill of heka data into BQ, we want records to reside in `_stable` tables, meaning they must be fully deduplicated on document_id within each date. We could load into live tables and then run deduplication queries, but managing those queries could be difficult and it adds expense.

It may be feasible to handle the deduplication as part of the batch-mode Dataflow jobs we use to run the backfill. This would introduce a GroupByKey operation, slowing down and complicating the process, but will likely be less headache than handling it in a separate step.

Basic beam code would look like:

```
        .apply(MapElements
            .into(TypeDescriptors.kvs(TypeDescriptors.strings(), TypeDescriptor.of(PubsubMessage.class)))
            .via(message -> KV.of(message.getAttribute("document_id"), message)))
        .apply(GroupByKey.create())
        .apply(MapElements.into(TypeDescriptor.of(PubsubMessage.class))
            .via(kv -> {
              return StreamSupport.stream(kv.getValue().spliterator(), false)
                  .min((m1, m2) ->
                      m1.getAttribute(Attribute.SUBMISSION_TIMESTAMP) - m2.getAttribute(Attribute.SUBMISSION_TIMESTAMP)).get();
            }))

```

jklukas:
Closes #821

cc @whd 

This would be enabled for heka backfills so that we could load directly into stable tables rather than live tables. I don't know how this is going to perform in practice.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
@@ -13,10 +13,12 @@ import com.mozilla.telemetry.decoder.ParseProxy;
 import com.mozilla.telemetry.decoder.ParseUri;
 import com.mozilla.telemetry.decoder.ParseUserAgent;
 import com.mozilla.telemetry.transforms.DecompressPayload;
+import com.mozilla.telemetry.transforms.DeduplicateByDocumentId;
 import com.mozilla.telemetry.transforms.LimitPayloadSize;
 import com.mozilla.telemetry.transforms.NormalizeAttributes;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Optional;
 import org.apache.beam.sdk.Pipeline;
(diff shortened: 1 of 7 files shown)


whd:
I suppose it would be possible to test the performance of this before the heka ingestion code is ready by e.g. creating a job with a `BigQueryInput` on the prod live/decoded tables and attempting to write to the stage stable tables. AFAIK we haven't tested the BQ input either but the performance characteristics of each individual piece should be easily discernible.

jklukas:
> I suppose it would be possible to test the performance of this before the heka ingestion code is ready by e.g. creating a job with a `BigQueryInput` on the prod live/decoded tables and attempting to write to the stage stable tables. AFAIK we haven't tested the BQ input either but the performance characteristics of each individual piece should be easily discernible.

That sounds reasonable and worth doing. I agree we should test this independently from the Heka work.

PR #827, opened 2019-09-12
Add script to flatten schema hierarchy for updating bq datasets and tables

acmiyaguchi:
This adds a new script `copy-bq-schemas` that can be used in conjunction with `update-bq-tables` for testing out the BigQuery sink.

The following fixes were made:

* mounting `GOOGLE_APPLICATION_CREDENTIALS` into the mvn container
* adding `--clustering_fields` to the `bq mk` calls that cause failures in the beam job
* load schemas into a single dataset instead of multiple

I tested that the tables can be populated from the sampled-landfill data.

```
bin/download-sampled-landfill
gsutil cp landfill-integration.ndjson gs://amiyaguchi-dev/data/

# follow the rest of the ingestion-beam/integration_testing_workflow document
bin/download-schemas
bin/copy-bq-schemas
REPLACE_EXISTING=true bin/update-bq-table

export GOOGLE_APPLICATION_CREDENTIALS=...
BUCKET="gs://$PROJECT"

path="$BUCKET/data/*.ndjson"
bin/mvn compile exec:java -Dexec.args="\
    --runner=Dataflow \
    --project=$PROJECT \
    --autoscalingAlgorithm=NONE \
    --numWorkers=1 \
    --inputFileFormat=json \
    --inputType=file \
    --input=$path\
    --outputType=bigquery \
    --output=$PROJECT:test_ingestion.\${document_namespace}__\${document_type}_v\${document_version} \
    --bqWriteMethod=file_loads \
    --errorOutputType=file \
"
```

EDIT: This works properly now, the table must be removed in order for clustering to be effective. A new option has been added to overwrite the existing tables.

### DIFF
--- a/bin/mvn
+++ b/bin/mvn
@@ -30,6 +30,7 @@ docker run $INTERACTIVE_FLAGS --rm \
     -w /var/maven/project \
     -v ~/.m2:/var/maven/.m2 \
     -e MAVEN_CONFIG=/var/maven/.m2 \
-    -e GOOGLE_APPLICATION_CREDENTIALS \
+    -e GOOGLE_APPLICATION_CREDENTIALS=/app/.credentials \
+    -v $GOOGLE_APPLICATION_CREDENTIALS:/app/.credentials \
     maven:3-jdk-8 \
     mvn -Duser.home=/var/maven "${PROFILE[@]}" "$@"
--- a/ingestion-beam/bin/copy-bq-schemas
+++ b/ingestion-beam/bin/copy-bq-schemas
@@ -1,8 +1,8 @@
 #!/bin/bash
 
 # Copies BigQuery schemas for integration testing and validation from the
-# generated schema # archive fetched by `bin/download-schemas`. This script will
-# write to a flat # directory called `bq-schemas` containing BQ schemas.
+# generated schema archive fetched by `bin/download-schemas`. This script will
+# write to a flat directory called `bq-schemas` containing BQ schemas.
 
 set -e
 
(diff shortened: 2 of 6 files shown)


jklukas:
> On another note, it might be cleaner to set up a single dataset using the following naming convention "{namespace}__{doctype}_v{docversion}" so the single dataset can be deleted afterwards. Any thoughts on this suggestion?

That sounds fine to me and indeed would be easier to clean up.

bin/mvn
```diff
@@ -30,6 +30,7 @@ docker run $INTERACTIVE_FLAGS --rm \
     -w /var/maven/project \
     -v ~/.m2:/var/maven/.m2 \
     -e MAVEN_CONFIG=/var/maven/.m2 \
-    -e GOOGLE_APPLICATION_CREDENTIALS \
+    -e GOOGLE_APPLICATION_CREDENTIALS=/app/.credentials \
+    -v $GOOGLE_APPLICATION_CREDENTIALS:/app/.credentials \
```
jklukas: This is very nice.

ingestion-beam/bin/copy-bq-schemas
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Copies BigQuery schemas for integration testing and validation from the
+# generated schema # archive fetched by `bin/download-schemas`. This script will
+# write to a flat # directory called `bq-schemas` containing BQ schemas.
```
jklukas: There appear to be extraneous `#`s here.

acmiyaguchi:
@jklukas I've updated the script so it writes to a single dataset and updated the testing workflow to match. This looks like it works as expected after overwriting all of the existing schemas an running the BQ sink on ndjson data sampled from the AWS pipeline.

PR #832, opened 2019-09-16
Add script to flatten schema hierarchy for updating bq datasets and tables

acmiyaguchi:
This PR a duplicate of #827, without modifications to the `ingestion-beam/bin/mvn` script. Instead of modifying the script which is used in production, I've annotated the integration workflow document to note that maven should be run from the host machine.

### DIFF
--- a/bin/mvn
+++ b/bin/mvn
@@ -22,11 +22,12 @@ if [ -z "$NONINTERACTIVE" ]; then
     INTERACTIVE_FLAGS="-it"
 fi
 
-if [[ -f ${GOOGLE_APPLICATION_CREDENTIALS} ]]; then
+MOUNT_CREDENTIALS_FLAGS=""
+if [[ -f "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then
     # Credentials are detected on the local filesystem via environment. This is
     # mounted into the maven container into an arbitrary location, with the
     # container environment set to point at the mounted volume.
-    echo "Assuming local service account credentials..."
+    echo "Using GOOGLE_APPLICATION_CREDENTIALS..."
     mount_location="/app/.credentials"
     MOUNT_CREDENTIALS_FLAGS="-v ${GOOGLE_APPLICATION_CREDENTIALS}:${mount_location}"
     GOOGLE_APPLICATION_CREDENTIALS=$mount_location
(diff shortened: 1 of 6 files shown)


whd:
To expound a bit on the issue of #827, the [precedence](https://cloud.google.com/docs/authentication/production#providing_credentials_to_your_application) of `GOOGLE_APPLICATION_CREDENTIALS`, when set, precludes the use of the default service account supplied to a GCP instance. In the case of Jenkins, we run a GCP slave that [passes the standard metadata services networking into the container](https://github.com/mozilla-services/cloudops-infra-jenkinslib/blob/master/vars/gcp.groovy#L31), without setting `GOOGLE_APPLICATION_CREDENTIALS` explicitly, so that the container performing the build has access to the host jenkins service account. It looks like the original PR had `-v $GOOGLE_APPLICATION_CREDENTIALS:/app/.credentials` as an argument to docker, which when `GOOGLE_APPLICATION_CREDENTIALS` is unset should probably fail, but according to @acmiyaguchi did not.

Regardless, the wrapper script might instead check whether `GOOGLE_APPLICATION_CREDENTIALS` is set to an existing file, and in that case set the environment and secrets mounting options to docker, but to do neither of those things otherwise (which would emulate the previous behavior on Jenkins of passing in an empty `GOOGLE_APPLICATION_CREDENTIALS`). This seems to me like the behavior most consistent with both user and machine expectations.

acmiyaguchi:
I used the same instructions from #827 to test this. This works properly in the local development workflow, as well as on a GCE instance when the variable is unset.

bin/mvn
```diff
@@ -22,6 +22,16 @@ if [ -z "$NONINTERACTIVE" ]; then
     INTERACTIVE_FLAGS="-it"
 fi
 
+if [[ -f ${GOOGLE_APPLICATION_CREDENTIALS} ]]; then
```
whd: My style preference is to always quote arguments that you don't need the shell to parse in any way, which is consistent with this file's current style.

```suggestion
if [[ -f "$GOOGLE_APPLICATION_CREDENTIALS" ]]; then
```

bin/mvn
```diff
@@ -31,5 +41,6 @@ docker run $INTERACTIVE_FLAGS --rm \
     -v ~/.m2:/var/maven/.m2 \
     -e MAVEN_CONFIG=/var/maven/.m2 \
     -e GOOGLE_APPLICATION_CREDENTIALS \
+    ${MOUNT_CREDENTIALS_FLAGS} \
```
whd: For stylistic consistency:

```suggestion
    $MOUNT_CREDENTIALS_FLAGS \
```

PR #840, opened 2019-09-20
Support running Decoder without Redis

### LINKED ISSUE #839: Support running Decoder without Redis
For various testing scenarios or batch jobs, we may want to run the Decoder without having Redis-based deduplication.

We should skip deduplication when no redis url is configured.

cc @wlach @whd

jklukas:
Closes #839

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/DeduplicateTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/DeduplicateTest.java
@@ -11,6 +11,7 @@ import com.mozilla.telemetry.transforms.WithErrors;
 import java.util.Arrays;
 import java.util.UUID;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider;
 import org.apache.beam.sdk.testing.PAssert;
 import org.apache.beam.sdk.testing.TestPipeline;
 import org.apache.beam.sdk.transforms.Create;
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/DeduplicateTest.java
```diff
@@ -100,6 +101,16 @@ public void testOutput() {
 
     // run RemoveDuplicates
     pipeline.run();
+
+    PCollection<String> nonDedupledIds = pipeline
+        .apply("ids", Create.of(Arrays.asList(newId, duplicatedId)))
+        .apply("create messages", mapStringsToId)
+        .apply("no-op deduplicate", Deduplicate.removeDuplicates(StaticValueProvider.of(null)))
+        .ignoreDuplicates().output().apply("get non-deduped IDs", mapMessagesToId);
+    PAssert.that(nonDedupledIds).containsInAnyOrder(newId, duplicatedId);
```
wlach: It doesn't appear that we're testing what should be the case here, that is, that the duplicates aren't removed -- there don't appear to be any duplicates in this list.

jklukas: Thank you. You're right. Will fix.

jklukas: Turns out the test _is_ valid, but difficult to reason through. I've added new comments to explain the intent.

By the time this test is running, we've already sent two IDs through Redis to be marked as seen. If we passed the test redis uri here, then IDs that have already been seen would be removed, thus `seenId` would not show up in the output. But the `seenId` passes through when uri is null.

PR #866, opened 2019-09-27
Add timestamps in HekaIO

jklukas:
We weren't populating submission_timestamp in BQ, and we also need record timestamps if we want to write out with windowing.


ingestion-beam/src/test/java/com/mozilla/telemetry/heka/HekaIOTest.java
```diff
@@ -0,0 +1,143 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
wlach: It's not great to move files in a PR while making changes, as it makes it quite difficult to review what actually changed. I don't think it's that big a deal in this case, but something to bear in mind for the future.

wlach:
Note that the unit tests seem to be failing for some reason, this will need to be fixed before this is merged

PR #877, opened 2019-10-01
Window by triggeringFrequency in BQ batch writes

jklukas:
Also adds windowing to DedpulicateByDocumentId to allow processing multiple days in a single batch job, as we will want to do for a 1% backfill.


jklukas:
I ran a successful test last night with `triggeringFrequency=1h` although we still had a very long tail waiting for BQ load jobs to complete. Given the most recent feedback we have from Google about load failures, this change may not guarantee us the stability we were hoping for. But we can configure the window to be full days if we determine smaller windows don't help.

PR #887, opened 2019-10-02
Replace maven inter-module deps with copying src

relud:
This allows running maven in module directories and removes the need for running maven with `-P {module}` and `-DfailIfNoTests=false`, while avoiding duplication of maven plugin configuration, shared code, and shared dependencies.

### DIFF
--- a/docs/ingestion-beam/sink-job.md
+++ b/docs/ingestion-beam/sink-job.md
@@ -174,14 +174,15 @@ Note: `-Dexec.args` does not handle newlines gracefully, but bash will remove
 
 ### Locally
 
-If you install Java and maven, you can invoke `mvn -P ingestion-beam` from the
-git root in the following commands instead of using `./bin/mvn` from the
-`ingestion-beam/` directory; be aware, though, that Java 8 is the target JVM
+If you install Java and maven, you can invoke `mvn` in the following commands
+instead of using `./bin/mvn`; be aware, though, that Java 8 is the target JVM
 and some reflection warnings may be thrown on newer versions, though these are
 generally harmless.
 
-The provided `bin/mvn` script downloads and runs maven with the correct profile
-via docker so that less setup is needed on the local machine.
+The provided `bin/mvn` script downloads and runs maven via docker so that less
+setup is needed on the local machine. For prolonged development performance is
+likely to be significantly better, especially in osx, if `mvn` is installed and
+run natively without docker.
 
 ```bash
 # create a test input file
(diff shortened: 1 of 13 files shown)


ingestion-beam/checkstyle
```diff
@@ -0,0 +1 @@
+../checkstyle/
```
relud: softlinks are necessary for checkstyle to work in both submodule directories and the parent module directory.

docs/ingestion-beam/sink-job.md
```diff
@@ -174,14 +174,13 @@ Note: `-Dexec.args` does not handle newlines gracefully, but bash will remove
...
-`ingestion-beam/` directory; be aware, though, that Java 8 is the target JVM
+If you install Java and maven, you can invoke `mvn` in the following commands
+instead of using `./bin/mvn`; be aware, though, that Java 8 is the target JVM
 and some reflection warnings may be thrown on newer versions, though these are
 generally harmless.
 
-The provided `bin/mvn` script downloads and runs maven with the correct profile
-via docker so that less setup is needed on the local machine.
+The provided `bin/mvn` script downloads and runs maven via docker so that less
+setup is needed on the local machine.
```
jklukas: Can we go ahead and add a note here saying that `bin/mvn` is intended for getting started quickly and that for prolonged development, performance will be much better if you install the tools locally?

relud: added ```For prolonged development performance is
likely to be significantly better, especially in osx, if `mvn` is installed and
run natively without docker.```

docs/ingestion-beam/sink-job.md
```diff
@@ -174,14 +174,15 @@ Note: `-Dexec.args` does not handle newlines gracefully, but bash will remove
...
+If you install Java and maven, you can invoke `mvn` in the following commands
+instead of using `./bin/mvn`; be aware, though, that Java 8 is the target JVM
 and some reflection warnings may be thrown on newer versions, though these are
 generally harmless.
 
-The provided `bin/mvn` script downloads and runs maven with the correct profile
-via docker so that less setup is needed on the local machine.
+The provided `bin/mvn` script downloads and runs maven via docker so that less
+setup is needed on the local machine. For prolonged development performance is
+likely to be significantly better, especially in osx, if `mvn` is installed and
```
jklukas:
```suggestion
likely to be significantly better, especially in MacOS, if `mvn` is installed and
```

pendantry

relud: wfm

pom.xml
```diff
@@ -235,6 +244,66 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <version>3.0.0</version>
+                <executions>
+                    <!-- add ingestion-core/src to all modules -->
```
jklukas: I would not have thought of generate-sources as a method for making this work. This is cool!

PR #915, opened 2019-10-10
Make sure we preserve clustering definition by using TableRowCoderV3

jklukas:
We may have been losing the clustering definition based on errors we see in large-scale batch jobs.


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -72,8 +71,6 @@ public static PubsubMessageToTableRow of(ValueProvider<List<String>> strictSchem
 
   public static final String PAYLOAD = "payload";
   public static final String ADDITIONAL_PROPERTIES = "additional_properties";
-  public static final TimePartitioning TIME_PARTITIONING = new TimePartitioning()
-      .setField(Attribute.SUBMISSION_TIMESTAMP);
```
jklukas: Also removing this vestigial constant.

PR #918, opened 2019-10-11
Add patch for clustering with many writers

jklukas:
BigQueryIO does a kind of weird thing where it spills over writers and
reshuffles data if there are lots of destination tables or high volume of data.
Turns out I missed a case of passing the clustering configuration on that only
crops up in this case where data is reshuffled.

We replace the underlying file from Beam to get the change immediately,
but I will PR the same to upstream Beam so we can remove this after the next
beam release (see https://github.com/mozilla/gcp-ingestion/issues/919).

I also add another patch here for BatchLoads to change the maximum size per partition. I found that my effort at windowing using the `triggeringFrequency` parameter is useless because BigQueryIO.Write immediately rewindows to the global window. Changing the maximum partition size appears to be the only way to force smaller loads, but there's no user-level hook exposed to change it. I will also file a beam JIRA issue to discuss making more of these options configurable at the user level.


ingestion-beam/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/DynamicDestinationsHelpers.java
```diff
@@ -0,0 +1,407 @@
...
+      if (existingTable == null) {
+        return wrappedDestination;
+      } else {
+        return new TableDestination(
+            wrappedDestination.getTableSpec(),
+            existingTable.getDescription(),
+            existingTable.getTimePartitioning(),
+            // TODO: Remove this file; the following line is a custom Mozilla bug fix which we
+            // plan to fix in upstream Beam.
+            existingTable.getClustering());
```
jklukas: This is the one line added in this file. Everything else is directly copied.

ingestion-beam/src/main/java/org/apache/beam/sdk/io/gcp/bigquery/BatchLoads.java
```diff
@@ -0,0 +1,638 @@
...
+  // Maximum number of files in a single partition.
+  static final int DEFAULT_MAX_FILES_PER_PARTITION = 10000;
+
+  @VisibleForTesting
+  // TODO: Remove this custom lower threshold; this forces a day of main pings
+  // to be split to about 100 pieces, sidestepping the bug where a single large
+  // BQ load into a clustered tables often fails with memory exceeded errors.
+  static final long DEFAULT_MAX_BYTES_PER_PARTITION = 100 * (1L << 30);
+  // // Maximum number of bytes in a single partition -- 11 TiB just under BQ's 12 TiB limit.
+  // static final long DEFAULT_MAX_BYTES_PER_PARTITION = 11 * (1L << 40);
```
jklukas: This is the one change in this file. Otherwise, it is copied verbatim from Beam 2.16.

PR #926, opened 2019-10-15
Support sanitized landfill input files

jklukas:
Also, support "ignore" option for output.

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -49,6 +49,11 @@
         </dependency>
         <dependency>
             <groupId>org.apache.beam</groupId>
+            <artifactId>beam-sdks-java-io-amazon-web-services</artifactId>
+            <version>${beam.version}</version>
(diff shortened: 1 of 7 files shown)


ingestion-beam/pom.xml
```diff
@@ -47,6 +47,11 @@
             <artifactId>beam-sdks-java-extensions-json-jackson</artifactId>
             <version>${beam.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.apache.beam</groupId>
+            <artifactId>beam-sdks-java-io-amazon-web-services</artifactId>
+            <version>${beam.version}</version>
+        </dependency>
```
jklukas: We probably don't want to merge in this dependency, but the other code in here is probably safe to add for potential future use.

relud: i would be okay with merging this dependency and removing it later when the s3 landfill is deleted.

Issue #948, opened 2019-10-25
Coerce all values to match expected types in BigQuery sink


jklukas:
Currently, we coerce values to string where string is expected. This allows us to handle userPrefs and histograms.

We have run into an issue where we sometimes see a boolean show up in an int column (see https://bugzilla.mozilla.org/show_bug.cgi?id=1590559) and should handle that case, and perhaps the more general case of values not matching expectations.

We could cast these booleans to integers or we could insert null in these situations. Perhaps if we insert null we should put the value in additionalProperties.

relud:
> We could cast these booleans to integers or we could insert null in these situations. Perhaps if we insert null we should put the value in additionalProperties.

+1 to cast if possible, otherwise insert null and put the value in `additional_properties`

jklukas:
This is done in #951

PR #951, opened 2019-10-28
Add additional type coercions to BigQueryToTableRow

jklukas:
Addresses https://github.com/mozilla/gcp-ingestion/issues/948

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -373,8 +373,21 @@ public class PubsubMessageToTableRow
       if (!repeatedAdditionalProperties.stream().allMatch(Objects::isNull)) {
         additionalProperties.put(jsonFieldName, repeatedAdditionalProperties);
       }
+      // This is a basic type.
     } else {
-      value.ifPresent(v -> parent.put(name, v));
+      value.ifPresent(v -> {
+        Object coerced = coerceToBqType(v, field);
+        // A null coerced value means the actual type didn't match expected and we don't define
+        // a coercion. We put the value to additional_properties instead.
+        if (coerced == null) {
+          if (additionalProperties != null) {
+            additionalProperties.put(jsonFieldName, v);
+          }
+          parent.remove(name);
+        } else {
+          parent.put(name, coerced);
+        }
+      });
     }
   }
 
@@ -462,6 +475,44 @@ public class PubsubMessageToTableRow
     });
   }
 
+  /**
+   * This method gives us a chance to perform some additional type coercions in case the BigQuery
+   * field type is different from the source data type. This should rarely happen, since only
+   * validated payloads get through to this BQ sink path, but there are sets of probes with
+   * heterogeneous types that appear as explicit fields in BQ, but are treated as loosely typed
+   * maps at the validation phase; we need to catch these or they can cause the entire pipeline
+   * to stall.
+   *
+   * <p>Returning {@code null} here indicates that no coercion is defined and that the field should
+   * be put to {@code additional_properties}.
+   */
+  private static Object coerceToBqType(Object o, Field field) {
+    if (field.getMode() == Mode.REPEATED) {
+      return o;
+    } else {
+      if (field.getType() == LegacySQLTypeName.STRING) {
+        return coerceToString(o);
+      } else if (field.getType() == LegacySQLTypeName.INTEGER) {
+        if (o instanceof Integer || o instanceof Long) {
+          return o;
+        } else if (o instanceof Boolean) {
+          // We assume that false is equivalent to zero and true to 1.
+          return (Boolean) o ? 1 : 0;
+        } else {
+          return null;
+        }
+      } else if (field.getType() == LegacySQLTypeName.BOOLEAN) {
+        if (o instanceof Boolean) {
+          return o;
+        } else {
+          return null;
+        }
+      } else {
+        return o;
+      }
+    }
+  }
+
   private static String coerceToString(Object o) {
     if (o instanceof String) {
       return (String) o;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -427,7 +460,7 @@ public void testDoublyNestedList() throws Exception {
         + "}\n").getBytes(StandardCharsets.UTF_8));
     List<Field> bqFields = ImmutableList.of(Field.newBuilder("payload", LegacySQLTypeName.RECORD, //
         Field.newBuilder("list", LegacySQLTypeName.RECORD, //
-            Field.of("list", LegacySQLTypeName.INTEGER)) //
+            Field.newBuilder("list", LegacySQLTypeName.INTEGER).setMode(Mode.REPEATED).build()) //
```
jklukas: This field should have been defined as REPEATED all along in this test, but the new code changes introduced in this PR caused the inner lists to get incorrectly interpreted as int fields and sent to additional_properties. Changing the mode here causes the test to pass again.

jklukas:
> This looks good to me. In short, (boolean -> integer) is valid, while (not(boolean) -> boolean) will get sent to additional properties. This should cover all the cases in scalars section in the main ping where this problem started because the types are limited to boolean, integers, and strings.

This also covers (string -> integer) getting sent to additional properties.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -462,6 +475,44 @@ private void expandNestedListType(String jsonFieldName, Object val, Field field,
...
+   * to stall.
+   *
+   * <p>Returning {@code null} here indicates that no coercion is defined and that the field should
+   * be put to {@code additional_properties}.
+   */
+  private static Object coerceToBqType(Object o, Field field) {
+    if (field.getMode() == Mode.REPEATED) {
+      return o;
+    } else {
+      if (field.getType() == LegacySQLTypeName.STRING) {
```
relud: this condition will never evaluate as true when called from `processField`

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -462,6 +475,44 @@ private void expandNestedListType(String jsonFieldName, Object val, Field field,
...
+   * field type is different from the source data type. This should rarely happen, since only
+   * validated payloads get through to this BQ sink path, but there are sets of probes with
+   * heterogeneous types that appear as explicit fields in BQ, but are treated as loosely typed
+   * maps at the validation phase; we need to catch these or they can cause the entire pipeline
+   * to stall.
+   *
+   * <p>Returning {@code null} here indicates that no coercion is defined and that the field should
+   * be put to {@code additional_properties}.
+   */
+  private static Object coerceToBqType(Object o, Field field) {
```
relud: this seems like a case where we should have return type `Optional<Object>` and be returning `Optional.empty()`, rather than `null`

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
jklukas: You are correct. I was attempting to change the existing code as little as possible, but it should be possible to remove one of the cases in `processField`.

Really, I should remove a few of the cases and handle coercing `List`s as well. I think I'll go ahead and make those changes.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -462,6 +475,44 @@ private void expandNestedListType(String jsonFieldName, Object val, Field field,
...
+    if (field.getMode() == Mode.REPEATED) {
+      return o;
+    } else {
+      if (field.getType() == LegacySQLTypeName.STRING) {
+        return coerceToString(o);
+      } else if (field.getType() == LegacySQLTypeName.INTEGER) {
```
relud: I think it would be more consistent with the existing code to add cases for `field.getType() == LegacySQLTypeName.INTEGER && field.getMode() != Mode.REPEATED` and `field.getType() == LegacySQLTypeName.BOOLEAN && field.getMode() != Mode.REPEATED` to `processField`, rather than breaking it out into a separate function. this would also make it more obvious that the case of `field.getType() == LegacySQLTypeName.STRING` is already handled.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: see also my other comment about whether this shouldn't be in a separate function

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
jklukas: I'm inclined to remove both of the existing cases that do string coercion and pull that all into a single coerce method as entrypoint.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -462,10 +462,62 @@ private void expandNestedListType(String jsonFieldName, Object val, Field field,
...
+   */
+  private static Optional<Object> coerceToBqType(Object o, Field field) {
+    if (field.getMode() == Mode.REPEATED) {
+      if (o instanceof List) {
+        return Optional.of(((List<Object>) o).stream().map(v -> coerceSingleValueToBqType(v, field))
+            // We have not yet observed a case where an array type contains values that cannot
+            // be coerced to appropriate values, so this filter should not be getting rid of
+            // elements; a future improvement would be to refactor so that we can insert
+            // non-coerceable values from a list into additional_properties.
+            .filter(Optional::isPresent).map(Optional::get).collect(Collectors.toList()));
```
jklukas: We now coerce objects within a list here. This handles the case of JSON-ifying items in a list that was previously done in `processField`.

jklukas:
Added metrics to keep track of how often these coercions are being applied. Numbers should be low except for string coercions, which are expected for userPrefs and histograms.

Issue #964, opened 2019-10-31
Null pointer exception with raw streams republisher


whd:
Example error GCS path: gs://moz-fx-data-prod-data/telemetry-raw_republisher/error/2019-10-31/18/error-2019-10-31T18-30-00.000Z-2019-10-31T18-40-00.000Z-0-00039-of-00060.ndjson.gz

Stack trace:

    java.net.URI$Parser.parse(URI.java:3042)
    java.net.URI.<init>(URI.java:588)
    java.net.URI.create(URI.java:850)
    org.apache.beam.sdk.options.ValueProvider$NestedValueProvider.get(ValueProvider.java:129)
    com.mozilla.telemetry.decoder.Deduplicate$MarkAsSeen.processElement(Deduplicate.java:220)
    com.mozilla.telemetry.decoder.Deduplicate$MarkAsSeen.processElement(Deduplicate.java:189)
    com.mozilla.telemetry.transforms.MapElementsWithErrors$DoFnWithErrors.processElementOrError(MapElementsWithErrors.java:85)
    com.mozilla.telemetry.transforms.MapElementsWithErrors$DoFnWithErrors$DoFnInvoker.invokeProcessElement(Unknown Source)
    org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.invokeProcessElement(SimpleDoFnRunner.java:218)
    org.apache.beam.runners.dataflow.worker.repackaged.org.apache.beam.runners.core.SimpleDoFnRunner.processElement(SimpleDoFnRunner.java:183)
    org.apache.beam.runners.dataflow.worker.SimpleParDoFn.processElement(SimpleParDoFn.java:335)
    org.apache....

[...]

PR #967, opened 2019-11-01
Allow null redisUri in Republisher

jklukas:
Closes https://github.com/mozilla/gcp-ingestion/issues/964

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
@@ -7,6 +7,7 @@ import com.mozilla.telemetry.util.Time;
 import java.net.URI;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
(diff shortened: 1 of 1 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublisherOptions.java
```diff
@@ -143,7 +144,8 @@ static void enrichRepublisherOptions(Parsed options) {
     options
         .setDeduplicateExpireSeconds(NestedValueProvider.of(options.getDeduplicateExpireDuration(),
             value -> Ints.checkedCast(Time.parseSeconds(value))));
-    options.setParsedRedisUri(NestedValueProvider.of(options.getRedisUri(), URI::create));
+    options.setParsedRedisUri(NestedValueProvider.of(options.getRedisUri(),
+        s -> Optional.ofNullable(s).map(URI::create).orElse(null)));
```
jklukas: This now matches exactly what we do in `DecoderOptions`. I just forgot that RepublisherOptions needs to do the same thing.

Issue #935, opened 2019-10-19
activity-stream records all have sample_id=51


jklukas:
I had expected that if the pipeline finds no `client_id` or `clientId` field, sample_id would be null, but it appears it's always 51. We should figure out why and potentially use `impression_id` to calculate sample_id in the case in the case that `client_id` is not present.

jklukas:
Found it. The `activity-stream` schemas do have a client_id field that is always populated with "n/a"; this is an artifact of the Tiles pipeline where the schema was shared with other types that might have a client_id set.

The crc32 sample_id calculation yields 51 for value "n/a".

So we should probably check that the field is a valid UUID before creating sample_id.

PR #969, opened 2019-11-01
Support sample_id for activity-stream pings

jklukas:
Currently, all activity-stream pings are being assigned `sample_id=51`
because they all contain `"client_id":"n/a"`. The pipeline currently discovers
that value, loads it as an attributes, and happily calculates a sample_id from
it.

The change here involves requiring that we have a valid UUID for clientIds
and then also supporting the `impression_id` field in activity-stream pings
that actually does contain a UUID representing the client.

Closes https://github.com/mozilla/gcp-ingestion/issues/935

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -17,7 +17,9 @@ import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
+import java.util.UUID;
 import java.util.stream.Stream;
 import java.util.zip.CRC32;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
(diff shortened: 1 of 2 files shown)


jklukas:
cc @mreid-moz as this is relevant to your plan for tiered storage and keeping only subsets of older data.

Also cc @whd.

whd:
It would be nice to have a general mechanism (e.g. schema metadata) to identify the field to be used for sampling without having to write special cases, but looking at the code the impression id needs parsing anyway so that's probably unrealistic.

> The change here involves requiring that we have a valid UUID for clientIds
and then also supporting the impression_id field in activity-stream pings
that actually does contain a UUID representing the client.

That client id should be a valid UUID is probably documented somewhere. Does "require" here mean that having a bad client id will be a new class of decoder error? In the past when we've made UUID parsing stricter (for document ids, aws -> gcp) we did hit the eng-workflow case, so making sure that's not going to affect any present or future data is probably a good idea.

jklukas:
> It would be nice to have a general mechanism (e.g. schema metadata) to identify the field to be used for sampling without having to write special cases, but looking at the code the impression id needs parsing anyway so that's probably unrealistic.

The parsing needed for impression_id is unfortunate, but at least it's still a UUID display variant, so fairly easy to accomodate generically. I agree schema metadata would a nice way to go, but probably not yet worth investing in until we have a clearer value proposition.

> That client id should be a valid UUID is probably documented somewhere. Does "require" here mean that having a bad client id will be a new class of decoder error? In the past when we've made UUID parsing stricter (for document ids, aws -> gcp) we did hit the eng-workflow case, so making sure that's not going to affect any present or future data is probably a good idea.

We already check sampleId is a UUID via regex in the common ping schemas, so pings with an invalid UUID for clientId should already be getting rejected. In any case, the code change here would not cause any new validation errors; it could only potentially mean that a record ends up with a null sample_id (which seems better to me than getting a bogus static value like we currently do for activity-stream payloads).

PR #991, opened 2019-11-18
Bug 1595899 - Handle bigquery maps without value field

relud:


ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
```diff
@@ -149,6 +154,17 @@ public void testUnmap() throws Exception {
...
 
+  @Test
+  public void testUnmapWithoutValue() throws Exception {
+    Map<String, Object> parent = new HashMap<>();
+    Map<String, Object> additionalProperties = new HashMap<>();
+    parent.put("mapField", new HashMap<>(ImmutableMap.of("foo", 3, "bar", 4)));
+    List<Field> bqFields = ImmutableList.of(MAP_FIELD_WITHOUT_VALUE);
+    String expected = "{\"map_field\":[{\"key\":\"bar\"},{\"key\":\"foo\"}]}";
+    TRANSFORM.transformForBqSchema(parent, bqFields, additionalProperties);
+    assertEquals(expected, Json.asString(parent));
```
acmiyaguchi: Could you add a test case for the expected structure of the `additional_properties` field when the schema specifies a valueless map, please? For the concrete example of XPI installs, I'm wondering how the value for a specific key can be accessed using `JSON_EXTRACT` udf.

relud:
> I'm worried that it will be difficult to recover the values from a map if they've been stripped of their keys in additional properties. We should probably keep the keys in additional properties.

I wasn't planning to strip out the keys, because `additional_properties` retains the original ping format, not the array-of-entries bigquery format.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -432,17 +434,24 @@ private void expandMapType(String jsonFieldName, Object val, Field field,
...
     value.filter(Map.class::isInstance).map(Map.class::cast).ifPresent(m -> {
       Map<String, Object> map = m;
-      Field valueField = field.getSubFields().get(1);
       Map<String, Object> props = additionalProperties == null ? null : new HashMap<>();
+      Optional<Field> valueFieldOption;
+      if (field.getSubFields().size() == 2) {
+        valueFieldOption = Optional.of(field.getSubFields().get(1));
+      } else {
+        valueFieldOption = Optional.empty();
+        props.putAll(map);
```
relud: pretty sure this needs a null check:
```suggestion
        if (props != null) {
          props.putAll(map);
        }
```

PR #1000, opened 2019-11-20
Prefer normalizing os to android over linux for non-desktop pings

### LINKED ISSUE #999: normalized_os may be incorrect for mobile
In AWS we differentiated between mobile pings and everything else when normalizing os[0], but in GCP we don't[1], resulting in cases where mobile pings may have a normalized os of `Linux` instead of `Android` or `Other`[2]

[0] https://github.com/mozilla-services/lua_sandbox_extensions/blob/e708139/moz_telemetry/io_modules/decoders/moz_ingest/telemetry.lua#L260-L271
[1] https://github.com/mozilla/gcp-ingestion/blob/cd7f89d/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java#L64-L85
[2] https://github.com/mozilla/bigquery-etl/pull/491/files#diff-4e8efa269829ac96a57ff7dffc6bdd99R93-R96

relud:
Not sure if we actually want to do this, but providing an implementation just in case.

fixes #999

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
@@ -53,20 +53,30 @@ public class NormalizeAttributes
    * Normalize an operating system string to one of the three major desktop platforms,
    * one of the two major mobile platforms, or "Other".
    *
-   * <p>Reimplements and combines the logic of {@code os} and {@code mobile_os} from
+   * <p>Previously implemented by {@code os} and {@code mobile_os} from
    * https://github.com/mozilla-services/lua_sandbox_extensions/blob/607b3c7b9def600ddbd373a14f67a55e1c69f7a1/moz_telemetry/modules/moz_telemetry/normalize.lua#L184-L215
    */
-  public static String normalizeOs(String name) {
-    if (name.startsWith("Windows") || name.startsWith("WINNT")) {
-      return WINDOWS;
-    } else if (name.startsWith("Darwin")) {
-      return MAC;
-    } else if (name.contains("Linux") || name.contains("BSD") || name.contains("SunOS")) {
-      return LINUX;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
```diff
@@ -53,20 +53,30 @@ public static String normalizeChannel(String name) {
...
-    if (name.startsWith("Windows") || name.startsWith("WINNT")) {
-      return WINDOWS;
-    } else if (name.startsWith("Darwin")) {
-      return MAC;
-    } else if (name.contains("Linux") || name.contains("BSD") || name.contains("SunOS")) {
-      return LINUX;
-    } else if (name.startsWith("iOS") || name.contains("iPhone")) {
+  public static String normalizeOs(String name, Optional<String> normalizedAppName) {
+    boolean appIsNotDesktop = normalizedAppName //
+        .map(v -> !OTHER.equals(v) && !FIREFOX.equals(v)).orElse(false);
```
jklukas:
```suggestion
        .filter(v -> !OTHER.equals(v) && !FIREFOX.equals(v)).isPresent();
```

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
```diff
@@ -53,20 +53,30 @@ public static String normalizeChannel(String name) {
...
+    // Don't look for desktop os names if app is not desktop
+    if (!appIsNotDesktop) {
```
jklukas: There are logically two "not"s here, which we should avoid. I think perhaps we invert the boolean and call it `mightBeDesktop`, so that this condition becomes `if (mightBeDesktop)`.

relud:
closing per https://github.com/mozilla/gcp-ingestion/issues/999#issuecomment-561898626

PR #1013, opened 2019-11-27
Implement BigQuery mixed mode

### LINKED ISSUE #505: Create a process to periodically load files into BigQuery
When ingestion-beam is set to output to the avro output type, it will generate avro files on GCS into paths derived from a template and fully qualified names (i.e. namespace, type, and version). The files will be written to disk every 5-10 minutes.

BigQuery has [native support for loading avro from GCS](https://cloud.google.com/bigquery/docs/loading-data-cloud-storage-avro). There are a few options for handling mini-batching of files into BQ:

1. A cron/airflow job that runs the CLI tool every 5-10 minutes
2. A beam pipeline with a file watcher

The set of files can be determined enumerating fully qualified names in mozilla-pipeline-schemas. 

This is related to #280, #353, and #372.

### LINKED ISSUE #751: Support file loads in ingestion-sink
cc @relud - You said you already had a branch going with an implementation of file loads to BQ for ingestion-sink, but I wanted to write down my thoughts to see if they mesh with your design.

Based on my understanding of the Java client for GCS, we are able to open a writeable stream and pass infinite amounts of data to it, with the [client handling uploading the data in chunks](https://github.com/googleapis/google-cloud-java/blob/master/google-cloud-clients/google-cloud-storage/src/main/java/com/google/cloud/storage/BlobWriteChannel.java#L50-L69), by default ~2 MB per chunk, so we can open a single blob per table and not have to worry about logic around flushing periodically.

Furthermore, BQ is able to read a single GCS blob in parallel in the case of unzipped ndjson files. And it supports loading uncompressed files up to 5 TB. So we can theoretically send data to a single file in GCS per table and issue a load job with that single object. The BQ docs also indicate that loading uncompressed json is faster than loading compressed.

So, the simplest viable implementation of file loads for ingestion-sink would look roughly like:

- Alter the `Batch` class to have a default timeout of 2 to 5 minutes rather than 1 second
- Alter the `Batch` class to open a GCS blob channel and write each row to that rather than accumulating a list
- Have the `insertAll` method close the channel, initiate a BQ load job pointing to the blob, and then wait for the load job to finish; on finish, delete the blob and return success or throw an exception to cause all the messages to nack

I assume we'd actually create `StreamingBatch` and `FileLoadBatch` subclasses to be able to handle either case.

relud:
fixes #751 
fixes #505


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -72,31 +84,58 @@ Output getOutput(Env env) {
...
+        return 1_000_000_000L; // 1GB
+      }
+    },
+    bigQueryLoad {
+
+      @Override
+      Output getOutput(Env env) {
+        return new Output(env, this, new BigQuery.Load(getBigQueryService(env), getGcsService(env),
+            // BigQuery Load API limits maximum bytes per request to 15TB
+            env.getLong(LOAD_MAX_BYTES, 15_000_000_000_000L), // default 15TB
```
jklukas: We patched beam in ingestion-beam to lower this limit, since load jobs start failing on main ping data as the data volume gets to this magnitude. We're using 100 GB as the limit currently.

Though we're unlikely to get anywhere near this volume for streaming jobs.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -72,31 +84,58 @@ Output getOutput(Env env) {
...
-                env.getLong(BATCH_MAX_BYTES, 100_000_000L), // default 100MB
-                env.getInt(BATCH_MAX_MESSAGES, 1_000_000), // default 1M messages
-                env.getDuration(BATCH_MAX_DELAY, "10m"), // default 10 minutes
-                PubsubMessageToTemplatedString.of(gcsPrefix), getFormat(env), batchCloseHook));
+      @Override
+      Output getOutput(Env env) {
+        Output pubsubWrite = pubsub.getOutput(env);
+        return new Output(env, this, new Gcs.Write.Ndjson(getGcsService(env),
+            // BigQuery Load API limits maximum bytes per file to 4GB compressed or 5TB uncompressed
+            env.getLong(BATCH_MAX_BYTES, 100_000_000L), // default 100MB
```
jklukas: This value seems low compared to what's discussed the comment.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -109,22 +148,67 @@ long getDefaultMaxOutstandingRequestBytes() {
...
+            // 10MB API limit by default.
+            env.getLong(BATCH_MAX_BYTES, 6_000_000L), // default 6MB
+            // BigQuery Streaming API Limits maximum rows per request to 10,000
+            env.getInt(BATCH_MAX_MESSAGES, 10_000), // default 10K messages
+            env.getDuration(BATCH_MAX_DELAY, "1s"), // default 1 second
+            PubsubMessageToTemplatedString.forBigQuery(env.getString(OUTPUT_TABLE)),
+            getFormat(env)));
+      }
+    },
+    bigQueryMixed {
```
jklukas: Are we allowed to add an extra newline between enum values, or does spotless remove them? It's killing me that there's no space between cases, but there's a newline before each method.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -109,22 +148,67 @@ long getDefaultMaxOutstandingRequestBytes() {
...
+
+      @Override
+      Output getOutput(Env env) {
+        final com.google.cloud.bigquery.BigQuery bigQuery = getBigQueryService(env);
+        final Storage storage = getGcsService(env);
+        Function<PubsubMessage, CompletableFuture<Void>> bigQueryLoad = new BigQuery.Load(bigQuery,
+            storage,
+            // BigQuery Load API limits maximum bytes per request to 15TB
+            env.getLong(LOAD_MAX_BYTES, 15_000_000_000_000L), // default 15TB
```
jklukas: Given that this value is used in two places, it should be a named constant, probably `DEFAULT_LOAD_MAX_BYTES`. That likely applies to max files too.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -109,22 +148,67 @@ long getDefaultMaxOutstandingRequestBytes() {
...
+            // efficient than the JSON that actually gets sent over HTTP, so we use to 60% of the
+            // 10MB API limit by default.
+            env.getLong(STREAMING_BATCH_MAX_BYTES, 6_000_000L), // default 6MB
+            // BigQuery Streaming API Limits maximum rows per request to 10,000
+            env.getInt(STREAMING_BATCH_MAX_MESSAGES, 10_000), // default 10K messages
+            env.getDuration(STREAMING_BATCH_MAX_DELAY, "1s"), // default 1 second
+            PubsubMessageToTemplatedString.forBigQuery(env.getString(OUTPUT_TABLE)),
+            getFormat(env));
+        // BigQuery Streaming API Limits maximum row size to 1MiB
+        long maxStreamingSize = env.getLong(STREAMING_MESSAGE_MAX_BYTES, 1_000_000L); // default 1MB
```
jklukas: Note that Beam in a few places constructs default values in GiB/TiB like `(1L << 20)` for MiB. If we want to use exact values that match BQ limits, then I like that format. But `1_000_000L` is probably overall clearer to readers.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -142,15 +226,25 @@ long getDefaultMaxOutstandingRequestBytes() {
...
+      if (env.containsKey(OUTPUT_BUCKET) && !hasBigQueryOutputMode) {
         return OutputType.gcs;
-      } else if (env.containsKey(OUTPUT_TOPIC)) {
+      } else if (env.containsKey(OUTPUT_TOPIC) && !hasBigQueryOutputMode) {
         return OutputType.pubsub;
-      } else if (env.containsKey(OUTPUT_TABLE)) {
-        return OutputType.bigQuery;
+      } else if (env.containsKey(OUTPUT_TABLE) || hasBigQueryOutputMode) {
+        final String outputMode = env.getString(BIG_QUERY_OUTPUT_MODE, "streaming").toLowerCase();
+        if (outputMode.equals("streaming")) {
```
jklukas: Because we're immediately returning or throwing in all cases, this could likely be an elegant switch statement.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -142,15 +226,25 @@ long getDefaultMaxOutstandingRequestBytes() {
...
+          return OutputType.bigQueryMixed;
+        }
+        if (outputMode.equals("file_loads")) {
+          return OutputType.bigQueryFiles;
+        }
+        throw new IllegalArgumentException("Unsupported BIG_QUERY_OUTPUT_MODE: " + outputMode);
       } else {
-        throw new IllegalArgumentException("Could not find at least one of [OUTPUT_BUCKET,"
-            + " OUTPUT_TOPIC, OUTPUT_TABLE] in environment");
+        return OutputType.bigQueryLoad;
```
jklukas: Why are we defaulting to bigQueryLoad now?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
relud: confirmed spotless doesn't remove an extra line between enum values, will add.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
relud: Yeah, the default value wasn't chosen based on the limit, it was just copied from the gcs output. I'll add a justification of the number, which we can update

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
relud: I've been using the decimal values because it leaves room for disagreements about what constitutes size, but I also think it's easier to read than the bit-shift version.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
relud: `bigQueryLoad` is the one that consumes notifications from `bigQueryFiles`, and we default to it because unlike the rest of the modes it doesn't have any required output options.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
relud: I've moved all of the defaults to constants, because almost all of them were used at least twice.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -20,21 +20,76 @@
...
+  private static final int DEFAULT_LOAD_MAX_FILES = 10_000; // 10,000
+  // BigQuery Load API limits maximum load requests per table per day to 1,000, so with
+  // three instances for redundancy that means each instance should load at most once
+  // every 259.2 seconds or approximately 4.3 minutes. Messages not delivered via streaming are
+  // expected to have a total pipeline delay (including edge and decoder) of less than 1 hour, and
+  // ideally less than 10 minutes, so this plus DEFAULT_BATCH_MAX_DELAY should be about 10 minutes.
+  private static final String DEFAULT_LOAD_MAX_DELAY = "9m"; // 9 minutes
+  // BigQuery Load API limits maximum load requests per table per day to 1,000 but mixed mode
+  // expects fewer than 1,000 messages per day to need file loads, so use streaming max delay.
+  private static final String DEFAULT_STREAMING_LOAD_MAX_DELAY = DEFAULT_STREAMING_BATCH_MAX_DELAY;
```
relud: moved defaults here, re-r? because I also updated the docs while I was at it.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
jklukas: Can you add a comment to that effect?

PR #1021, opened 2019-12-06
 Bug 1601139 - Update ingestion-beam integration workflow with sampled documents

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1601139)

This updates the integration testing workflow to utilize the `monitoring.document_sample_nonprod_v1` dataset instead of `sanitized_landfill_sample_v3` added in mozilla/bigquery-etl#570.

There are about 42k documents, taking up 600MB on disk.


ingestion-beam/bin/download-document-sample
```diff
@@ -0,0 +1,87 @@
...
+SELECT
+  STRUCT( document_namespace,
+    document_type,
+    document_version ) AS attributeMap,
+  payload
+FROM
+  `moz-fx-data-shared-prod`.monitoring.document_sample_nonprod_v1
+WHERE
+  document_decoded
+  AND submission_timestamp = (
```
jklukas: This looks like it's only going to return a single row. Is that what's intended?

ingestion-beam/bin/download-document-sample
```diff
@@ -0,0 +1,87 @@
...
+  SELECT
+    MAX(submission_timestamp)
+  FROM
+    `moz-fx-data-shared-prod`.monitoring.document_sample_nonprod_v1 )
```
jklukas: I'd prefer to avoid subqueries and instead factor this out to a named CTE in a WITH statement to improve readability.

ingestion-beam/bin/download-document-sample
```diff
@@ -0,0 +1,87 @@
...
+ORDER BY
+  document_namespace,
+  document_type,
+  document_version
+"""
+
+
+def extract_samples(query):
+    """A generator for reading documents in the sampled landfill dataset that exist in the schemas.
```
jklukas: We should probably avoid the "landfill" jargon now. We've been tending to call this data "payload_bytes" since that's how the tables are named.

acmiyaguchi: Sorry, this was missed during the rewrite of `download-sampled-landfill`, I've tried to be consistent with `payload_bytes_(decoded|error)` and "document sample" wording.

ingestion-beam/bin/download-document-sample
acmiyaguchi: Each time the query is run, the result set is given a unique timestamp. So this is returning the last set of documents that have been sampled.

I also went through the entire workflow to enumerate the tables, and I've validated that the total number of documents in the dataset matches what I get locally.


```sql
select split(_TABLE_SUFFIX, '__')[offset(0)] as document_namespace, count(*)
from `amiyaguchi-dev.test_ingestion.*`
group by 1 order by 2
```

Row | document_namespace | f0_ |  
-- | -- | -- | --
1 | eng_workflow | 59 |  
2 | firefox_installer | 1000 |  
3 | mobile | 1000 |  
4 | firefox_launcher_process | 1000 |  
5 | pocket | 1000 |  
6 | org_mozilla_tv_firefox | 1011 |  
7 | org_mozilla_fenix_nightly | 1051 |  
8 | activity_stream | 3000 |  
9 | messaging_system | 3374 |  
10 | org_mozilla_fenix | 5000 |  
11 | telemetry | 24565 |  

```bash
jq '.attributeMap.document_namespace' document_sample.ndjson | uniq -c       
3000 "activity-stream"
  59 "eng-workflow"
1000 "firefox-installer"
1000 "firefox-launcher-process"
3374 "messaging-system"
1000 "mobile"
5000 "org-mozilla-fenix"
1051 "org-mozilla-fenix-nightly"
1011 "org-mozilla-tv-firefox"
1000 "pocket"
24565 "telemetry"
```

Issue #957, opened 2019-10-29
Normalize destination table case in ingestion-sink


whd:
[Example](https://console.cloud.google.com/errors/CKOT1uiwneC2rAE?time=P1D&project=moz-fx-data-inge-nonprod-9cc2).

PR #1028, opened 2019-12-12
Normalize destionation table case in ingestion-sink

jklukas:
Closes https://github.com/mozilla/gcp-ingestion/issues/957


ingestion-beam/src/main/java/com/mozilla/telemetry/schemas/SchemaStore.java
```diff
@@ -9,7 +9,7 @@
 import com.google.common.collect.Multimap;
 import com.google.common.util.concurrent.UncheckedExecutionException;
 import com.mozilla.telemetry.transforms.MapElementsWithErrors.BubbleUpException;
-import com.mozilla.telemetry.util.SnakeCase;
+import com.mozilla.telemetry.ingestion.core.util.SnakeCase;
```
jklukas: We move SnakeCase to the core module to enable this PR.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -95,7 +96,8 @@ long getDefaultMaxOutstandingRequestBytes() {
             // BigQuery Streaming API Limits maximum rows per request to 10,000
             env.getInt(BATCH_MAX_MESSAGES, 10_000), // default 10K messages
             env.getDuration(BATCH_MAX_DELAY, "1s"), // default 1 second
-            env.getString(OUTPUT_TABLE), getFormat(env));
+            PubsubMessageToTemplatedString.forBigQuery(env.getString(OUTPUT_TABLE)),
```
jklukas: We change a few signatures to expect a `PubsubMessageToTemplatedString` rather than a bare String, so that we can control whether special BigQuery-specific transformations need to be applied. We provide two different static factory methods.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToTemplatedString.java
```diff
@@ -17,18 +26,77 @@
...
+    return new PubsubMessageToTemplatedString(template, DestinationType.BIGQUERY);
   }
 
   @Override
   public String apply(PubsubMessage message) {
-    String batchKey = StringSubstitutor.replace(template,
-        DerivedAttributesMap.of(message.getAttributesMap()));
+    final String batchKey;
+    if (destinationType == DestinationType.BIGQUERY) {
+      batchKey = applyForBigQuery(message);
```
jklukas: Most of the new logic here is in this applyForBigQuery code path.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToTemplatedString.java
```diff
@@ -17,18 +26,77 @@
...
+    }
+    if (docType != null) {
+      attributes.put(Attribute.DOCUMENT_TYPE, getAndCacheNormalizedName(docType));
+    }
+
+    attributes = Maps.transformValues(DerivedAttributesMap.of(attributes),
+        v -> v.replaceAll("-", "_"));
+
+    return StringSubstitutor.replace(template, attributes);
+  }
```
jklukas: This logic is almost exactly copied from KeyByBigQueryTableDestination. We will need a bigger refactor in order to support writing into live tables in ingestion-sink, and that would be the point at which to consolidate logic and remove this duplication.

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/BigQueryTest.java
```diff
@@ -147,21 +149,35 @@ public void canHandleDynamicTableIdWithEmptyValues() {
...
+  public void canHandleDynamicTableIdWithHyphens() {
+    output = new BigQuery.Write(bigQuery, MAX_BYTES, MAX_MESSAGES, NO_DELAY,
+        PubsubMessageToTemplatedString
+            .forBigQuery("${document_namespace}.${document_type}_${suffix}"),
+        Format.raw);
+    output.apply(PubsubMessage.newBuilder().putAttributes("document_namespace", "my-namespace")
+        .putAttributes("document_type", "myDocType").putAttributes("suffix", "my-suffix").build())
+        .join();
+    assertNotNull(output.batches.get(TableId.of("my_namespace", "my_doc_type_my_suffix")));
+  }
```
jklukas: This is the test that exercises the template transformation logic.

PR #1042, opened 2019-12-19
 Bug 1604910  - Publish a docker image for ingestion-beam to GCR

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1604910)

### DIFF
--- a/ingestion-beam/Dockerfile
+++ b/ingestion-beam/Dockerfile
@@ -1,0 +1,22 @@
+FROM maven:3-jdk-8 AS base
+WORKDIR /app/ingestion-beam
+
+COPY LICENSE /app/LICENSE
+COPY pom.xml /app/pom.xml
+COPY ingestion-beam/pom.xml /app/ingestion-beam/pom.xml
+COPY checkstyle /app/checkstyle
+COPY ingestion-beam/checkstyle /app/ingestion-beam/checkstyle
+COPY ingestion-core/src /app/ingestion-core/src
+COPY ingestion-beam/src /app/ingestion-beam/src
+COPY ingestion-beam/bin /app/ingestion-beam/bin
+
+FROM base AS build
+RUN /app/ingestion-beam/bin/download-cities15000
+RUN /app/ingestion-beam/bin/download-geolite2
+RUN /app/ingestion-beam/bin/download-schemas
+RUN mvn package
+
+FROM base
+COPY --from=build /app/ingestion-beam/target/*.jar /app/ingestion-beam/target/
+RUN apt-get update && apt-get install -y python3 python3-pip jq
+RUN pip3 install google-cloud-bigquery
--- a/ingestion-beam/bin/download-document-sample
+++ b/ingestion-beam/bin/download-document-sample
@@ -73,8 +73,9 @@ def main(args):
     start = time.time()
     with open(args.output_file, "w") as fp:
         for pubsub_document in extract_samples(QUERY):
-            fp.write(f"{json.dumps(pubsub_document)}\n")
-    logging.info(f"Done in {time.time()-start} seconds!")
+            fp.write(json.dumps(pubsub_document))
+            fp.write("\n")
+    logging.info("Done in {} seconds!".format(time.time()-start))
 
 
 def parse_arguments():


ingestion-beam/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+WORKDIR /app/ingestion-beam
+
+COPY LICENSE /app/LICENSE
+COPY pom.xml /app/pom.xml
+COPY ingestion-beam/pom.xml /app/ingestion-beam/pom.xml
+COPY checkstyle /app/checkstyle
+COPY ingestion-beam/checkstyle /app/ingestion-beam/checkstyle
+COPY ingestion-core/src /app/ingestion-core/src
+COPY ingestion-beam/src /app/ingestion-beam/src
+COPY ingestion-beam/bin /app/ingestion-beam/bin
```
jklukas: It may be more efficient to implement a .dockerignore file to handle excluding the content we don't need before it even gets transferred to the Docker context. Then we could simplify these COPY statements.

acmiyaguchi: I'll think about this as an optimization in a second pass. I'm working through getting the integration workflow running inside of the container, but this would be desirable. There's >1GB of context that's sent to docker everytime I build, likely leftovers from development that I've done. I do like this pattern of being explicit about the dependencies, but it's pretty verbose too.

acmiyaguchi:
## Notes

```
docker built -t ingestion-beam:latest -f ingestion-beam/Dockerfile .
cd ingestion-beam

export GOOGLE_APPLICATION_CREDENTIALS=...

docker run \
	-v $GOOGLE_APPLICATION_CREDENTIALS:/tmp/credentials \
	-e GOOGLE_APPLICATION_CREDENTIALS=/tmp/credentials \
	-v $(pwd)/GeoLite2-City.mmdb:/app/ingestion-beam/GeoLite2-City.mmdb \
	-v $(pwd)/cities15000:/app/ingestion-beam/cities15000: \
	-v $(pwd)/document_sample.ndjson:/app/ingestion-beam/document_sample.ndjson \
	-it ingestion-beam:latest \
	bash

# should these be included in the image?
bin/download-schemas
bin/download-geolite2
bin/download-cities15000

# this should certainly not be included in the image
bin/download-document-sample

# heap needs to be increased, as well as sharding due to OOM
MAVEN_OPTS="-Xmx4g" \
    --inputFileFormat=json \
    --inputType=file \
    --input=/app/ingestion-beam/document_sample.ndjson \
    --outputFileFormat=json \
    --outputType=file \
    --output=/tmp/output/out \
    --errorOutputType=file \
    --errorOutput=/tmp/error \
    --outputNumShards=10
"
# this may take a while, can this be optimized?
```

acmiyaguchi:
I'm going to close this PR, due to the poor performance on the document sample. When running within the docker container, the job does not finish within 20 minutes. When running via Dataflow, the job takes 5 minutes:

```
    --runner=DataflowRunner \
    --inputFileFormat=json \
    --inputType=file \
    --input=gs://amiyaguchi-dev/document_sample.ndjson \
    --outputFileFormat=json \
    --outputType=file \
    --output=gs://amiyaguchi-dev/tmp/output/out \
    --errorOutputType=file \
    --errorOutput=gs://amiyaguchi-dev/tmp/error \
"
```

This is for the sink, which does not do any validation of the data. When running the decoder on dataflow, the job scales up to 15 workers over the course of 20 minutes and processes only 3 or 4 documents.

https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2019-12-31_11_27_19-5855416270605556042?project=amiyaguchi-dev

```
mvn compile exec:java -Dexec.mainClass=com.mozilla.telemetry.Decoder -Dexec.args="\
    --runner=DataflowRunner \
    --geoCityDatabase=src/test/resources/cityDB/GeoIP-City-Test.mmdb \
    --schemasLocation=schemas.tar.gz \
    --inputFileFormat=json \
    --inputType=file \
    --input=gs://amiyaguchi-dev/document_sample.ndjson \
    --outputFileFormat=json \
    --outputType=file \
    --output=gs://amiyaguchi-dev/tmp/output/out \
    --errorOutputType=file \
    --errorOutput=gs://amiyaguchi-dev/tmp/error \
"
```

Note that the test db is used here as a stub -- IP address are not included in the document sample.

What will probably need to be done is to write a custom script to do validation instead of relying on this pipeline. I will likely be toying with a json schema 6/7 compliant validator in python.

PR #1061, opened 2020-01-07
Copy GeoIP2City resource to disk to allow memory mapping

jklukas:
This should reduce memory footprint in all cases, but is particularly relevant
for usage with Dataflow's streaming engine as discussed in
https://bugzilla.mozilla.org/show_bug.cgi?id=1598387

Dataflow generally spawns one worker thread per vCPU and this change also
makes it possible for all worker threads to share a single DatabaseReader
object so we don't duplicate the memory footprint.


jklukas:
This PR is attempting to use the singleton design pattern, but there's plenty of room for nuance and subtle bugs in that space, so I'd appreciate thorough review here with caps on for Java expertise and Beam programming model expertise.

The general approach here is that we have a static field on `GeoCityLookup` for the database reader. The pipeline gets inititalized, at which point the `GeoCityLookup` constructor is called and the static field is set to null. The `GeoCityLookup` instance gets serialized and sent to workers, who then start processing data. On each worker, the first thread to process a record will end up calling the static synchronized `getOrCreateDatabaseReader` method for fetching the GeoCity db from GCS; while that is happening, other worker threads will block on that same method call. Once the first thread finishes that call, the DB reader is initialized and all other threads will be passed the same instance. Each worker thread ends up with its own reference to the singleton and they happily go on their way.

jklukas:
I've kicked off a Dataflow batch job running from this branch, and it's successfully processing records. So at least from a high level, this does not obviously break the decoder.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -173,47 +257,10 @@ private void loadResourcesOnFirstMessage() throws IOException {
       if (geoCityDatabase == null || !geoCityDatabase.isAccessible()) {
         throw new IllegalArgumentException("--geoCityDatabase must be defined for GeoCityLookup!");
       }
-      loadGeoIp2City();
+      geoIP2City = GeoCityLookup.getOrCreateSingletonGeoCityReader(geoCityDatabase);
```
relud: Pretty sure the class name isn't necessary here:
```suggestion
      geoIP2City = getOrCreateSingletonGeoCityReader(geoCityDatabase);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -173,47 +257,10 @@ private void loadResourcesOnFirstMessage() throws IOException {
...
-        Matcher matcher = GEO_NAME_PATTERN.matcher(line);
-        if (matcher.find()) {
-          Integer geoNameId = Integer.valueOf(matcher.group(1));
-          allowedCities.add(geoNameId);
-        } else {
-          throw new IllegalStateException(
-              "Line of geoCityFilter file does not begin with a geoName integer ID: " + line);
-
-        }
+        allowedCities = GeoCityLookup.getOrCreateSingletonAllowedCities(geoCityFilter);
```
relud: nor here:
```suggestion
        allowedCities = getOrCreateSingletonAllowedCities(geoCityFilter);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
jklukas: > Pretty sure the class name isn't necessary here:

You are correct. I intentionally included the class name to be explicit about it being a static method, but I can remove if that seems overly verbose. Having `singleton` in the name is probably sufficient to alert future devs to what's going on.

PR #1063, opened 2020-01-08
Add a BoundedSink util for sink integration tests

relud:
And move GCP resources into JUnit rules so they are cleaned up even when
tests fail

I broke this out of my work to write tests for #1013 because I didn't want to do too much at once there.


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -35,19 +36,42 @@
       OUTPUT_TABLE, OUTPUT_TOPIC, OUTPUT_TOPIC_EXECUTOR_THREADS, MAX_OUTSTANDING_ELEMENT_COUNT,
       MAX_OUTSTANDING_REQUEST_BYTES);
 
+  @VisibleForTesting
+  protected static class Output implements Function<PubsubMessage, CompletableFuture<Void>> {
```
relud: this was added to simplify separating `getOutput` from `getInput` so that `BoundedSink` could inject a termination condition.

Issue #1005, opened 2019-11-22
Lowercase UUIDs in the decoder


jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1598536 for a case where we had a regression in ETL that removed lowercasing of client_id and caused KPI reporting to jump up due to some Fennec iOS clients being counted twice (Fennec iOS reports uppercase client_ids).

It seems reasonable to normalize docId and clientId at the decoder step and we'd get a bit of extra stability downstream.

jklukas:
This would need to happen in [ParsePayload](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java)

jklukas:
Reopening for the additional scope discussed in https://github.com/mozilla/gcp-ingestion/pull/1064#pullrequestreview-340633909

PR #1064, opened 2020-01-09
Lowercase UUIDs in the decoder

scholtzan:
This will return lowercased UUIDs when using `normalizeUUID`. 
Resolves https://github.com/mozilla/gcp-ingestion/issues/1005


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -235,7 +235,8 @@ long calculateSampleId(String clientId) {
     return crc32.getValue() % 100;
   }
 
-  private static String normalizeUuid(String v) {
+  @VisibleForTesting
+  static String normalizeUuid(String v) {
```
scholtzan: I made this public to be able to write unit tests for it. But I'm not sure if there might be a good reason for it to stay `private`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -244,7 +245,7 @@ private static String normalizeUuid(String v) {
     try {
       // Will raise an exception if not a valid UUID.
       UUID.fromString(v);
-      return v;
+      return v.toLowerCase();
```
scholtzan: This will also affect `impression_id`s. But I guess UUIDs should be lowercase in general

jklukas: Yes, I have no objection to lowercasing all UUIDs that pass through this.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
jklukas: I'm fine with this. You've marked it `@VisibleForTesting` which is perfectly in line with established practice in this codebase.

scholtzan:
Alright, I see. I'll open a separate PR for this.

PR #1069, opened 2020-01-10
Lowercase normalized UUIDs in JSON payload

scholtzan:
This addresses https://github.com/mozilla/gcp-ingestion/issues/1005 and https://github.com/mozilla/gcp-ingestion/pull/1064#pullrequestreview-340633909

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -208,15 +208,21 @@ public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<Pubs
           .ifPresent(v -> attributes.put(Attribute.OS_VERSION, v));
     }
 
-    if (Objects.isNull(attributes.get(Attribute.CLIENT_ID))) {
-      // Try extracting variants of top-level client id and normalize UUID in JSON payload.
-      Stream.of(Attribute.CLIENT_ID, "clientId") //
-          .filter(v -> Objects.nonNull(ParsePayload.normalizeUuid(json.path(v).textValue()))) //
-          .findFirst() //
+    if (attributes.get(Attribute.CLIENT_ID) == null) {
+      Optional.ofNullable(json.path(Attribute.CLIENT_ID).textValue()) //
+          .map(ParsePayload::normalizeUuid) //
           .ifPresent(v -> {
-            final String normalizedUuid = ParsePayload.normalizeUuid(json.path(v).textValue());
-            json.put(v, normalizedUuid);
-            attributes.put(Attribute.CLIENT_ID, normalizedUuid);
+            attributes.put(Attribute.CLIENT_ID, v);
+            json.put(Attribute.CLIENT_ID, v);
+          });
+    }
+
+    if (attributes.get(Attribute.CLIENT_ID) == null) {
+      Optional.ofNullable(json.path("clientId").textValue()) //
+          .map(ParsePayload::normalizeUuid) //
+          .ifPresent(v -> {
+            attributes.put(Attribute.CLIENT_ID, v);
+            json.put("clientId", v);
           });
     }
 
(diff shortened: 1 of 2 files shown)


ingestion-beam/pom.xml
```diff
@@ -138,6 +138,12 @@
             <artifactId>slf4j-jdk14</artifactId>
             <version>1.7.30</version>
         </dependency>
+        <dependency>
+            <groupId>com.mozilla.telemetry</groupId>
+            <artifactId>ingestion-core</artifactId>
+            <version>0.1-SNAPSHOT</version>
+            <scope>compile</scope>
+        </dependency>
```
scholtzan: I had to add this to make it compile in IntelliJ. I can also git ignore this if it should be removed.

scholtzan: Okay, needs to be removed otherwise CircleCI cannot compile

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -204,14 +208,17 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
-    Stream
-        .of(attributes.get(Attribute.CLIENT_ID), json.path(Attribute.CLIENT_ID).textValue(),
-            json.path("clientId").textValue())
-        .map(ParsePayload::normalizeUuid) //
-        .filter(Objects::nonNull) //
-        .findFirst() //
-        .ifPresent(v -> attributes.put(Attribute.CLIENT_ID, v));
+    if (Objects.isNull(attributes.get(Attribute.CLIENT_ID))) {
+      // Try extracting variants of top-level client id and normalize UUID in JSON payload.
+      Stream.of(Attribute.CLIENT_ID, "clientId") //
```
scholtzan: This works for now since the JSON structure is not nested. To update the json payload the field names are required. However, if in the future the structure gets more complex/nested then this needs to be changed.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -204,14 +208,17 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
+          .filter(v -> Objects.nonNull(ParsePayload.normalizeUuid(json.path(v).textValue()))) //
+          .findFirst() //
+          .ifPresent(v -> {
+            final String normalizedUuid = ParsePayload.normalizeUuid(json.path(v).textValue());
+            json.put(v, normalizedUuid);
+            attributes.put(Attribute.CLIENT_ID, normalizedUuid);
+          });
+    }
```
jklukas: I think this block has gotten complex enough that it;s hard to parse through what's going on in the set of stream operations. And we have to call `ParsePayload.normalizeUuid(json.path(v).textValue())` twice on the same data.

I think I'd prefer that we have separate blocks at this point for CLIENT_ID and clientId. Or maybe it's worth  factoring out a function that accepts an `ObjectNode` and a string for the target key within it.

I am also okay if you want to make the judgment call that what's already here is the best option.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -204,14 +208,17 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
 
-    // Try extracting variants of top-level client id.
-    Stream
-        .of(attributes.get(Attribute.CLIENT_ID), json.path(Attribute.CLIENT_ID).textValue(),
-            json.path("clientId").textValue())
-        .map(ParsePayload::normalizeUuid) //
-        .filter(Objects::nonNull) //
-        .findFirst() //
-        .ifPresent(v -> attributes.put(Attribute.CLIENT_ID, v));
+    if (Objects.isNull(attributes.get(Attribute.CLIENT_ID))) {
```
jklukas:
```suggestion
    if (attributes.get(Attribute.CLIENT_ID) == null) {
```

Is there a reason to use Objects.isNull here vs. the simpler `== null`? I think `isNull` and `nonNull` are primarily useful because they can be used as method references in Stream computations.

I'd also replace the `Objects.nonNull` call three lines down with `!= null` since it's no longer a simple method reference.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
scholtzan: I agree, having them in separate blocks makes it easier to see what's going on.

jklukas:
This should be good to merge now.

Issue #668, opened 2019-06-20
Consider adding a document id derived from pubsub message id for pings without a document id


whd:
The legacy systems we ingest from (sslreports, stub installer) don't necessarily provide a document id. We could add an arbitrary one in the decoder but the pubsub message id (or in ingestion edge) would provide a more stable value since it will be unique and persisted to durable storage in case of reprocessing. We want the document id to be a UUID; presumably there is some casting that would be required for this.

jklukas:
One of the recent versions of Beam added messageId to the PubSubMessage class and we had to avoid serializing it in order to keep our tests from breaking.

But I'm pretty sure we'll still have the messageId available when reading from PubSub. Anyways, this sounds doable. We might just have to be careful about where we put it and we'll likely have to refactor some tests.

jklukas:
See our [`PubsubMessageMixin`](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java#L221-L234) to see how we avoid serializing `messageId`. We'd need to remove the alteration of `messageId` or somehow make that optional for tests. We might need to add a transform for optionally nulling messageIds and add a pipeline option to make that configurable.

jklukas:
See [StubUri](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java#L101-L103) in ParsePayload for an example of handling a legacy system. Maybe we add handling there for converting a pubsub messageId into a docId and explicitly drop messageId there. I suppose we already do drop messageId considering we return a value where we use the `PubsubMessage` constructor without a messageId parameter.

PR #1074, opened 2020-01-15
Document id derived from pubsub message id

scholtzan:
I started looking into https://github.com/mozilla/gcp-ingestion/issues/668

I'm not sure if I'm on the right track. Can it be assumed the PubSub messages always have a message ID?

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
@@ -9,11 +9,8 @@ import com.mozilla.telemetry.transforms.MapElementsWithErrors;
 import com.mozilla.telemetry.transforms.PubsubConstraints;
 import com.mozilla.telemetry.util.Json;
 import java.io.IOException;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.nio.charset.StandardCharsets;
+import java.util.*;
 import java.util.function.BiConsumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -77,6 +74,8 @@ public class ParseUri extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMe
   @Override
   protected PubsubMessage processElement(PubsubMessage message)
       throws InvalidUriException, IOException {
+    System.out.println("messageID 1: " + message.getMessageId());
+
     message = PubsubConstraints.ensureNonNull(message);
     // Copy attributes
     final Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
@@ -100,10 +99,24 @@ public class ParseUri extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMe
           zip(GENERIC_URI_SUFFIX_ELEMENTS, uri.substring(GENERIC_URI_PREFIX.length()).split("/")));
     } else if (uri.startsWith(StubUri.PREFIX)) {
       payload = StubUri.parse(uri, attributes);
+
+      System.out.println("message " + message.getAttributeMap());
+      System.out.println("message " + Arrays.toString(message.getPayload()));
+
+      System.out.println("messageID: " + message.getMessageId());
+      if (message.getMessageId() != null) {
+        // convert PubSub message ID to document ID
+        UUID documentId = UUID
+            .nameUUIDFromBytes(message.getMessageId().getBytes(StandardCharsets.UTF_8));
+        attributes.put(Attribute.DOCUMENT_ID, documentId.toString().toLowerCase());
+      }
+
+      System.out.println("attributes: " + attributes);
+
     } else {
       throw new InvalidUriException("Unknown URI prefix");
     }
-    return new PubsubMessage(payload, attributes);
+    return new PubsubMessage(payload, attributes, message.getMessageId());
   }
 
   /**
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
@@ -10,7 +10,12 @@ import com.mozilla.telemetry.transforms.PubsubConstraints;
 import com.mozilla.telemetry.util.Json;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
-import java.util.*;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.UUID;
 import java.util.function.BiConsumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -74,8 +79,6 @@ public class ParseUri extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMe
   @Override
   protected PubsubMessage processElement(PubsubMessage message)
       throws InvalidUriException, IOException {
-    System.out.println("messageID 1: " + message.getMessageId());
-
     message = PubsubConstraints.ensureNonNull(message);
     // Copy attributes
     final Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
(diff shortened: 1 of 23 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
```diff
@@ -38,27 +38,25 @@
   @Test
   public void testOutput() {
     final List<String> validInput = Arrays.asList(//
-        "{\"attributeMap\":{},\"payload\":\"\"}", //
-        "{\"attributeMap\":" //
+        "{\"attributeMap\":{},\"payload\":\"\",\"messageId\":\"0000000001\"}", "{\"attributeMap\":" //
             + "{\"uri\":\"/submit/telemetry/ce39b608-f595-4c69-b6a6-f7a436604648"
             + "/main/Firefox/61.0a1/nightly/20180328030202\"" //
-            + "},\"payload\":\"\"}",
+            + "},\"payload\":\"\",\"messageId\":\"0000000001\"}",
```
scholtzan: Right now for this to the pubsub message ID is required. Otherwise if the field is missing or `null`: `org.apache.beam.sdk.Pipeline$PipelineExecutionException: java.lang.RuntimeException: org.apache.beam.sdk.coders.CoderException: cannot encode a null String`

jklukas:
I had to read through some of the existing code again to remember relevant details.

First, PubsubIO has a few different methods for reading. We use `readMessagesWithAttributes` but we'll need to switch to `readMessagesWithAttributesAndMessageId`.

Next, there are multiple `Coder` implementations available for `PubsubMessage`. In several places in our code, we specify `PubsubMessageWithAttributesCoder`, but there is also a `PubsubMessageWithAttributesAndMessageIdCoder` that we'll need to switch to. IIRC, Beam is not able to infer a default coder for PubsubMessage, so we always have to explicitly define a Coder for each transform that interacts with PCollections of PubsubMessage.

Next, we'll need to remove the `@JsonIgnore` annotation on our PubsubMixin if we want to be able to read in messageIds from JSON files in our tests.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
jklukas: I don't think there's going to be a way around this (it's true that any value that Beam is going to encode/decode cannot be null), which is the potentially really tricky part of this change. It may require updating all of our test cases to use messageIds, which would be onerous.

jklukas:
Here's a potential way forward: we could try to isolate the change to the smallest section of the code possible. We could update just our PubsubInput implementation to use `readMessagesWithAttributesAndMessageId` and immediately strip the `messageId` to an attribute `message_id` in a `.apply(MapElements...)` call within `PubsubInput#expand`.

That particular section of code might be impossible to unit test, but we're basically relying on  PubsubIO to do what it advertises. Then, we'd be able to write tests where the inputs have a `message_id` attribute in the `attributeMap`.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -223,12 +219,11 @@ public static String asString(Object data) throws IOException {
 
     @JsonCreator
     public PubsubMessageMixin(@JsonProperty("payload") byte[] payload,
-        @JsonProperty("attributeMap") Map<String, String> attributes) {
+        @JsonProperty("attributeMap") Map<String, String> attributes,
+        @JsonProperty("messageId") String messageId) {
     }
 
     @JsonIgnore
```
scholtzan: I do think that `@JsonIgnore` is necessary here otherwise when serializing `PubsubMessage` it'll be part of the resulting json

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -100,9 +102,18 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+        // convert PubSub message ID to document ID
+        UUID documentId = UUID
+            .nameUUIDFromBytes(attributes.get("messageId").getBytes(StandardCharsets.UTF_8));
+        attributes.put(Attribute.DOCUMENT_ID, documentId.toString().toLowerCase());
+      }
     } else {
       throw new InvalidUriException("Unknown URI prefix");
     }
+
+    attributes.remove("messageId");
```
scholtzan: I guess we'll probably want to remove messageId from the attributes?

jklukas: That's probably a good idea, though worth adding a comment here about the fact that we expect messageId to be relevant only for coverting to docId, and thus isn't useful past this point.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -100,9 +102,18 @@ protected PubsubMessage processElement(PubsubMessage message)
           zip(GENERIC_URI_SUFFIX_ELEMENTS, uri.substring(GENERIC_URI_PREFIX.length()).split("/")));
     } else if (uri.startsWith(StubUri.PREFIX)) {
       payload = StubUri.parse(uri, attributes);
+
+      if (attributes.get("messageId") != null) {
```
jklukas: We've been using snake_case for attributes, so I'd prefer we use `"message_id"`. Also, can you factor this string out to an `Attributes.MESSAGE_ID` constant?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -100,9 +102,18 @@ protected PubsubMessage processElement(PubsubMessage message)
...
+        // convert PubSub message ID to document ID
+        UUID documentId = UUID
+            .nameUUIDFromBytes(attributes.get("messageId").getBytes(StandardCharsets.UTF_8));
```
jklukas: This is the first time I've encountered `nameUUIDFromBytes` and it looks like there's some need to document the specific approach here. The RFC for UUIDs discusses creating a "name space" by using a static UUID. It looks like the `nameUUIDFromBytes` expects you to prepend that namespace to the bytes you pass in; otherwise, you're assuming a null namespace. See https://stackoverflow.com/a/55296637

I think it's probably fine to use the call as written here, but we should document that it's a V3 UUID using a null namespace so that it's more obvious how you would implement the same calculation in a different language if needed.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -46,8 +48,15 @@ public PubsubInput(ValueProvider<String> subscription) {
     @Override
     public Result<PCollection<PubsubMessage>> expand(PBegin input) {
       return input //
-          .apply(PubsubIO.readMessagesWithAttributes().fromSubscription(subscription))
-          .apply(ToPubsubMessageFrom.identity());
+          .apply(PubsubIO.readMessagesWithAttributesAndMessageId().fromSubscription(subscription))
+          .apply(MapElements.via(new SimpleFunction<PubsubMessage, PubsubMessage>() {
```
jklukas: We've generally been preferring to use lambdas where possible rather than anonymous subclasses, so this would be:

```
MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via(message -> {
  ...
})
```

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -46,8 +48,15 @@ public PubsubInput(ValueProvider<String> subscription) {
...
+
+            public PubsubMessage apply(PubsubMessage input) {
```
jklukas: `input` is the variable name for the input to the outer `expand` method, so it's best to avoid masking that here.

scholtzan:
Should be ready for another look

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -46,8 +48,12 @@ public PubsubInput(ValueProvider<String> subscription) {
     @Override
     public Result<PCollection<PubsubMessage>> expand(PBegin input) {
       return input //
-          .apply(PubsubIO.readMessagesWithAttributes().fromSubscription(subscription))
-          .apply(ToPubsubMessageFrom.identity());
+          .apply(PubsubIO.readMessagesWithAttributesAndMessageId().fromSubscription(subscription))
+          .apply(MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via(message -> {
+            Map<String, String> attributesWithMessageId = message.getAttributeMap();
+            attributesWithMessageId.put(Attribute.MESSAGE_ID, message.getMessageId());
```
jklukas: This line is mutating the `attributeMap` from the input PubsubMessage. We are in general not guaranteed that the Map implementation used in a PubsubMessage is mutable, so I would normally recommend that we populate `new HashMap<>(message.getAttributeMap())`. In practice, this transform is happening directly after `PubsubIO.readMessagesWithAttributesAndMessageId()` which returns a mutable map, so probably fine.

If we want to keep this as a mutation to save allocating a new map, it would be simpler and more explicit to call `message.getAttributeMap().put(...)` and return `message` rather than constructing a new one. Otherwise, let's copy the map.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -204,17 +205,12 @@ public static String asString(Object data) throws IOException {
    * {@code getAttributeMap} method returns the value for the {@code attributes} parameter.
    * Additionally jackson doesn't like that there are no setter methods on {@link PubsubMessage}.
    *
-   * <p>Beam 2.16 added the {@code messageId} field, which is not relevant for our purposes;
```
jklukas: I think we should still have some commentary here about why we mark the field as `@JsonIgnore`.

Am I correct that with the changes here, we are able to read messageId when reading JSON input, but we won't output messageId when serializing to JSON?

scholtzan: Yes, that is correct

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
scholtzan: Okay it seems this is causing some problems: https://circleci.com/gh/mozilla/gcp-ingestion/18849?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

Will create a PR to fix this.

PR #1081, opened 2020-01-24
Remove some old beam code and add GRAVEYARD.md

jklukas:


GRAVEYARD.md
```diff
@@ -0,0 +1,5 @@
+# Link to code removed from this repository
```
wlach: This file is a great idea, should we add updating this (if applicable) to a pull request checklist? (can be done in a seperate PR)

jklukas: Graveyarding code doesn't happen all that often. I worry that putting this in a PR checklist will just get people in the habit of ignoring the checklist. I welcome more discussion on this point, though. Relying on developer discipline is definitely not a great solution either.

Issue #118, opened 2018-10-23
Use a standard failure output transform


relud:
To match [the upstream patch @jklukas did for failure handling](https://github.com/apache/beam/pull/6586), I think using it should look like:

```java
FailureOutput.withSuccessTag(success).withFailureTag(tag, output).withFailureTag(...)
```

or perhaps instead of framing this around failures, frame it as side outputs, e.g.

```java
SideOutputs.withOutput(output, tag, ...).withOutput(...).getLastOutput()
```

jklukas:
This is blocked for now as discussion stalled on the upstream patch. I'm considering some options for moving that effort forward.

jklukas:
A failure handling interface has been accepted and merged in https://github.com/apache/beam/pull/7736 and should be available in Beam 2.12 to be released in a few weeks, so we'll eventually be unblocked on this refactor.

jklukas:
2.12 is released, so we can pursue this now.

scholtzan:
I'm trying to think this through. So is the idea here to change [MapElementsWithErrors](https://github.com/mozilla/gcp-ingestion/blob/a8040f945e0ff386f119dfefa6f4b71326d87a1f/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/MapElementsWithErrors.java) to instead use [MapElementsWithFailures](https://beam.apache.org/releases/javadoc/2.12.0/org/apache/beam/sdk/transforms/MapElements.MapWithFailures.html)? So that it would not be necessary anymore to create custom [FailureMessages](https://github.com/mozilla/gcp-ingestion/blob/a8040f945e0ff386f119dfefa6f4b71326d87a1f/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/FailureMessage.java) and process those?

jklukas:
> I'm trying to think this through. So is the idea here to change [MapElementsWithErrors](https://github.com/mozilla/gcp-ingestion/blob/a8040f945e0ff386f119dfefa6f4b71326d87a1f/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/MapElementsWithErrors.java) to instead use [MapElementsWithFailures](https://beam.apache.org/releases/javadoc/2.12.0/org/apache/beam/sdk/transforms/MapElements.MapWithFailures.html)? 

[...]

PR #1083, opened 2020-01-25
Refactoring to use MapElements.WithFailure

scholtzan:
I started looking into https://github.com/mozilla/gcp-ingestion/issues/118

For now, this only refactors `LimitPayloadSize` to use `MapElements.WithFailures`. It's probably best to discuss on a small example whether it makes sense to go forward and refactor the rest. Refactoring the remaining classes should be pretty similar.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
@@ -46,7 +46,8 @@ public class LimitPayloadSize {
           try {
             throw ee.exception();
           } catch (PayloadTooLargeException e) {
-            return FailureMessage.of("LimitPayloadSize", ee.element(), ee.exception());
+            return FailureMessage.of(LimitPayloadSize.class.getSimpleName(), ee.element(),
+                ee.exception());
           }
         });
   }
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
@@ -3,8 +3,15 @@ package com.mozilla.telemetry.transforms;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.metrics.Counter;
 import org.apache.beam.sdk.metrics.Metrics;
+import org.apache.beam.sdk.transforms.MapElements;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.WithFailures.ExceptionElement;
+import org.apache.beam.sdk.transforms.WithFailures.Result;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TypeDescriptor;
 
-public class LimitPayloadSize extends MapElementsWithErrors.ToPubsubMessageFrom<PubsubMessage> {
+public class LimitPayloadSize extends
+    PTransform<PCollection<PubsubMessage>, Result<PCollection<PubsubMessage>, PubsubMessage>> {
 
   /** Exception to throw for messages whose payloads exceed the configured size limit. */
   public static class PayloadTooLargeException extends Exception {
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -25,15 +32,21 @@ public static LimitPayloadSize toMB(int mb) {
...
-    return message;
+  public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
+      PCollection<PubsubMessage> elements) {
+    return elements.apply(
+        MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via((PubsubMessage message) -> {
+          message = PubsubConstraints.ensureNonNull(message);
+          int numBytes = message.getPayload().length;
+          if (numBytes > maxBytes) {
+            countPayloadTooLarge.inc();
+            throw new RuntimeException(new PayloadTooLargeException("Message payload is " + numBytes
```
scholtzan: It's only possible to throw `RuntimeExceptions`. Not sure if there is a better way than to just wrap all other exception types into `RuntimeExceptions`?

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -25,15 +32,21 @@ public static LimitPayloadSize toMB(int mb) {
...
-  protected PubsubMessage processElement(PubsubMessage message) throws Exception {
-    message = PubsubConstraints.ensureNonNull(message);
-    int numBytes = message.getPayload().length;
-    if (numBytes > maxBytes) {
-      countPayloadTooLarge.inc();
-      throw new PayloadTooLargeException("Message payload is " + numBytes + "bytes, larger than the"
-          + " configured limit of " + maxBytes);
-    }
-    return message;
+  public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
```
scholtzan: Basically, for every class that currently extends `MapElementsWithErrors` the logic from `processElements` can be moved into `expands`.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -25,15 +32,21 @@ public static LimitPayloadSize toMB(int mb) {
...
+          int numBytes = message.getPayload().length;
+          if (numBytes > maxBytes) {
+            countPayloadTooLarge.inc();
+            throw new RuntimeException(new PayloadTooLargeException("Message payload is " + numBytes
+                + "bytes, larger than the" + " configured limit of " + maxBytes));
+          }
+          return message;
+        }).exceptionsInto(TypeDescriptor.of(PubsubMessage.class))
+            .exceptionsVia((ExceptionElement<PubsubMessage> ee) -> FailureMessage.of(this,
+                ee.element(), ee.exception())));
```
jklukas: We could reraise `ee.exception` here so that we can catch only the expected case (`PayloadTooLarge`), which is a potential benefit we get from this pattern vs. what we were doing in MapElementsWithFailures.

Also, we could sidestep the wrapped exception problem here by inspecting `ee.exception().getCause()`.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
jklukas: I think it's lambdas that can't have runtime exceptions. I think we could make turn this into a `new ProcessFunction {...}` and we'd be able to throw `PayloadTooLargeException` here.

I'm also uncertain as to the value we're getting from having our custom exception classes be checked, since it does make us have to deal with additional complications like this. Checked exceptions are nice in theory, though I believe much of the Java community considers checked exceptions a failed experiment. I don't have great references for that, though.

jklukas: We discussed with @relud and he supports abandoning checked exceptions, and making all of our custom exception types be runtime.

jklukas:
[An interesting discussion relevant to this refactoring came up on the dev@beam mailing list](https://lists.apache.org/thread.html/rce7569aff44f58d3ee8be7f378c0fe1cb96fb3ccb4783bc9a212373a%40%3Cdev.beam.apache.org%3E). We could refactor this class so it no longer extends PTransform, but rather returns the MapElements transform directly. It would eliminate quite a bit of boilerplate, so would be good for readability. But we lose control over how the transform is named in the pipeline graph.

I don't think we should do that in this PR, but we might want to consider that approach as part of the larger refactor.

scholtzan:
Cool! I can try to update this PR or open a follow-up PR for the refactoring.

jklukas:
> Cool! I can try to update this PR or open a follow-up PR for the refactoring.

Let's go ahead and do it here. The reduction in code size and increase in readability feels like a significant benefit.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -3,47 +3,51 @@
...
+    return MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via((PubsubMessage msg) -> {
+      msg = PubsubConstraints.ensureNonNull(msg);
+      int numBytes = msg.getPayload().length;
+      if (numBytes > maxBytes) {
+        countPayloadTooLarge.inc();
+        throw new PayloadTooLargeException("Message payload is " + numBytes
+            + "bytes, larger than the" + " configured limit of " + maxBytes);
+      }
+      return msg;
+    }).exceptionsInto(TypeDescriptor.of(PubsubMessage.class))
```
scholtzan: I guess we want to do the exception handling in here?

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/LimitPayloadSize.java
```diff
@@ -3,47 +3,51 @@
...
+        .exceptionsVia((ExceptionElement<PubsubMessage> ee) -> {
+          try {
+            throw ee.exception();
+          } catch (PayloadTooLargeException e) {
+            return FailureMessage.of("LimitPayloadSize", ee.element(), ee.exception());
```
jklukas:
```suggestion
            return FailureMessage.of(LimitPayloadSize.getClass().getSimpleName(), ee.element(), ee.exception());
```

I have a slight preference for accessing the class name this way rather than repeating it as a string literal. Not sure if these are exactly the right method names.

jklukas: And to be clear, I think it makes sense to use string literals when we call `.apply(...)` in the main pipeline definitions. It's just within the context of this class itself that I'd like to avoid repeating the name.

PR #1102, opened 2020-02-03
Add Pub/Sub compression handling to ingestion-sink

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -117,10 +121,16 @@ public Output via(Function<PubsubMessage, CompletableFuture<Void>> write) {
   enum OutputType {
     pubsub {
 
+      private CompressPayload getOutputCompression(Env env) {
+        return CompressPayload.valueOf(env.getString(OUTPUT_COMPRESSION, "none"));
```
jklukas:
```suggestion
        return CompressPayload.valueOf(env.getString(OUTPUT_COMPRESSION, CompressPayload.none.toString()));
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/CompressPayload.java
```diff
@@ -0,0 +1,44 @@
...
+import com.google.protobuf.ByteString;
+import com.google.pubsub.v1.PubsubMessage;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.function.Function;
+import java.util.zip.GZIPOutputStream;
+
+public enum CompressPayload implements Function<PubsubMessage, PubsubMessage> {
+  gzip {
```
jklukas: Do we have other enums in this codebase yet? I'm feeling some regret about the decision to use lowercase enum values in ingestion-beam. It looks nicer on the command-line, but breaks with the style used by enum options used by Beam and breaks with established Java convention for all-caps enum values.

relud: we have `PubsubMessageToObjectNode.Format`, which I'm planning to refactor as part of implementing `OUTPUT_FORMAT=payload`. I'm fine with switching to all-caps.

relud: I switched both enums to use all-caps values

relud: (and use `.toUpperCase()`, for backward compatiblity)

PR #1106, opened 2020-02-04
Perform GeoIP lookup first in Decoder

### LINKED ISSUE #1096: Perform GeoIP lookup first in Decoder
We would like to remove the `ip` field from payload_bytes_error. To do this, we need to make sure that we always have performed GeoIP lookup before a record is marked invalid and sent to error output.

We need to make geo the first transform in the Decoder and ensure that any exceptions in that transform bubble up to shut down the worker rather than going to error output.

cc @whd @relud

jklukas:
To accomplish this, we have to refactor the Read transform so that it does
not return an error collection. Motivation for why this will have no effect
in practice is given in https://github.com/mozilla/gcp-ingestion/issues/1096#issuecomment-581990747

Closes #1096


ingestion-beam/src/test/java/com/mozilla/telemetry/SinkMainTest.java
```diff
@@ -103,28 +103,6 @@ public void canWriteSingleRecord() {
     assertThat(outputLines, matchesInAnyOrder(inputLines));
   }
 
-  @Test
-  public void canWriteErrorOutput() {
```
jklukas: We remove this test entirely because it was only testing handling of malformed PubsubMessage-serialized-to-json messages, a format that doesn't occur in production.

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/PubsubIntegrationTest.java
```diff
@@ -230,20 +230,16 @@ private void publishLines(List<String> lines) throws Exception {
...
-      public void receiveMessage(PubsubMessage message, AckReplyConsumer consumer) {
-        try {
-          String encoded = Json.asString(new org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage(
-              message.getData().toByteArray(), message.getAttributesMap()));
-          received.add(encoded);
-        } catch (IOException e) {
-          throw new UncheckedIOException(e);
-        }
-        consumer.ack();
+    MessageReceiver receiver = ((PubsubMessage message, AckReplyConsumer consumer) -> {
```
jklukas: Unrelated change. While updating this file, I saw IntelliJ suggesting this refactor to reduce verbosity.

ingestion-beam/src/test/resources/testdata/decoder-integration/error-input.ndjson
```diff
@@ -1,4 +1,3 @@
-{}
```
jklukas: We remove this case to avoid throwing an exception in the DecoderMainTest.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/DecodePubsubMessages.java
```diff
@@ -1,56 +1,40 @@
...
 
   /*
    * Static factory methods.
    */
 
   /** Decoder from non-json text to PubsubMessage. */
-  public static Text text() {
-    return new Text();
+  public static PTransform<PCollection<? extends String>, PCollection<PubsubMessage>> text() {
+    return PTransform.compose("DecodePubsubMessages.Text",
```
jklukas: Using PTransform.compose is a trick to be able to return a transform with a descriptive name rather than returning a transform with the generic name "Map". This trick doesn't work for LimitPayloadSize (where @scholtzan recently did a similar refactor) since that transform has to return a `Result` class.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParsePayloadTest.java
```diff
@@ -153,7 +153,7 @@ public void testVersionInPayload() {
         + ",\"document_type\":\"main\"" + "},\"payload\":\"eyJ2ZXJzaW9uIjo0fQ==\"}";
 
     WithErrors.Result<PCollection<PubsubMessage>> result = pipeline.apply(Create.of(input))
-        .apply(InputFileFormat.json.decode()).output()
+        .apply(InputFileFormat.json.decode())
```
scholtzan: Out of curiosity, what's the rule for putting `//` at the end of a line? Seems like it's sometimes there when `apply`ing multiple transforms and sometimes it isn't

jklukas: It's purely there to force google-java-format to allow a line break. There are lots of cases where it's not actually needed since the following line is already long enough that the formatter would have put it on its own line anyway. I personally tend to add `//` wherever I want a linebreak to be whether the formatter would remove it or not. There's definitely some personal preference involved.

jklukas: Generally, the longer the list of `apply`s, the more likely that I'll want to add `//` to force each `apply` to its own line to make it easier to skim the code and quickly understand the sequence of transforms.

PR #1107, opened 2020-02-05
Move SchemaStore into ingestion-core

relud:


jklukas:
Do you have a concrete sense of what it's going to look like to use `SchemaStore` in ingestion-sink? How are you planning to pick up the object from GCS vs. local file system without using Beam's filesystem API?

If we're going to have to implement logic for reading from local filesystem vs. GCS, I'd prefer to use that consistently between ingestion-beam and ingestion-sink rather than subclassing the schema stores to provide a Beam-specific implementation in one case.

I worry that this PR is adding avoidable complexity to the implementation of schema stores.

relud:
> Do you think it makes sense to use unified logic between the two projects for reading from GCS/local filesystem?

I think that having support in ingestion-sink for reading schemas from GCS isn't needed, and implementing that without including the beam libraries in ingestion-core would be more complex than mixing in beam's `FileSystems` outside of ingestion-core.

relud:
> Do you have a concrete sense of what it's going to look like to use `SchemaStore` in ingestion-sink? How are you planning to pick up the object from GCS vs. local file system without using Beam's filesystem API?

`ingestion-sink` will instantiate `BigQuerySchemaStore` by providing `String`s, rather than `ValueProvider<String>`, and will receive local filesystem paths.

> If we're going to have to implement logic for reading from local filesystem vs. GCS, I'd prefer to use that consistently between ingestion-beam and ingestion-sink rather than subclassing the schema stores to provide a Beam-specific implementation in one case.

that sounds like a good plan, I just don't know how to do it well. This PR was originally using `Supplier<String>` to get around the `ValueProvider` dependency and `Function<String, InputStream>` to get around the `Filesystems` dependency, but that didn't work because it caused issues with serializing `AvroSchemaStore`.

> I worry that this PR is adding avoidable complexity to the implementation of schema stores.

it's a good worry to have. I feel like most of the complexity here could be simplified, but getting around issues with serializing `AvroSchemaStore` got in the way.

I'd like to video chat about this later today, if you have time, to discuss what the best path forward is.

relud:
Updated `AvroSchemaStore` to be used as `transient` and instantiated after values are available from `ValueProvider` in `ingestion-beam`. re-r?

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -229,45 +265,20 @@ public AvroOutput(ValueProvider<String> outputPrefix, Duration windowDuration,
...
           .<List<String>, PubsubMessage>writeDynamic() //
           .by(message -> pathTemplate.get().extractValuesFrom(message.getAttributeMap()))
           .withDestinationCoder(ListCoder.of(StringUtf8Coder.of())) //
           .withCompression(compression) //
-          .via(Contextful.fn((List<String> dest, Contextful.Fn.Context ctx) -> {
-            Map<String, String> attributes = pathTemplate.get().getPlaceholderAttributes(dest);
-            Schema schema = ctx.sideInput(schemaSideInput).getSchema(attributes);
-            return AvroIO.sinkViaGenericRecords(schema, binaryFormatter);
-          }, Requirements.requiresSideInputs(schemaSideInput))) //
+          .via(Contextful.fn(encoder::getSink)) //
```
relud: cc @acmiyaguchi re: changing this from a side input to a transient class variable

PR #1109, opened 2020-02-05
Draft docs for AET decryption support

jklukas:
Details here will likely change, but the intention of this draft PR is to make a first pass at a concrete specification of the interface for AET decryption to help move the overall AET design discussion forward.

The current version of the rendered docs (without the changes in this PR) is available at https://mozilla.github.io/gcp-ingestion/

The basic interface proposed here means that clients sending AET pings would include the anon_id as a header named `X-Ecosystem-Anon-ID`. The schemas for the target tables in BigQuery would be expected to include a top-level string-type field named `ecosystem_user_id` that would get populated with the decrypted value.

### DIFF
--- a/docs/architecture/edge_service_specification.md
+++ b/docs/architecture/edge_service_specification.md
@@ -50,6 +50,8 @@ required group attributes {
   optional string x_pingsender_version // example: "1.0"
   optional string x_debug_id           // example: "my_debug_session_1"
   optional string x_pipeline_proxy     // time that the AWS->GCP tee received the message, example: "2018-03-12T21:02:18.123456Z"
+  optional string x_ecosystem_anon_id  // Anonymized identifier for Account Ecosystem Telemetry
+  optional string x_prev_ecosystem_anon_id  // Previous anon_id value; sent when rotating identifiers
 }
 ```
 
--- a/docs/architecture/overview.md
+++ b/docs/architecture/overview.md
@@ -49,7 +49,7 @@ This document specifies the architecture for GCP Ingestion as a whole.
 - Should retry transient Cloud Storage errors indefinitely
     - Should use exponential back-off to determine retry timing
 
-### `EcosystemDecoder`
+### `EcosystemDecryptor`
 
 - Must have access restricted to a limited set of operators
 - Must load a bundle of private keys from Cloud KMS
--- a/docs/architecture/overview.md
+++ b/docs/architecture/overview.md
@@ -10,6 +10,9 @@ This document specifies the architecture for GCP Ingestion as a whole.
   Firefox) to PubSub `Raw Topics`
 - The Dataflow `Landfill Sink` job copies messages from PubSub `Raw Topics` to
   `Cloud Storage`
+- The Dataflow `EcosystemDecryptor` job reads from a `Raw Topic` specific to
+  Account Ecosystem Telemetry (AET), decrypting AET-specific identifiers
+  stored as attributes and republishing to a special decrypted `Raw Topic`
 - The Dataflow `Decoder` job decodes messages from PubSub `Raw Topics` to
   PubSub `Decoded Topics`
     - The Dataflow `Decoder` job checks for existence of `document_id`s in
@@ -45,6 +48,15 @@ This document specifies the architecture for GCP Ingestion as a whole.
 - Must accept configuration mapping PubSub topics to Cloud Storage locations
 - Should retry transient Cloud Storage errors indefinitely
     - Should use exponential back-off to determine retry timing
+
+### `EcosystemDecoder`
+
+- Must have access restricted to a limited set of operators
+- Must load a bundle of private keys from Cloud KMS
+- Must replace attribute `x_ecosystem_anon_id` with the decrypted form
+  `ecosystem_user_id`
+- Must replace attribute `x_prev_ecosystem_anon_id` (if present) with the
+  decrypted form `prev_ecosystem_user_id`
 
 ### Decoder
 
--- a/docs/ingestion-beam/ecosystem-decryptor-job.md
+++ b/docs/ingestion-beam/ecosystem-decryptor-job.md
@@ -1,0 +1,26 @@
+# EcosystemDecryptor Job
+
+A job for decrypting identifiers used in Account Ecosystem Telemetry.
+Defined in the `com.mozilla.telemetry.EcosystemDecryptor` class ([source](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/EcosystemDecryptor.java)).
+On startup, this job loads a bundle of private keys by contacting GCP's KMS service.
+
+## Transforms
+
+These transforms are currently executed against each message in order.
+
+### Decrypt Account Ecosystem Telemetry Identifiers
+
+Strips encrypted AET attributes from the message, decrypts them using the
+loaded bundle of data pipeline private keys, and adds attributes for the
+decrypted IDs:
+
+* Input attribute `x_ecosystem_anon_id` will be transformed to output attribute
+  `ecosystem_user_id`
+* Input attribute `x_prev_ecosystem_anon_id` (if present) will be transformed to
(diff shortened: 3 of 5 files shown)


docs/architecture/edge_service_specification.md
```diff
@@ -50,6 +50,8 @@ required group attributes {
   optional string x_pingsender_version // example: "1.0"
   optional string x_debug_id           // example: "my_debug_session_1"
   optional string x_pipeline_proxy     // time that the AWS->GCP tee received the message, example: "2018-03-12T21:02:18.123456Z"
+  optional string x_ecosystem_anon_id  // Anonymized identifier for Account Ecosystem Telemetry
+  optional string x_prev_ecosystem_anon_id  // Previous anon_id value; sent when rotating identifiers
```
jklukas: I need to get more detail on the size of these `anon_id` payloads. The maximum size that a `PubsubMessage` supports per attribute value is 1024 bytes. If `anon_id` needs to be larger than this, than we have to encode it in the payload. We likely support header values up to at least 8K, though, so we'd likely still be able to keep headers as the interface for clients to send this info.

docs/architecture/overview.md
```diff
@@ -46,6 +49,15 @@ This document specifies the architecture for GCP Ingestion as a whole.
 - Should retry transient Cloud Storage errors indefinitely
     - Should use exponential back-off to determine retry timing
 
+### `EcosystemDecoder`
```
jklukas:
```suggestion
### `EcosystemDecryptor`
```

docs/architecture/edge_service_specification.md
jklukas: Matt Miller indicates that the payloads may be 300-400 bytes, but not as larger as 1024, so we should be fine to rely on Pubsub attributes.

docs/ingestion-beam/ecosystem-decryptor-job.md
```diff
@@ -0,0 +1,26 @@
+# EcosystemDecryptor Job
+
+A job for decrypting identifiers used in Account Ecosystem Telemetry.
+Defined in the `com.mozilla.telemetry.EcosystemDecryptor` class ([source](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/EcosystemDecryptor.java)).
```
jklukas: This does not exist yet.

docs/ingestion-beam/ecosystem-decryptor-job.md
```diff
@@ -0,0 +1,26 @@
...
+* Input attribute `x_ecosystem_anon_id` will be transformed to output attribute
+  `ecosystem_user_id`
+* Input attribute `x_prev_ecosystem_anon_id` (if present) will be transformed to
+  output attribute `prev_ecosystem_user_id`
+
+## Formats
+
+The input `x_ecosystem_anon_id` and `x_prev_ecosystem_anon_id` attribute strings must
+be JOSE JWE objects in Compact Serialization form. The emitted `ecosystem_user_id`
+and `prev_ecosystem_user_id` attribute strings are UUIDs in 36-character hex standard form.
```
linuxwolf: Is it important that the `ecosystem_user_id` strictly be a UUID?  We can guarantee the "alphabet" (hexadecimal) and its uniqueness, but I don't know the format can be strictly guaranteed.   UUID carries some restrictions that are kind of inconvenient.

The Firefox Account `uid` is a 32-character string of hexadecimal, but is not strictly a UUID.  I think it safer to expect the `ecosystem_user_id` match an identical format.

docs/architecture/overview.md
```diff
@@ -10,6 +10,9 @@ This document specifies the architecture for GCP Ingestion as a whole.
   Firefox) to PubSub `Raw Topics`
 - The Dataflow `Landfill Sink` job copies messages from PubSub `Raw Topics` to
   `Cloud Storage`
+- The Dataflow `EcosystemDecryptor` job reads from a `Raw Topic` specific to
```
linuxwolf: Would it be possible to indicate where this operates in relation to the jobs?  Ideally, updating the diagram would help, otherwise some simple textual description for when this invokes relative to (I think?) `Ingestion Edge`, `Landfill Sink`, and `Decoder` would help me.

docs/ingestion-beam/ecosystem-decryptor-job.md
jklukas: UUID is not a hard requirement; I'm mostly concerned with having the format be predictable and well-documented. A 32-character hex string sounds reasonable.

docs/architecture/overview.md
```diff
@@ -46,6 +49,15 @@ This document specifies the architecture for GCP Ingestion as a whole.
 - Should retry transient Cloud Storage errors indefinitely
     - Should use exponential back-off to determine retry timing
 
+### `EcosystemDecryptor`
+
+- Must have access restricted to a limited set of operators
+- Must load a bundle of private keys from Cloud KMS
```
scholtzan: It doesn't seem to be possible to get the private keys that are stored in Cloud KMS. It is also mentioned that decryption/encryption needs to be done within KMS (for example mentioned here: https://cloud.google.com/kms/docs/faq#export_keys)

jklukas: This requires some more investigation, then. From my understanding of KMS, one possibility is that you can ask KMS to hand you a session key derived from the actual private key that it's holding, and then you use that locally held session key to do the actual decrypting of payloads. But it's not clear to me whether we can take advantage of that in the context of our particular use case.

If we truly do need access to the exact private key associated with the public key that the client used to encrypt the original value, then indeed it sounds like KMS is not the right thing to use and we'd instead likely want to store the key in GCS with highly restricted access.

linuxwolf: A KMS will never provide the private key, that is true.  I see a couple of options:

* Use the KMS directly.  My gut reaction is this gets very expensive very quickly, but I don't have any real numbers to extrapolate to that result.
* Generate in software a public/private keypair for actually decrypting payloads (call it the "payload keypair"), encrypt **that** private key using the KMS and store it.  When the payload keypair needs to be used, load it and decrypt it through the KMS.

scholtzan: The [pricing for KMS](https://cloud.google.com/kms/pricing) is $0.03 per 10,000 operations. Looking at the number of daily active users for Firefox Accounts, it would get quite expensive ($300-$400 per day).

linuxwolf: The second option seems the most cost-effective.  It does assume the bundle is loaded (and therefore decrypted) fairly infrequently.

It could be worthwhile to check with the Autograph team (cc @jvehent ).

jvehent: I wouldn't use a KMS directly. You probably want to process pings much faster than the KMS could provide. The typical use of a KMS is to decrypt a known key or secret at initialization, use that decrypted key or secret from memory, and never write it to disk. This is how autograph loads some of its secrets (when they are not stored in an HSM).

The second option is to use an HSM. HSMs are designed for speed, and can handle many thousands operations per seconds at low latency. They aren't that expensive anymore (in the grand scheme of things), but integrating with them can be somewhat tricky, because you need to deal with all the pkcs11 shenanigans.

Autograph uses an AWS CloudHSM for most of the Firefox signing keys, so we have hands-on with that technology. GCP also has a CloudHSM, but we don't have any experience using it.

scholtzan: It looks like GCPs CloudHSM has the same pricing as KMS with $0.03 per 10,000 operations (or $0.15 for other key operations): https://cloud.google.com/hsm#pricing

linuxwolf: Thanks @jvehent .

I think our best course of action is to follow the "known key" option (my first option/jvehent's first option).

The important point is to *never write the decrypted known private key to disk*.  It's fine to write the _*encrypted*_ version of that private key.

jklukas:
Closing in favor of an updated design in https://github.com/mozilla/gcp-ingestion/pull/1264

PR #1112, opened 2020-02-05
Limit decoded payload size to 8 MB

### LINKED ISSUE #776: Consider sending large payloads to the errors stream
For the sync pings that were thrown away at the BQ sink output, it might be better to annotate these pings as errors at the decoder with "payload too large" or similar. The calculation on what the actual size would be is perhaps nontrivial if we're targeting BQ with various compression/decompression steps, but I was under the impression we had documented as a matter of policy (perhaps not enforced) the maximum payload size for records to be ingested by our pipeline(s). We do enforce this in AWS with a max message bytes of I believe 8MB.

jklukas:
Closes #776 and also addresses documentation request in
https://bugzilla.mozilla.org/show_bug.cgi?id=1605503

### DIFF
--- a/docs/architecture/overview.md
+++ b/docs/architecture/overview.md
@@ -211,7 +211,7 @@ view.
 ## Limits
 
 - The maximum `Content-Length` accepted at the edge is 1 MB; larger payloads
-  will be dropped
+  will be dropped and [the request will return a 413 response code](https://mozilla.github.io/gcp-ingestion/architecture/edge_service_specification/#server-requestresponse)
 - The maximum payload size after being decompressed in the Decoder is 8 MB;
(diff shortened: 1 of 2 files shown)


docs/architecture/overview.md
```diff
@@ -208,10 +208,15 @@ comfortable that the new implementation is stable, we could cut users over to
 the new implementation by simply changing the definition of the user-facing
 view.
 
-## Known Issues
-
-- Hard limit of 10,000 columns per table in BigQuery
-- Max of 100,000 streaming inserts per second per BigQuery table
+## Limits
+
+- The maximum `Content-Length` accepted at the edge is 1 MB; larger payloads
```
whd: It's possibly worth referencing the HTTP edge spec and that the "dropped" mechanism is a 413.

PR #1115, opened 2020-02-06
Bug: ParseProxy must occur before GeoCityLookup

jklukas:
Fixup to #1106 which has not yet been deployed to prod.
Geo lookup is broken in current master.


jklukas:
I'm going to merge this as soon as tests pass and ask for retroactive review. We'll be in a bad spot if current master gets deployed.

Issue #1059, opened 2020-01-06
Error message from worker: java.io.UncheckedIOException: com.maxmind.db.InvalidDatabaseException: The MaxMind DB file's data section contains bad data: pointer larger than the database.


whd:
[Example failing stage job](https://console.cloud.google.com/dataflow/jobs/us-west1/2019-12-30_16_14_59-11135197587585267802?project=moz-fx-data-beam-nonprod-b5f4&organizationId=442341870013) and [stack trace](https://console.cloud.google.com/errors/CLKijszqiayVPg?time=P1D&project=moz-fx-data-beam-prod-11f7&organizationId=442341870013).

GeoIP DB: gs://moz-fx-data-prod-geoip/GeoIP2-City/20191231/GeoIP2-City.mmdb 

This appears to be the result of an actual GeoIP DB change, as when I reverted production to 20191224 the error stopped.

jklukas:
I ran into a perhaps related error while running a backfill job to add geo info to payload_bytes_error:

```
 java.lang.RuntimeException: org.apache.beam.sdk.util.UserCodeException: java.io.UncheckedIOException: java.nio.charset.MalformedInputException: Input length = 1
```

I was using the latest DB: gs://moz-fx-data-prod-geoip/GeoIP2-City/20200204/GeoIP2-City.mmdb

Of note, I ran a sample job over just a single day of data and it completed successfully. This full backfill job looks like it still managed to process billions of messages successfully, so it's maybe a transient error? But I think the pipeline would have recovered if that were the case.

jklukas:
This case could actually be a bad record. The line in our ingestion-beam code where this is happening is not when we load the DB, but rather while processing a response.

jklukas:
Further digging into the stack trace shows that it's the call to `geoIP2.city` that throws the error. The MaxMind `CharsetDecoder#decodeLoop` method where the issue is thrown internally documents:

```
     * @throws  MalformedInputException
     *          If the byte sequence starting at the input buffer's current
     *          position is not legal for this charset and the current malformed-input action
     *          is {@link CodingErrorAction#REPORT}
```

Perhaps we should just catch this case?

jklukas:
I moved ahead with the catching this case for the backfill of geo to the errors table. I added a counter for the malformed input and there were only 2 instances. The job finished successfully. I will file a PR to add the same changes to gcp-ingestion.

[...]

PR #1123, opened 2020-02-10
Catch MalformedInputException in GeoCityLookup

jklukas:
See https://github.com/mozilla/gcp-ingestion/issues/1059


jklukas:
> If MalformedInputException is geoip-specific it might be good to have geoip in the counter somewhere e.g. `geoip_malformed_input`.

The path of the metric includes the transform name, so I was trying not to be redundant with that context. If you think it's likely to be ambiguous in practice, though, I'm happy to make the change.

whd:
> The path of the metric includes the transform name, so I was trying not to be redundant with that context. If you think it's likely to be ambiguous in practice, though, I'm happy to make the change.

Ok that's fine, I just wanted to make sure it was included in the metric path somewhere.

PR #1124, opened 2020-02-11
Implement live sink format in ingestion-sink

relud:


relud:
It needs test cases with a more complex schema, but should be otherwise complete.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -298,9 +302,21 @@ final long getMaxOutstandingRequestBytes(Env env) {
...
+    final String format = env.getString(OUTPUT_FORMAT, "raw").toLowerCase();
+    switch (format) {
+      case "raw":
+        return PubsubMessageToObjectNode.Raw.of();
+      case "decoded":
+        return PubsubMessageToObjectNode.Decoded.of();
+      case "payload":
+        // TODO
+        return PubsubMessageToObjectNode.Payload.of(env.getStrings(STRICT_SCHEMA_DOCTYPES, null),
+            env.getString(SCHEMAS_LOCATION), env.getString(SCHEMA_ALIASES_LOCATION, null),
```
relud: @whd for [bug 1612367](https://bugzilla.mozilla.org/show_bug.cgi?id=1612367) would you please deploy this branch as a nonprod live sink, with `STRICT_SCHEMA_DOCTYPES` as an optional CSV of `namespace/doctype` values, `SCHEMAS_LOCATION` as a local path to `schemas.tar.gz`, `SCHEMA_ALIASES_LOCATION` as an optional local path to an aliasing config, and `OUTPUT_FORMAT=payload`.

whd: Yep. I'll ping you if I have trouble setting this up. Do you want the 100% tests (i.e. different temp tables and prod subscriptions) or just the nonprod sample tables?

relud: let's go with 100%, and `OUTPUT_MODE=mixed`

jklukas: See https://github.com/mozilla/gcp-ingestion/issues/1136; I believe there's no point in supporting schema aliases.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -1,79 +1,536 @@
...
+          String namespace = message.getAttributesOrDefault(Attribute.DOCUMENT_NAMESPACE, "");
+          String docType = message.getAttributesOrDefault(Attribute.DOCUMENT_TYPE, "");
+          return strictSchemaDocTypeSet.contains(namespace + "/" + docType);
+        };
+      }
+      schemaStore = BigQuerySchemaStore.of(schemasLocation, schemaAliasesLocation, open);
+    }
+
+    /** measure rate of CoercedToInt. */
+    private void incrementCoercedToInt() {
```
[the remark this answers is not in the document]
relud: > I'm guessing the big problem here is that we need to manipulate TableRow objects in the one context and ObjectNode in the other. Do you also see that as the blocker? We should make sure we think through that problem and make sure there's no way around it.

I absolutely think we can share this code between the two. jackson's ObjectMapper can be used to convert the ObjectNode to a TableRow. It's inconvenient that PubsubMessage is different in beam, but we can also work around that. The only real reason I didn't make this shared is that it will take longer to write.

 I'll file an issue to come back and share the implementation.

relud:
This is now successfully delivering messages and making progress on backlog in stage, so I plan to check on it again tomorrow morning and if all is well I will merge.

relud:
> I plan to check on it again tomorrow morning and if all is well I will merge.

stage updated at around 6:30PM last night, which reverted to the `latest` tag, and reduced the autoscaling max down from 200 to 100.

I have now updated the `latest` tag to match the `b0caabd` tag, updated the kube deployment to pick up the updated image, and updated the autoscaling back to 200. I'll check on it again tomorrow and merge then.

@whd could you update the configs to have a max autoscaling to 200? 100 isn't enough to keep up, but it looks like 200 should be.

whd:
> @whd could you update the configs to have a max autoscaling to 200? 100 isn't enough to keep up, but it looks like 200 should be.

Done. Assuming there's a schemas change introduced around 6pm tonight, the same redeploy is going to happen. Given your other changes it looks like there should be no problem however.

relud:
> Assuming there's a schemas change introduced around 6pm tonight, the same redeploy is going to happen. Given your other changes it looks like there should be no problem however.

ah, schemas. I forgot those caused deploys. sounds good.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -240,15 +239,18 @@ Output getOutput(Env env) {
...
+              if (t.getCause() instanceof BigQuery.WriteErrors) {
+                BigQuery.WriteErrors cause = (BigQuery.WriteErrors) t.getCause();
+                if (cause.errors.size() == 1 && cause.errors.get(0).getMessage()
+                    .startsWith("Maximum allowed row size exceeded")) {
+                  return fileOutput.apply(message);
+                }
+              }
+              throw (RuntimeException) t;
+            }).thenCompose(v -> v);
+        return new Output(env, this, mixedOutput);
```
relud: > Row size is calculated for the encoded row, which is within a few bytes of protobuf serialized size, so the default is 1MB to leave room for differences.

@jklukas I rewrote this to use the file output only if BigQuery actually rejects the message, because:

- the encoded row for `payload` format is usually much smaller than the JSON
- the live sink is the only one currently using `mixed` mode
- we are limited 1,000 or fewer messages per table via `fileOutput` in `mixed` mode, because every worker submits their own load jobs

downsides of this change include:
- there is a reference to `message` kept until it is successfully delivered to `streamingOutput`, so it can't be garbage collected
  - In practice this doesn't seem to have any impact on memory consumption, so I suspect it's not the only lingering reference.
- messages sent to `fileOutput` go through `PubsubMessageToObjectNode` twice
  - we could apply that as a transform, but it's a non-trivial change that I don't want to make that change as part of this PR

jklukas: Glancing through the BigQuery code, it looks like we're able to reasonable efficiently map individual messages to their associated errors. This looks reasonable. It would be good to send a 2 MB message and test that this all behaves as we expect in practice.

relud: > It would be good to send a 2 MB message and test that this all behaves as we expect in practice.

already did ^_^ https://github.com/mozilla/gcp-ingestion/pull/1124/files#diff-a8c5a168b5959568d77819c522c24427R14-R34

PR #1126, opened 2020-02-11
Use MapWithFailures for KeyByBigQueryTableDestination

scholtzan:
I continued refactoring `KeyByBigQueryTableDestination` to use `MapWithFailure` as part of https://github.com/mozilla/gcp-ingestion/issues/118

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/KeyByBigQueryTableDestination.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/KeyByBigQueryTableDestination.java
@@ -28,14 +28,17 @@ import org.apache.beam.sdk.io.gcp.bigquery.TableDestination;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.transforms.MapElements;
-import org.apache.beam.sdk.transforms.MapElements.MapWithFailures;
+import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.WithFailures;
+import org.apache.beam.sdk.transforms.WithFailures.Result;
 import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
 import org.apache.beam.sdk.values.TypeDescriptor;
 import org.apache.beam.sdk.values.TypeDescriptors;
 import org.apache.commons.text.StringSubstitutor;
 
-public class KeyByBigQueryTableDestination {
+public class KeyByBigQueryTableDestination extends PTransform<PCollection<PubsubMessage>, //
+    Result<PCollection<KV<TableDestination, PubsubMessage>>, PubsubMessage>> {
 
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/KeyByBigQueryTableDestination.java
```diff
@@ -121,21 +121,23 @@ public TableDestination getTableDestination(Map<String, String> attributes) {
...
-      PCollection<PubsubMessage> input) {
-    WithErrors.Result<PCollection<KV<TableDestination, PubsubMessage>>> result = super.expand(
-        input);
-    result.output()
-        .setCoder(KvCoder.of(TableDestinationCoderV3.of(), PubsubMessageWithAttributesCoder.of()));
-    return result;
+  /**
+   * Returns a MapElements transform for retrieving the table destinations.
+   */
+  public MapWithFailures<PubsubMessage, KV<TableDestination, PubsubMessage>, PubsubMessage> map() {
```
scholtzan: I don't see how it's possible to refactor this to be a static method such as `LimitPayloadSize.of()` introduced in https://github.com/mozilla/gcp-ingestion/pull/1083/files 
Not sure if there is a better way or name for this in this case?

relud: something like this:
```java
  public MapWithFailures<PubsubMessage, KV<TableDestination, PubsubMessage>, PubsubMessage> of(
      ValueProvider<String> tableSpecTemplate, ValueProvider<String> partitioningField,
      ValueProvider<List<String>> clusteringFields) {
    KeyByBigQueryTableDestination transform = new KeyByBigQueryTableDestination(tableSpecTemplate, partitioningField,
        clusteringFields);
    return MapElements.into(TypeDescriptors.kvs(TypeDescriptor.of(TableDestination.class),
        TypeDescriptor.of(PubsubMessage.class))).via((PubsubMessage msg) -> {
          msg = PubsubConstraints.ensureNonNull(msg);
          return KV.of(transform.getTableDestination(msg.getAttributeMap()), msg);
        }).exceptionsInto(TypeDescriptor.of(PubsubMessage.class))
        .exceptionsVia((WithFailures.ExceptionElement<PubsubMessage> ee) -> {
          try {
            throw ee.exception();
          } catch (LimitPayloadSize.PayloadTooLargeException e) {
            return FailureMessage.of(KeyByBigQueryTableDestination.class.getSimpleName(),
                ee.element(), ee.exception());
          }
        });
  }
```

jklukas: Since this class also has the non-static `getTableDestination` method, it might make sense in this case to have `KeyByBigQueryTableDestination` extend `PTransform` and put the `MapElements` call into the `expand` method. That way we can keep the existing `.of` factory method.

scholtzan: Yeah, I think that might be the most elegant solution here.

scholtzan:
Ready for another look.

PR #1127, opened 2020-02-12
Bug 1489560 scrub messages with client_id c0ffeec0-ffee-c0ff-eec0-ffe…

scholtzan:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1489560

Scrubs all messages with client_id `c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0`

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -9,19 +9,24 @@ import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
 import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Stream;
-import org.apache.beam.sdk.metrics.Counter;
+
+import com.mozilla.telemetry.transforms.MapElementsWithErrors.MessageShouldBeDroppedException;
 import org.apache.beam.sdk.metrics.Metrics;
 
 public class MessageScrubber {
+  private enum ScrubAction {
+    DROP,
+    REDACT,
+    SEND_TO_ERRORS
+  }
 
-  private static final Counter countScrubbedBug1567596 = Metrics.counter(MessageScrubber.class,
-      "bug_1567596");
-  private static final Counter countScrubbedBug1562011 = Metrics.counter(MessageScrubber.class,
-      "bug_1562011");
-  private static final Counter countScrubbedBug1489560 = Metrics.counter(MessageScrubber.class,
-      "bug_1489560");
-  private static final Counter countRedactedBug1602844 = Metrics.counter(MessageScrubber.class,
-      "bug_1602844");
+  /**
+   * Special exception to signal that a message is affected by a specific bug and should
+   * be written to error output.
+   */
+  public static class AffectedByBugException extends Exception {
+    public AffectedByBugException(String bugNumber) {}
+  }
 
   /**
    * Inspect the contents of the payload and return true if the content matches a known pattern
@@ -29,7 +34,8 @@ public class MessageScrubber {
    *
    * <p>This is usually due to some potential for PII having leaked into the payload.
    */
-  public static boolean shouldScrub(Map<String, String> attributes, ObjectNode json) {
+  public static void scrub(Map<String, String> attributes, ObjectNode json)
+      throws MessageShouldBeDroppedException, AffectedByBugException {
     // https://bugzilla.mozilla.org/show_bug.cgi?id=1567596
     if (ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
         && "crash".equals(attributes.get(Attribute.DOCUMENT_TYPE))
@@ -39,8 +45,7 @@ public class MessageScrubber {
             .map(j -> j.path("payload").path("metadata").path("MozCrashReason").textValue())
             .filter(s -> s.contains("do not use eval with system privileges")) //
             .isPresent()) {
-      countScrubbedBug1567596.inc();
-      return true;
+      handleBug("1567596", ScrubAction.DROP);
     } else if (ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
         && "crash".equals(attributes.get(Attribute.DOCUMENT_TYPE))
         && (("nightly".equals(attributes.get(Attribute.APP_UPDATE_CHANNEL))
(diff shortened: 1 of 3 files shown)


whd:
Note that the ask in https://bugzilla.mozilla.org/show_bug.cgi?id=1489560#c5 is to keep these data accessible as e.g. errors, while this change as-is would mean this data is only accessible via payload_bytes_raw.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -44,6 +44,14 @@ public static ParsePayload of(ValueProvider<String> schemasLocation,
     return new ParsePayload(schemasLocation, schemaAliasesLocation);
   }
 
+  /** Exception to throw for messages whose payloads exceed the configured size limit. */
+  public static class MessageToErrorsException extends RuntimeException {
```
scholtzan: Maybe a better name would be good?

badboy:
Quick question: does this apply to only legacy Telemetry or also new and shiny Glean data?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -98,6 +106,11 @@ protected PubsubMessage processElement(PubsubMessage message)
       throw new MessageShouldBeDroppedException();
     }
 
+    if (MessageScrubber.shouldScrubAndWriteToErrors(attributes, json)) {
+      // Write the message to error output.
+      throw new MessageToErrorsException("Message should be scrubbed");
+    }
+
```
jklukas: I think there's enough complexity here now that we should refactor all the responsibility into `MessageScrubber` itself. So in ParsePayload, we would just have a single call:

```
// For various messages signatures, this may redact fields from json or throw an exception
// that causes the message to be dropped or routed to error output.
MessageScrubber.scrub(attributes, json);
```

Then within `scrub`, each conditional would either modify the json payload, throw `MessageShouldBeDroppedException` or a new `AffectedByBugException` where we pass the bug number into the error message. 

To avoid having to repeat the bug number several times (I'm getting worried at this point that there's a high likelihood we'll have a copy/paste problem where we end up incrementing the wrong counter, etc.), perhaps we should have a new method:

```
public enum ScrubAction {
  DROP,
  REDACT,
  SEND_TO_ERRORS
}

public static void handleBug(String bugNumber, ScrubAction action) {
  Metrics.counter(MessageScrubber.class, "bug_" + bugNumber).inc();
  switch(action) {
    case DROP:
      throw new MessageShouldBeDroppedException();
    case SEND_TO_ERRORS:
      throw new AffectedByBugException(bugNumber);
    case REDACT:
      return;
  }
}
```

jklukas: Although now that I write this and consider @badboy's comment, we probably should delay the handling for this bug until near the end of this transform so that we can take advantage of the logic later on for parsing clientId out of various potential locations within the payload.

If some glean clients also are having their id set to "coffee", we should include those too.

I think we want to drop toxic messages as early as possible in case they raise some exeption within the transform that causes them to go to error output rather than being dropped. But for messages that we want to send to error output rather than dropping, we can wait until the end of the method. So I think that's an argument for having a `shouldDrop` method and a `shouldWriteToErrors` method.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -145,6 +143,9 @@ protected PubsubMessage processElement(PubsubMessage message)
 
     addAttributesFromPayload(attributes, json);
 
+    // Write messages that are affected by a specific bug to error output.
+    MessageScrubber.writeToErrors(attributes, json);
```
scholtzan: Moving this down here, both Glean and legacy telemetry data should get handled correctly.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
scholtzan: I ended up refactoring `MessageScrubber` a little and still keep separate methods that can be called for scrubbing or writing messages to error. I liked the idea of throwing Exceptions in `MessageScrubber` since that reduces the complexity a little, but I'm also open to change them back to being `boolean` and throw exceptions when parsing the payload.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -84,16 +87,48 @@ static boolean bug1602844Affected(Map<String, String> attributes) {
...
       json.path("events").elements().forEachRemaining(event -> {
         JsonNode eventMapValues = event.path(5);
         if (eventMapValues.has("fxauid")) {
           ((ObjectNode) eventMapValues).replace("fxauid", NullNode.getInstance());
-          countRedactedBug1602844.inc();
+        }
+
+        try {
+          handleBug("1602844", ScrubAction.REDACT);
+        } catch (Exception e) {
```
jklukas: I can't think of a situation where handleBug would throw an exception other than the two checked ones, so this is probably fine. But as a rule, it's dangerous to catch `Exception` generically and drop it.

This line should catch only MessageShouldBeDroppedException and AffectedByBugException

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -6,28 +6,36 @@
...
+    DROP, REDACT, SEND_TO_ERRORS
+  }
+
+  /**
+   * Special exception to signal that a message is affected by a specific bug and should
+   * be written to error output.
+   */
+  public static class AffectedByBugException extends Exception {
+
+    public AffectedByBugException(String bugNumber) {
```
jklukas: We never do anything with the bugNumber here. In the constructor body here, we should call `super(...)` in order to pass a message like `"Affected by bug " + bugNumber`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -84,16 +87,48 @@ static boolean bug1602844Affected(Map<String, String> attributes) {
...
     }
   }
 
+  /**
+   * Inspect the contents of the payload and return true if the content matches a known pattern
+   * we want to scrub the message and write to errors.
+   */
+  public static void writeToErrors(Map<String, String> attributes, ObjectNode json)
+      throws MessageShouldBeDroppedException, AffectedByBugException {
+    if (ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE)) //
```
jklukas: For Glean pings, the document namespace will not be "telemetry"; rather, it is generally the application ID. As merged here, this PR supports removing the coffee id from desktop telemetry and from core pings, but we are not catching coffee IDs for Glean.

PR #1131, opened 2020-02-12
Upgrade to python 3.8

relud:
This relies on [relud/wrk](https://hub.docker.com/r/relud/wrk) to build wrk from source on debian buster so that we are no longer stuck on debian stretch (see also #743).

This removes some dependencies because `pluggy` and `pytest` required `importlib-metadata` for `python < 3.8`

This fixes dependabot PRs because dependabot is building `requirements.txt` with `python3.8`.

### DIFF
--- a/ingestion-edge/Dockerfile
+++ b/ingestion-edge/Dockerfile
@@ -1,11 +1,22 @@
-FROM python:3.7-stretch
+ARG PYTHON_VERSION=3.8
+
+# build requirements in separate stage because it requires gcc and libc-dev
+FROM python:${PYTHON_VERSION}-slim
 WORKDIR /app
-RUN apt-get update && apt-get install -y wrk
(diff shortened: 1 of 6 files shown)


ingestion-edge/Dockerfile
```diff
@@ -1,11 +1,19 @@
...
+RUN apt-get update && apt-get install -qqy gcc libc-dev
 COPY requirements.txt /app/
 COPY bin/include/common.sh /app/bin/include/
 COPY bin/build /app/bin/
 ENV VENV=false
 RUN bin/build
+
+FROM python:${PYTHON_VERSION}-slim
+WORKDIR /app
+COPY --from=relud/wrk:4.1.0-buster /usr/local/bin/wrk /usr/local/bin/wrk
```
jklukas: Please add a comment here that links to the source for `relud/wrk` and explains why this is needed.

PR #1132, opened 2020-02-12
Wait for all 5 coverage reports from CI in codecov

relud:


codecov.yml
```diff
@@ -0,0 +1,3 @@
+codecov:
+  notify:
+    after_n_builds: 5
```
jklukas: Some of the tasks short-circuit if relevant code hasn't been touched, right? Do we not generate reports in that case? Will this potentially be waiting forever?

relud: oh, good point. i'll close this for now. Maybe I later I can set up sending the codecov report from cache, and then we can consider re-opening this.

PR #1134, opened 2020-02-13
Create test-schemas.tar.gz with maven

relud:
this makes reviewing new test schemas easier, and allows schemas for ingestion-sink and ingestion-beam tests to be defined in their own `src/test/resources`


relud:
> This gets run as part of mvn test?

yeah, it gets run in the `process-test-resources` phase, which depends on the `generate-test-resources` phase and is a prerequisite of the `test` phase

PR #1135, opened 2020-02-13
Bug 1614428 c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0 sending malformed payloads

scholtzan:
In https://github.com/mozilla/gcp-ingestion/pull/1127 scrubbing of messages with client_id `c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0` was implemented. There messages are scrubbed after parsing the payload and schema validation.

In https://bugzilla.mozilla.org/show_bug.cgi?id=1614428 I noticed that these clients send malformed payload. In this specific case `app_build` is not part of `client_info`. In this PR,  message scrubbing and writing them to errors has been moved before schema validation. I also did some refactoring as suggested in https://github.com/mozilla/gcp-ingestion/pull/1127#discussion_r378269519

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -69,6 +69,18 @@ public class MessageScrubber {
             .filter(JsonNode::isTextual) //
             .anyMatch(j -> j.textValue().startsWith("webIsolated="))) {
       handleBug("1562011", ScrubAction.DROP);
+    } else if ("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0".equals(attributes.get(Attribute.CLIENT_ID))
+        || Optional.of(json) // Glean pings: client_info.client_id
+            .map(j -> j.path("client_info").path("client_id").textValue())
+            .filter(s -> s.contains("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0")) //
+            .isPresent()
+        || Optional.of(json) // legacy telemetry pings: client_id
+            .map(j -> j.path("client_id").textValue())
+            .filter(s -> s.contains("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0")) //
+            .isPresent()) {
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1489560
+      // https://bugzilla.mozilla.org/show_bug.cgi?id=1614428
+      handleBug("1489560", ScrubAction.SEND_TO_ERRORS);
     }
   }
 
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -69,6 +69,18 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
             .filter(JsonNode::isTextual) //
             .anyMatch(j -> j.textValue().startsWith("webIsolated="))) {
       handleBug("1562011", ScrubAction.DROP);
+    } else if ("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0".equals(attributes.get(Attribute.CLIENT_ID))
```
relud: We should make this client id a constant, so that we can't misspell it.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -69,6 +69,18 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
...
+        || Optional.of(json) // Glean pings: client_info.client_id
+            .map(j -> j.path("client_info").path("client_id").textValue())
+            .filter(s -> s.contains("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0")) //
+            .isPresent()
+        || Optional.of(json) // legacy telemetry pings: client_id
+            .map(j -> j.path("client_id").textValue())
```
relud:
```suggestion
            .map(j -> j.path(Attribute.CLIENT_ID).textValue())
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -69,6 +69,18 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
             .filter(JsonNode::isTextual) //
             .anyMatch(j -> j.textValue().startsWith("webIsolated="))) {
       handleBug("1562011", ScrubAction.DROP);
+    } else if ("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0".equals(attributes.get(Attribute.CLIENT_ID))
+        || Optional.of(json) // Glean pings: client_info.client_id
+            .map(j -> j.path("client_info").path("client_id").textValue())
```
relud:
```suggestion
            .map(j -> j.path("client_info").path(Attribute.CLIENT_ID).textValue())
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
scholtzan: I put it into a constant and refactored it out into a separate message which makes it hopefully a little more readable

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -69,9 +69,28 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
...
+  private static boolean bug1489560Affected(Map<String, String> attributes, ObjectNode json) {
+    final String affectedClientId = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
+
+    return affectedClientId.equals(attributes.get(Attribute.CLIENT_ID))
+        // glean pings: client_info.client_id
+        || Optional.of(json).map(j -> j.path("client_info").path(Attribute.CLIENT_ID).textValue())
+            .filter(s -> s.contains(affectedClientId)) //
+            .isPresent()
+        // legacy telemetry pings: client_id
+        || Optional.of(json).map(j -> j.path(Attribute.CLIENT_ID).textValue())
```
jklukas: For legacy telemetry pings, the client ID appears as `"clientId"` in camelCase, but this line will only catch `"client_id"` in snake case.

jklukas:
Is this change necessary? Both the static client id and the missing app_build ID are problems with the message; the behavior seems "correct" whichever problem we end up hitting first.

scholtzan:
My thinking was that this would make analysis of validation issues easier. Right now the majority of validation errors in `org_mozilla_fenix.metrics` for `#/client_info` comes from this clientID. So it might be possible that other related issues from other client might stay unnoticed. Also, for investigating problems it might be more convenient to query error tables for just one error_message.

But if we care about validating these pings then this PR can also get reverted.

jklukas:
I agree that it is convenient to have errors marked as affected by the coffee-ID bug rather than the other, but the cost here of maintaining duplicate logic for parsing out clientId seems high. 

My main concern is that I believe the code as currently merged is not correctly parsing `clientId` out of payloads, so I expect we are not actually marking affected payloads.

scholtzan:
Right, there would be duplicate logic. One solution for this could be moving the logic for parsing client ids out of `addAttributesFromPayload` into a separate static method which can then be used for parsing attributes and in the `MessageScrubber`.

Not sure how elegant that is.

jklukas:
> Right, there would be duplicate logic. One solution for this could be moving the logic for parsing client ids out of `addAttributesFromPayload` into a separate static method which can then be used for parsing attributes and in the `MessageScrubber`.
> 
> Not sure how elegant that is?

We probably should factor it out, but I agree it's not going to be very elegant. It would be nice to have separate methods for each of the fields we want to pull out of the payload into attributes, but in practice they're pretty correlated, so we would end up having the same code multiple times to determine whether the ping looks like a desktop "common ping", a mobile telemetry ping, or a glean ping in each of the methods.

PR #1138, opened 2020-02-14
Refactor parsing of client ID from payload

scholtzan:
As discussed in https://github.com/mozilla/gcp-ingestion/pull/1135#issuecomment-586454028 factoring out parsing of client ID would help avoiding duplicate code when scrubbing messages.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -4,6 +4,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Maps;
 import com.google.common.collect.Streams;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
 import com.mozilla.telemetry.transforms.MapElementsWithErrors.MessageShouldBeDroppedException;
@@ -79,16 +80,10 @@ public class MessageScrubber {
   // see bug 1489560
   private static boolean bug1489560Affected(Map<String, String> attributes, ObjectNode json) {
     final String affectedClientId = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
+    Map<String, String> tempAttributes = Maps.newHashMap();
+    ParsePayload.addClientIdFromPayload(tempAttributes, json);
 
-    return affectedClientId.equals(attributes.get(Attribute.CLIENT_ID))
-        // glean pings: client_info.client_id
-        || Optional.of(json).map(j -> j.path("client_info").path(Attribute.CLIENT_ID).textValue())
-            .filter(s -> s.contains(affectedClientId)) //
-            .isPresent()
-        // legacy telemetry pings: client_id
-        || Optional.of(json).map(j -> j.path(Attribute.CLIENT_ID).textValue())
-            .filter(s -> s.contains(affectedClientId)) //
-            .isPresent();
+    return affectedClientId.equals(tempAttributes.get(Attribute.CLIENT_ID));
   }
 
   // See bug 1603487 for discussion of affected versions, etc.
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -214,7 +214,13 @@ public class ParsePayload extends MapElementsWithErrors.ToPubsubMessageFrom<Pubs
         .ifPresent(v -> attributes.put(Attribute.SAMPLE_ID, Long.toString(calculateSampleId(v))));
   }
 
-  /** Extracts the client ID from the payload and adds it to `attributes`. */
+  /**
+   * Extracts the client ID from the payload and adds it to `attributes`.
+   *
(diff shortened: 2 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -79,16 +80,10 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
   // see bug 1489560
   private static boolean bug1489560Affected(Map<String, String> attributes, ObjectNode json) {
     final String affectedClientId = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
+    Map<String, String> tempAttributes = Maps.newHashMap();
+    ParsePayload.addClientIdFromPayload(tempAttributes, json);
```
scholtzan: `addClientIdFromPayload` potentially modifies the json object when normalizing the client ID. I don't think it's a problem here though, but certainly not great to have this side-effect...

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -208,7 +201,38 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
+        .map(ParsePayload::normalizeUuid) //
+        .filter(Objects::nonNull) //
+        .findFirst() //
+        .ifPresent(v -> attributes.put(Attribute.SAMPLE_ID, Long.toString(calculateSampleId(v))));
+  }
+
+  /** Extracts the client ID from the payload and adds it to `attributes`. */
+  public static void addClientIdFromPayload(Map<String, String> attributes, ObjectNode json) {
+    // Try to get glean-style client_info object.
+    JsonNode gleanClientInfo = json.path("client_info");
```
relud: let's also extract this one to a separate `getGleanClientInfo` function. It's short, but it's subject to the same hazard of introducing errors via repeated logic.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -208,7 +201,38 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
           .ifPresent(v -> attributes.put(Attribute.OS_VERSION, v));
     }
 
+    ParsePayload.addClientIdFromPayload(attributes, json);
```
relud: this method doesn't need to be qualified with a class name
```suggestion
    addClientIdFromPayload(attributes, json);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -208,7 +201,38 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
+        // that do not contain a client_id.
+        json.path("impression_id").textValue()) //
+        .map(ParsePayload::normalizeUuid) //
+        .filter(Objects::nonNull) //
+        .findFirst() //
+        .ifPresent(v -> attributes.put(Attribute.SAMPLE_ID, Long.toString(calculateSampleId(v))));
+  }
+
+  /** Extracts the client ID from the payload and adds it to `attributes`. */
+  public static void addClientIdFromPayload(Map<String, String> attributes, ObjectNode json) {
```
relud: I think this should be refactored to return an `Optional<String>` rather than setting it in `attributes`.
```suggestion
  public static Optional<String> getNormalizedClientId(ObjectNode json) {
```
that way `bug1489560Affected` doesn't pull it out of a temporary hashmap or try to modify `gleanClientInfo`.

in `addAttributesFromPayload` the call to `getNormalizedClientId` would need to occur before `if (gleanClientInfo.isObject()) {`, and be used to set client_id within `gleanClientInfo` in the same place as before this PR.

scholtzan: The problem here is that not only the glean ping needs to get potentially updated, but also the common ping or pings sending `clientId`. It's possible to return the client ID, but I don't think modifying the json structure can be avoided without duplicating the logic. I felt like it felt more intuitive to modify the attributes since it's more similar to `addAttributesFromPayload`.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -208,7 +200,38 @@ private void addAttributesFromPayload(Map<String, String> attributes, ObjectNode
...
+        // "impression_id" is a client_id-like identifier used in activity-stream ping
+        // that do not contain a client_id.
+        json.path("impression_id").textValue()) //
+        .map(ParsePayload::normalizeUuid) //
+        .filter(Objects::nonNull) //
+        .findFirst() //
+        .ifPresent(v -> attributes.put(Attribute.SAMPLE_ID, Long.toString(calculateSampleId(v))));
+  }
+
+  /** Extracts the client ID from the payload and adds it to `attributes`. */
```
jklukas: Let's also make it clear in the docstring that we have the side effect of normalizing clientId in place in the payload.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
jklukas: I agree with @scholtzan that having this as a side-effect is the practical thing to do here. I think we can rely on copious documentation to explain that we're doing something weird here and why.

scholtzan:
This should probably get merged after the fix in https://github.com/mozilla/gcp-ingestion/pull/1143 got merged

PR #1142, opened 2020-02-18
Revert "Refactor PubsubMessageToTableRow"

scholtzan:
Reverts mozilla/gcp-ingestion#1139 due to errors in https://circleci.com/gh/mozilla/gcp-ingestion/21386?utm_campaign=workflow-failed&utm_medium=email&utm_source=notification

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -40,20 +40,17 @@ import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import org.apache.beam.sdk.coders.KvCoder;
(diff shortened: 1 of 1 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -55,7 +59,8 @@
  * <p>We also perform some manipulation of the parsed JSON to match details of our table
  * schemas in BigQuery.
  */
-public class PubsubMessageToTableRow {
+public class PubsubMessageToTableRow
+    extends MapElementsWithErrors<PubsubMessage, KV<TableDestination, TableRow>> {
```
jklukas: If we mark this class as `implements Serializable`, that should be sufficient to get the tests passing again. `PTransform` implements Serializable, so removing the subclass here also removed the serializable interface.

scholtzan: Okay, I see. I created a follow-up in https://github.com/mozilla/gcp-ingestion/pull/1143

PR #1154, opened 2020-02-20
Prevent setting unallowed settings for storageapi query input

Ben-Wu:
Query inputs don't allow row restriction and selected fields (compilation error).  This change got query input working.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
@@ -151,7 +151,7 @@ public abstract class Read extends PTransform<PBegin, PCollection<PubsubMessage>
         case QUERY:
           read = read.fromQuery(tableSpec).usingStandardSql();
       }
-      if (method == BigQueryReadMethod.storageapi) {
+      if (source == Source.TABLE && method == BigQueryReadMethod.storageapi) {
         if (rowRestriction != null) {
           read = read.withRowRestriction(rowRestriction);
         }


jklukas:
I have successfully used `inputType=bigquery_query` without running into errors. Were you also specifying `bqReadMethod=direct`?

It looks like this is not documented, but there was an assumption here that if you set `inputType=bigquery_query`, you would use the default "export" readMethod.

With the change in this PR, I believe we are now ignoring the value of `bqReadMethod` if `inputType=bigquery_query`. That's probably fine, but might be even better to catch that case and raise an exception instead to give the user feedback about the invalid combination of options.

jklukas:
I think I'm wrong in the above. It looks like we still set readMethod regardless of whether we're using a query or a table as source.

It's unclear to me, then, why we wouldn't be able to set rowRestriction. Can you post the compilation error you were getting?

Ben-Wu:
Here are the two errors:

`java.lang.IllegalArgumentException: Invalid BigQueryIO.Read: Specifies selected fields, which only applies when reading from a table`

`java.lang.IllegalArgumentException: Invalid BigQueryIO.Read: Specifies row restriction, which only applies when reading from a table`

I didn't see anything in the beam documentation that says these aren't allowed in queries but it does make sense since row restriction and selected fields should be implied by the sql anyway.  

I just needed to do `--bqReadMethod=storageapi` since I was hitting the bq export quota.

Ben-Wu:
Also the errors happen even if these options aren't explicitly set.

PR #1157, opened 2020-02-20
Add custom stackdriver metrics to ingestion-sink

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -362,6 +365,8 @@ public static Output getOutput() {
             .build()),
         getInputCompression(output.env));
     output.env.requireAllVarsUsed();
+    // Setup OpenCensus stackdriver exporter after all measurement views have been registered.
+    StackdriverStatsExporter.createAndRegister();
```
jklukas: Can you add a link here to relevant OpenCensus docs that talk about restrictions on when this should be called?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -312,8 +314,9 @@ private static PubsubMessageToObjectNode getFormat(Env env) {
       case "decoded":
         return PubsubMessageToObjectNode.Decoded.of();
       case "payload":
-        return PubsubMessageToObjectNode.Payload.of(env.getStrings(STRICT_SCHEMA_DOCTYPES, null),
-            env.getString(SCHEMAS_LOCATION), FileInputStream::new);
+        return PubsubMessageToObjectNode.Payload.WithOpenCensusMetrics.of(
+            env.getStrings(STRICT_SCHEMA_DOCTYPES, null), env.getString(SCHEMAS_LOCATION),
+            FileInputStream::new);
```
jklukas: Would it be possible to amend the API so that this becomes:

```
return PubsubMessageToObjectNode.Payload
          .of(env.getStrings(STRICT_SCHEMA_DOCTYPES, null),
              env.getString(SCHEMAS_LOCATION), FileInputStream::new)
          .withOpenCensusMetrics();
```

So we'd construct a `Payload`, and then the `withOpenCensusMetrics()` method would return a `WithOpenCensusMetrics` with the relevant fields copied from the base `Payload` object.

relud: that sounds excellent, will amend

Issue #1136, opened 2020-02-14
Remove schema alias support from ingestion-beam


jklukas:
The concept of schema aliases is now deprecated; we instead generate concrete schemas for each glean app as part of mozilla-schema-generator. The schemaAliasesLocation configuration in ingestion-beam has no effect on the pipeline, and we should remove the option entirely.

Same for the new live sink support in ingestion-sink.

jklukas:
`ingestion-sink` is already alias-free, so no additional work needs to be done there.

PR #1163, opened 2020-02-26
Add missing auto-value dependency

relud:
`auto-value:1.4` is a transitive dependency from `google-cloud-pubsub:1.61.0`, but is used directly by `SchemaAlias` and `SchemaAliasingConfiguration`

`cheker-qual:2.10.0` is a transitive dependency from `guava:28.2-jre`, but is used directly by `DerivedAttributesMap` for `@Nullable` annotation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,6 +42,7 @@
         <!-- The classPath setting is a workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925 -->
         <argLine>-Xmx1024m -XX:MaxPermSize=256m -Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
 
+        <auto-value.version>1.7</auto-value.version>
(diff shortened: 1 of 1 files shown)


jklukas:
Instead of adding these dependencies, we should proceed with ripping out support for schema aliases (https://github.com/mozilla/gcp-ingestion/issues/1136), and I would advocate that we remove the use of `@Nullable` if we're only using it in one place.

jklukas:
See removal of checker-qual in https://github.com/mozilla/gcp-ingestion/pull/1164

jklukas:
It looks like we also explicitly use AutoValue for `WithErrors` and `Deduplicate`, so adding a dep on auto-value may be warranted by that.

jklukas:
We already explicitly declare an auto-value dep for ingestion-beam, though. So with #1165 we should be in a good state.

PR #1165, opened 2020-02-26
Remove schema alias support

### LINKED ISSUE #1136: Remove schema alias support from ingestion-beam

jklukas:
Fixes #1136


jklukas:
cc @whd

PR #1168, opened 2020-02-26
Use bom packages to handle versions for multi-package libraries

relud:
`google-cloud.version` is insufficient for `google-cloud-java`
dependencies, because the packages don't share version numbering.

[google-cloud-java#quickstart](https://github.com/googleapis/google-cloud-java#quickstart)
indicates the correct way to determine `google-cloud-java` dependency
versions is via the `libraries-bom` package from `com.google.cloud`.

In light of that, using Jackson and Beam BOMs seemed appropriate.

This unblocks updating opencensus in ingestion-sink.

This also adds `-DtrimStackTrace=false` to maven test commands, because I found it very helpful in debugging this change.

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -46,26 +46,60 @@
         </repository>
     </repositories>
 
+    <dependencyManagement>
+        <dependencies>
+            <!-- Keep these dependency versions in sync with what is pulled in by beam;
+                 check https://mvnrepository.com/artifact/org.apache.beam/beam-sdks-java-io-google-cloud-platform/2.19.0
+
+                 Does not use libraries-bom to determine versions as of beam 2.19 because the lowest
+                 version available uses google-cloud-bom 0.84.0-alpha, but 0.79.0-alpha is needed to
+                 match what is pulled in by beam
+
+                 Does not use google-cloud-bom 0.79.0-alpha to determine versions as of beam 2.19
+                 because that causes integration tests to fail on dataflow with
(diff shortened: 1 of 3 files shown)


ingestion-beam/pom.xml
```diff
@@ -24,37 +24,66 @@
...
+                 version available uses google-cloud-bom 0.84.0-alpha, but 0.79.0-alpha is needed to
+                 match what is pulled in by beam
+
+                 Does not use google-cloud-bom 0.79.0-alpha to determine versions as of beam 2.19
+                 because that causes integration tests to fail on dataflow with
+                 java.lang.NoSuchMethodError and java.lang.NoClassDefFoundError -->
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>google-cloud-bigquery</artifactId>
+                <version>1.61.0</version>
```
jklukas: Can we keep this `1.61.0` constant something that we define in the parent pom and always reference? Perhaps it should be called `google-cloud.beam-compatible.version` to make it more clear that it's a legacy version we're including for compatibility with beam.

relud: > Can we keep this `1.61.0` constant something that we define in the parent pom and always reference?

These versions are not shared via a property, because they are independently versioned and more recent versions do not match.

The version of `google-cloud-bigquery` in the parent is a placeholder that is overridden in downstream modules, and is only used when testing ingestion-core directly.

I've updated the PR with comments to indicate this.

PR #1169, opened 2020-02-26
Remove ValueProvider usage for storage api options

Ben-Wu:
Remove the ValueProvider so that the storage api can be used in batch jobs, and so that it can be run without explicitly setting the options.  If I'm understanding correctly, this prevents the options from being used in templated jobs but that isn't an issue right now?


jklukas:
>  If I'm understanding correctly, this prevents the options from being used in templated jobs but that isn't an issue right now?

It means that the options have to be specified at template creation time rather than at runtime. So you can still use these options in a template, but the specific values get baked into the template.

jklukas:
cc @scholtzan This is effectively rolling back the changes made in https://github.com/mozilla/gcp-ingestion/pull/1073 as those changes appear to have some unforeseen usability disadvantages.

PR #1170, opened 2020-02-26
Add support for batching loads via pubsub in mixed mode

relud:
This is to make the new live sink stop sending >1K load jobs per day for `crash_v4` pings, because it's technically working but I don't want to depend on exceeding that quota in prod.


relud:
@whd the live sink will need these updates:
- a new pubsub topic & subscription for pending gcs files
- add the new topic as `OUTPUT_TOPIC` to the sink with write permissions
- revoke bigquery load job permissions from the
- a new sink "load" service, running the same `mozilla/ingestion-sink:latest` image, with:
  - the new subscription as `INPUT_SUBSCRIPTION` with read permissions
  - bigquery load job permissions
  - 3 instances and no autoscaling

whd:
I should have the config for this done tomorrow; I'll let you know.

whd:
This config for this should be mostly implemented in https://github.com/mozilla-services/cloudops-infra/commit/6280e36a63f8980ea552570423905be2ab917fcf, but there may be some fixes that need to be applied.

PR #1171, opened 2020-02-26
Allow null records in PubsubMessageToObjectNode

relud:
Fixes [this trace](https://gist.github.com/relud/04959f1de0c097a6706552d18fcf908d) reported to me by @whd via slack on Monday.

I would expect the offending message to retry indefinitely, but I don't see any sign of a message like that in [the subcription](https://console.cloud.google.com/cloudpubsub/subscription/detail/telemetry-decoded.bq-sink-payload?project=moz-fx-data-inge-nonprod-9cc2).

Regardless I think it's appropriate to fix this, because it is a regression from the beam sink's behavior.

### DIFF
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
@@ -344,10 +344,13 @@ public abstract class PubsubMessageToObjectNode implements Function<PubsubMessag
           updateParent(parent, name, processTupleField(jsonFieldName, field.getSubFields(),
               (ArrayNode) value, additionalProperties));
         } else {
-          final ObjectNode props = additionalProperties == null ? null : Json.createObjectNode();
-          transformForBqSchema((ObjectNode) value, field.getSubFields(), props);
-          if (!Json.isNullOrEmpty(props)) {
-            additionalProperties.set(jsonFieldName, props);
+          // Only transform value if it is not null
+          if (!value.isNull()) {
+            final ObjectNode props = additionalProperties == null ? null : Json.createObjectNode();
+            transformForBqSchema((ObjectNode) value, field.getSubFields(), props);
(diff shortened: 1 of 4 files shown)


whd:
The subscription you want is [the structured one](https://console.cloud.google.com/cloudpubsub/subscription/detail/structured-decoded.bq-sink-payload?project=moz-fx-data-inge-nonprod-9cc2) which is showing the expected sadness.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -344,10 +344,13 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
           updateParent(parent, name, processTupleField(jsonFieldName, field.getSubFields(),
               (ArrayNode) value, additionalProperties));
         } else {
-          final ObjectNode props = additionalProperties == null ? null : Json.createObjectNode();
-          transformForBqSchema((ObjectNode) value, field.getSubFields(), props);
-          if (!Json.isNullOrEmpty(props)) {
-            additionalProperties.set(jsonFieldName, props);
+          // Only transform value if it is not null
+          if (!value.isNull()) {
```
jklukas:
```suggestion
          if (value.isObject()) {
```

We might as well test for exactly the cast we need to make.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -344,10 +344,13 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
...
+            final ObjectNode props = additionalProperties == null ? null : Json.createObjectNode();
```
jklukas:
```suggestion
            final ObjectNode props = (additionalProperties == null) ? null : Json.createObjectNode();
```

I find this easier to parse with parentheses.

PR #1172, opened 2020-02-27
Refactor to use MapElements.WithFailures in ParsePayload

scholtzan:
Refactoring `ParsePayload` to use `MapWithFailures` as part of #118

Only `Deduplicate` is left to be refactored.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -23,11 +23,18 @@ public class MessageScrubber {
    * Special exception to signal that a message is affected by a specific bug and should
    * be written to error output.
    */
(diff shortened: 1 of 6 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -23,13 +23,20 @@
...
       super(bugNumber);
     }
   }
 
+  /**
+   * Special exception class that signals that a given message should not be sent
+   * downstream to either success or error output.
+   */
+  public static class MessageShouldBeDroppedException extends RuntimeException {
+  }
```
scholtzan: I moved this out of `MapWithErrors` since `MapWithErrors` will eventually get removed after refactoring everything

scholtzan:
Made the changes sent to me via email.

PR #1175, opened 2020-02-27
Implement proposal for flagging unwanted data

jklukas:
See [the proposal Google
doc](https://docs.google.com/document/d/1VWI-PsD9tFaKVWp5hC75lu65lrtgDEBhO4S7XYvHkqQ/edit#)

We refactor some code here, fill in some documentation,
add a new exception type, and also add a single new case of unwanted data,
fixing https://bugzilla.mozilla.org/show_bug.cgi?id=1612933


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -7,36 +7,18 @@
...
+/**
+ * This class is called in {@link ParsePayload} to check for known signatures of potentially
+ * harmful data. The {@link #scrub(Map, ObjectNode)} method may throw an exception as a signal
+ * to route the message to error output or to be dropped.
+ */
 public class MessageScrubber {
 
-  private enum ScrubAction {
-    DROP, REDACT, SEND_TO_ERRORS
-  }
```
jklukas: We remove this enum and instead build metrics emission into the exceptions, so that the individual cases have explicit `throw` calls. The overall effect feels more legible to me with this change, but I'm definitely open to different opinions.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -45,46 +27,132 @@ public AffectedByBugException(String bugNumber) {
...
-    if (ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
-        && "crash".equals(attributes.get(Attribute.DOCUMENT_TYPE))
-        && "nightly".equals(attributes.get(Attribute.APP_UPDATE_CHANNEL))
-        && "20190719094503".equals(attributes.get(Attribute.APP_BUILD_ID)) //
+
+    String namespace = attributes.get(Attribute.DOCUMENT_NAMESPACE);
+    String docType = attributes.get(Attribute.DOCUMENT_TYPE);
+    String appVersion = attributes.get(Attribute.APP_VERSION);
+    String appUpdateChannel = attributes.get(Attribute.APP_UPDATE_CHANNEL);
+    String appBuildId = attributes.get(Attribute.APP_BUILD_ID);
```
jklukas: We factor out these repeated attributes to make the if conditions more compact. We also reorganize the cases to group them together by exception type that they throw.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -45,46 +27,132 @@ public AffectedByBugException(String bugNumber) {
...
-      // https://bugzilla.mozilla.org/show_bug.cgi?id=1614428
-      handleBug("1489560", ScrubAction.SEND_TO_ERRORS);
+      throw new MessageShouldBeDroppedException("1562011");
+    }
+
+    // Check for unwanted data; these messages aren't thrown out, but this class of errors will be
+    // ignored for most pipeline monitoring.
+    if ("com-turkcell-yaani".equals(namespace)) {
+      throw new UnwantedDataException("1612933");
+    }
```
jklukas: This is the only new case.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -45,46 +27,132 @@ public AffectedByBugException(String bugNumber) {
...
+   */
+  private abstract static class MessageScrubberException extends RuntimeException {
+    MessageScrubberException(String bugNumber) {
+      super(bugNumber);
+      markBugCounter(bugNumber);
     }
   }
 
+  /**
+   * Special exception to signal that a message matches a specific signature that we know we
```
scholtzan: tiny nit

```suggestion
   * Special exception to signal that a message matches a specific signature that we know
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -101,17 +101,16 @@ private ParsePayload(ValueProvider<String> schemasLocation) {
...
+          // redact sensitive fields.
           try {
             MessageScrubber.scrub(attributes, json);
           } catch (MessageShouldBeDroppedException e) {
             // This message should go to no output, so we return an empty list immediately.
             return Collections.emptyList();
           }
 
-          // Potentially mutates the value of json to redact specific fields.
-          MessageScrubber.redact(attributes, json);
```
jklukas: With the refactoring of MessageScrubber and grouping of like actions together, it seems less necessary to have `redact` be a separate method from `scrub`.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
scholtzan: This is definitely quite handy

PR #1176, opened 2020-02-28
Add labels from SinkConfig to custom metrics

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -58,6 +60,10 @@
       STREAMING_BATCH_MAX_BYTES, STREAMING_BATCH_MAX_DELAY, STREAMING_BATCH_MAX_MESSAGES,
       STRICT_SCHEMA_DOCTYPES);
 
+  // Env vars that should be attached to metrics
+  private static final List<String> METRIC_LABELS = ImmutableList.of(BIG_QUERY_OUTPUT_MODE,
+      INPUT_SUBSCRIPTION, OUTPUT_FORMAT);
```
relud: these seemed like missing things that it would be convenient to label metrics with when i was checking that custom metrics were working as expected in stackdriver

PR #1177, opened 2020-02-28
Bug 1595136 - Create job for ip privacy dataset generation

Ben-Wu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1595136

Decoder-like job that hashes client id and ip reading from `payload_bytes_raw`.

Additional files that don't really fit into the repo but should be reviewed can be found here: https://gist.github.com/Ben-Wu/071c79c3b744589a9aff6bca6707c920
This includes hash key generation code, output table schema, and the dataflow launch script.


ingestion-beam/src/main/java/com/mozilla/telemetry/IpPrivacyDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+    Optional.of(pipeline) //
+        .map(p -> p //
+            .apply(options.getInputType().read(options)) //
+            .apply(ParseUri.of()).failuresTo(errorCollections) //
+            .apply("RestrictToMainPings", Filter.by((message) -> {
+              String doctype = message.getAttribute(Attribute.DOCUMENT_TYPE);
+              return doctype != null && doctype.equals("main");
+            })).apply(ParseProxy.of()) //
+            .apply(ParseIp.of()) //
+            .apply(GeoCityLookup.of(options.getGeoCityDatabase(), options.getGeoCityFilter())) //
```
jklukas: This is pointing out some refactoring we should probably do in the future. GeoCityLookup implicitly throws out the IP, and that should potentially be made explicit with its own transform. Definitely not in scope for this PR.

ingestion-beam/src/main/java/com/mozilla/telemetry/IpPrivacyDecoder.java
```diff
@@ -0,0 +1,95 @@
...
+            .apply(DecompressPayload.enabled(options.getDecompressInputPayloads())) //
+            // See discussion in https://github.com/mozilla/gcp-ingestion/issues/776
+            .apply(LimitPayloadSize.toMB(8)).failuresTo(errorCollections) //
```
jklukas: We don't have to worry about payload size limits since we're dropping the payload anyway.

ingestion-beam/src/main/java/com/mozilla/telemetry/IpPrivacyDecoder.java
```diff
@@ -0,0 +1,88 @@
...
+                    .by((message) -> "main".equals(message.getAttribute(Attribute.DOCUMENT_TYPE))))
+            .apply(ParseProxy.of()) //
+            .apply(ParseIp.of()) //
+            .apply(GeoCityLookup.of(options.getGeoCityDatabase(), options.getGeoCityFilter())) //
+            .apply(DecompressPayload.enabled(options.getDecompressInputPayloads())) //
+            .apply(ExtractClientIdAndDropPayload.of()).failuresTo(errorCollections) //
+            .apply(HashClientInfo.of(options.getClientIdHashKey(), options.getClientIpHashKey())) //
+            .apply(NormalizeAttributes.of())) //
+        .map(p -> p //
+            .apply(RemoveAttributes.of()) //
```
jklukas: Looking at https://gist.github.com/Ben-Wu/071c79c3b744589a9aff6bca6707c920#file-launch-dataflow-job I finally get what's going on here; I had made assumptions that were blocking me from properly digesting the approach.

I had assumed that we would use `outputTableRowFormat=payload`, in which case the attributes are all dropped and what we write to BQ is based on the payload. In that scenario, instead of `RemoveAttributes`, we would have a `BuildIpPrivacyPayload` that creates a `Map` of field names to values, and then serializes that to a JSON string as the payload and attaches a null attribute map.

Instead, I see now that you're specifying `outputTableRowFormat=decoded` in your configuration and relying on the values of the attributes. That's a perfectly valid way to go, though I believe we'll need to use `outputTableRowFormat=raw` instead to pass through each attribute directly as a field.

jklukas:
The changes here look good. Well done integrating this in a style consistent with the existing codebase.

ingestion-beam/src/main/java/com/mozilla/telemetry/IpPrivacyDecoder.java
Ben-Wu: Oops, I actually am using `--outputTableRowFormat=raw` in my testing.  I just forgot to update the gist.  Yes raw is needed to pass the attribute as a field.

PR #1178, opened 2020-02-28
Remove `WithErrors`

### LINKED ISSUE #118: Use a standard failure output transform

scholtzan:
I think this closes #118


ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
```diff
@@ -51,7 +51,7 @@ public static PipelineResult run(String[] args) {
    */
   public static PipelineResult run(DecoderOptions.Parsed options) {
     final Pipeline pipeline = Pipeline.create(options);
-    final List<PCollection<PubsubMessage>> errorCollections = new ArrayList<>();
+    final List<PCollection<PubsubMessage>> failureCollections = new ArrayList<>();
```
scholtzan: I thought calling this `failureCollections` is more correct since it is now handling failures

PR #1181, opened 2020-03-02
Add STREAMING_DOCTYPES config to ingestion-sink

relud:
This should be able to mitigate the schema update issue for high volume doctypes by making them non-streaming, if that's not already fixed server-side.

This should also reduce cost vs streaming all doctypes.


relud:
@whd `STREAMING_DOCTYPES` must be a regex pattern that fully matches `namespace/doctype` for all messages that *may* be streamed, for example:
`activity_stream/impression_stats|telemetry/event|messaging_system/.*`

relud:
> @whd `STREAMING_DOCTYPES` must be a regex pattern that fully matches `namespace/doctype` for all messages that _may_ be streamed, for example:
> `activity_stream/impression_stats|telemetry/event|messaging_system/.*`

Actually, on second thought, @jklukas @whd how do you feel about changing this to `STREAMING_TABLES` and making it match the compiled `OUTPUT_TABLE` for each message?

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -269,6 +286,25 @@ Output getOutput(Env env) {
...
             }).thenCompose(v -> v);
+        // Send messages not matched by STREAMING_DOCTYPES directly to fileOutput
+        final Function<PubsubMessage, CompletableFuture<Void>> mixedOutput;
+        if (env.containsKey(STREAMING_DOCTYPES)) {
+          Pattern streamingDoctypes = env.getPattern(STREAMING_DOCTYPES);
+          PubsubMessageToTemplatedString doctypeTemplate = PubsubMessageToTemplatedString
+              .of("${document_namespace}/${document_type}");
+          mixedOutput = message -> {
+            try {
+              if (streamingDoctypes.matcher(doctypeTemplate.apply(message)).matches()) {
```
jklukas: Anywhere we're evaluating a regex on a per-message basis, I think we should be caching the result. Can we introduce a guava cache here?

whd:
Heh, I just noticed what is likely a misconfiguration in our current streaming beam config: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/beam/Jenkinsfile.build#L83 WRT doctype and namespace normalization.

> Actually, on second thought, @jklukas @whd how do you feel about changing this to STREAMING_TABLES and making it match the compiled OUTPUT_TABLE for each message?

I think this makes some sense given the above. STREAMING_TABLES would be unambiguously post normalization.

jklukas:
> Actually, on second thought, @jklukas @whd how do you feel about changing this to STREAMING_TABLES and making it match the compiled OUTPUT_TABLE for each message?

I agree that we need to be unambiguous and clearly support only one of `namespace/docType` or `dataset.table`.

I prefer `/` as a separator, particularly considering the regex where we would have to escape dots as `\.`. Using table names also has the ambiguity of whether the namespace should have a `_live` suffix.

So I think I prefer `namespace/docType`:

    activity-stream/impression-stats|mobile/.*

vs.  `dataset.table`:

    activity_stream\.impression_stats|mobile\..*

But we'd probably have to normalize `_` to `-` before doing the comparison if we use `namespace/docType`.

whd:
> Using table names also has the ambiguity of whether the namespace should have a _live suffix.

This is a good point. I presently inclined to `namespace/docType` as a result, but as long as we're consistent I'm easy.

relud:
I will stick with `namespace/docType`

> But we'd probably have to normalize `_` to `-` before doing the comparison if we use `namespace/docType`.

I think we should re-use `PubsubMessageToTemplatedStringForBigQuery`, so we would be normalizing `-` to `_`

jklukas:
> I think we should re-use PubsubMessageToTemplatedStringForBigQuery, so we would be normalizing - to _

I want the regex to be written in a logically consistent format, so if we're doing `namespace/docType`, the regex should be written with dashes. I believe we could still reuse `PubsubMessageToTemplatedStringForBigQuery` but replace `_` with `-` in the result before passing into the regex.

whd:
There's also old world cases like `untrustedModules` but upstream in schemas we handle that and if we wanted to stream such I think we'd want to use e.g. `telemetry/untrusted-modules` in the config option.

relud:
> I'd like to see the following before we merge:
> 
>     * Caching result of the streaming doctypes regex
>
>     * Documentation of the expected format for STREAMING_DOCTYPES

I moved the implementation to `DocumentTypePredicate`, added a cache, and documented the expected format there.

General documentation of ingestion-sink options does not currently exist, and I need to add that as a separate PR.

>     * A test exercising a more complex regex (one containing both `|` and `*`)

done

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/DocumentTypePredicate.java
```diff
@@ -0,0 +1,48 @@
...
+
+  public static DocumentTypePredicate of(Pattern pattern) {
+    return new DocumentTypePredicate(pattern);
+  }
+
+  @Override
+  public boolean test(PubsubMessage message) {
+    String doctype = template.apply(message);
+    try {
+      return cache.get(doctype, () -> pattern.matcher(doctype.replaceAll("_", "-")).matches());
```
jklukas:
```suggestion
    String docType = template.apply(message);
    try {
      return cache.get(docType, () -> pattern.matcher(docType.replaceAll("_", "-")).matches());
```

PR #1185, opened 2020-03-05
Fix ingestion-sink stackdriver resource

relud:
this should fix the `Unrecognized resource type: .` issue in stage


relud:
cc @whd

relud:
apparently custom labels are not allowed. this is throwing api errors >_<

PR #1187, opened 2020-03-09
Don't wrap exceptions in RuntimeException in ParsePayload

jklukas:
It looks like #1172 led to less interpretable error aggregation for monitoring.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -31,6 +31,7 @@ import org.apache.beam.sdk.metrics.Metrics;
 import org.apache.beam.sdk.options.ValueProvider;
 import org.apache.beam.sdk.transforms.FlatMapElements;
 import org.apache.beam.sdk.transforms.PTransform;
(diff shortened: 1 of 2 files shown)


jklukas:
cc @fbertsch @acmiyaguchi This should fix the issue we discussed in today's platform health check meeting.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -73,114 +74,115 @@ private ParsePayload(ValueProvider<String> schemasLocation) {
...
-          try {
-            json = parseTimed(message.getPayload());
-          } catch (IOException e) {
-            Map<String, String> attrs = schemaStore.docTypeExists(message.getAttributeMap())
-                ? message.getAttributeMap()
-                : null; // null attributes will cause docType to show up as "unknown_doc_type"
-            // in metrics
-            PerDocTypeCounter.inc(attrs, "error_json_parse");
-            PerDocTypeCounter.inc(attrs, "error_submission_bytes", submissionBytes);
-            throw new RuntimeException(e);
```
jklukas: This turns into `throw e`.

Issue #1199, opened 2020-03-17
beam live sink error: com.mozilla.telemetry.ingestion.core.util.BubbleUpException


whd:
Offshoot of https://github.com/mozilla/gcp-ingestion/issues/667, discovered while investigating https://bugzilla.mozilla.org/show_bug.cgi?id=1622977.

Example error file in GCS: gs://moz-fx-data-prod-data/telemetry-decoded_bq-sink/error/2020-03-11/14/error-2020-03-11T14-40-00.000Z-2020-03-11T14-50-00.000Z-0-00059-of-00060.ndjson.gz

Sample of relevant error fields:

    {
        "stack_trace_cause_2": "[java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:184), java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392), java.net.Socket.connect(Socket.java:589), sun.security.ssl.SSLSocketImpl.connect(SSLSocketImpl.java:673), sun.net.NetworkClient.doConnect(NetworkClient.java:175), sun.net.www.http.HttpClient.openServer(HttpClient.java:463), sun.net.www.http.HttpClient.openServer(HttpClient.java:558), sun.net.www.protocol.https.HttpsClient.<init>(HttpsClient.java:264), sun.net.www.protocol.https.HttpsClient.New(HttpsClient.java:367), sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.getNewHttpClient(AbstractDelegateHttpsURLConnection.java:191), sun.net.www.protocol.http.HttpURLConnection.plainConnect0(HttpURLConnection.java:1156), sun.net.www.protocol.http.HttpURLConnection.plainConnect(HttpURLConnection.java:1050), sun.net.www.protocol.https.AbstractDelegateHttpsURLConnection.connect(AbstractDelegateHttpsURLConnection.java:177), sun.net.www.protocol.https.HttpsURLConnectionImp...",
[...]

PR #1200, opened 2020-03-17
Remove BubbleUpException references in beam

scholtzan:
Fixes https://github.com/mozilla/gcp-ingestion/issues/1199


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/KeyByBigQueryTableDestination.java
```diff
@@ -133,10 +133,16 @@ public TableDestination getTableDestination(Map<String, String> attributes) {
               msg = PubsubConstraints.ensureNonNull(msg);
               return KV.of(getTableDestination(msg.getAttributeMap()), msg);
             }).exceptionsInto(TypeDescriptor.of(PubsubMessage.class))
-            .exceptionsVia((WithFailures.ExceptionElement<PubsubMessage> ee) -> FailureMessage.of(
-                KeyByBigQueryTableDestination.class.getSimpleName(), //
-                ee.element(), //
-                ee.exception())));
+            .exceptionsVia((WithFailures.ExceptionElement<PubsubMessage> ee) -> {
+              if (ee.exception().getCause() instanceof ExecutionException
+                  || ee.exception().getCause() instanceof UncheckedIOException) {
```
scholtzan: This is the best I could come up with to check for the exceptions that should be re-thrown. I wrapped checked exceptions into `RuntimeException` since it's not possible to throw checked exceptions.

jklukas: I'd prefer that we convert ExecutionException to UncheckedExecutionException:

```
     } catch (ExecutionException e) {
        throw new UncheckedExecutionException(e.getCause());
    }
```

Then here, we rethrow and catch only `IllegalArgumentException`. In general, I think we should only return a FailureMessage for a distinct list of exception types we expect rather than checking a specific list of exception types to bubble up to the pipeline level as we're doing here.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -202,7 +201,7 @@ private TableRow payloadTableRow(PubsubMessage message, TableReference ref, Stri
           }
         });
       } catch (ExecutionException e) {
-        throw new BubbleUpException(e.getCause());
+        throw new RuntimeException(e);
```
jklukas: I think this should be `UncaughtExecutionException(e.getCause())` and the same below.

scholtzan:
I removed `BubbleUpException` and made changes to use `UncheckedExecutionException` instead.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -555,7 +554,7 @@ private String getAndCacheBqName(String name) {
     try {
       return normalizedNameCache.get(name, () -> convertNameForBq(name));
     } catch (ExecutionException | UncheckedExecutionException e) {
```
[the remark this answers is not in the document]
scholtzan: I looked through the code and made this change for all occurrences.

PR #1213, opened 2020-03-25
Add bqMaxBytesPerPartition parameter to sink

jklukas:
Per Google case 22669857 we found that a recent batch job was failing due to
having too many tables created for `main_v4`. This increases the allowed size
for intermediate tables and exposes that value as a compile-time parameter.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
@@ -105,6 +105,19 @@ public interface SinkOptions extends PipelineOptions {
 
   void setBqTriggeringFrequency(String value);
 
+  // When writing to main_v4 in batch mode, we sometimes see memory exceeded errors for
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java
```diff
@@ -105,6 +105,19 @@
 
   void setBqTriggeringFrequency(String value);
 
+  // When writing to main_v4 in batch mode, we sometimes see memory exceeded errors for
+  // BigQuery load jobs; we have found empirically that limiting the total data size per
+  // load job to 100 GB leads to reliable performance.
```
relud: 500* ?

jklukas: Thank you. Will fix.

jklukas: Actually, I meant to delete this.

Issue #1082, opened 2020-01-24
Limit the Pocket republished stream to a single doctype


jklukas:
I got pinged by @ncloudioj earlier today about a question from Pocket. They've been seeing cost go up for processing the Pub/Sub topic that we republish into their project. It turns out they only care about the `impression-stats` doctype, but their topic is receiving all pings for the `activity-stream` namespace; the items they care about are only about 20% of what we're publishing.

I had initially thought we could just change configuration and publish to the existing topic, but had forgotten that only our per-namepace republishing option allows specifying a full topic path. 

For the time being, we can change to use the existing per-doctype configuration options, but we have to set `--perDocTypeDestination` to be exactly the path to the existing Pocket stream rather than a pattern. We can only do this because the structured republisher currently doesn't have any per-doctype destinations configured; if we find another docType we want to republish from there, we'd be stuck.

Longer-term, we should probably add a new `--perDocTypesDestination` option that would accept a map of topic names to lists of desired doctypes. We should probably deprecate the existing per-doctype option in favor of this; it's more verbose, but more flexible.

cc @relud @whd

jklukas:
The short-term change has been deployed, so we are now publishing only `impression-stats` to the pocket topic.

jklukas:
I think the right long-term design is to introduce a param called `--perDocTypeDestinations` which is parallel to `--perNamespaceDestinations`. The new param would be of type `Map<String, List<String>>` and would be passed in as json (Beam handles parsing json params into the declared types of params).

So our use of the option would become:

```
--perDocTypeDestination='{"pocket-topic-name":["activity-stream/impression-stats"]}'
``` 

Once we implement this option and deploy it, we can transition all existing uses of the old `perDocTypeDestination` param to the new one, then remove the old param from the code completely.

jklukas:
Reopening this since we have not yet actually updated the configuration to use the new option. We can close this once all uses of the deprecated option are gone and the underlying code is removed.

whd:
Configuration has been updated ([e.g.](https://console.cloud.google.com/dataflow/jobs/us-west1/2020-03-31_17_40_37-2856634058451422227?project=moz-fx-data-beam-prod-11f7&organizationId=442341870013)) so removal of deprecated options can probably commence.

PR #1219, opened 2020-03-27
Add --perDocTypeDestinations option

scholtzan:
This attempts to fix https://github.com/mozilla/gcp-ingestion/issues/1082

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/Republisher.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/Republisher.java
@@ -4,6 +4,7 @@ import com.mozilla.telemetry.decoder.Deduplicate;
 import com.mozilla.telemetry.republisher.RandomSampler;
 import com.mozilla.telemetry.republisher.RepublishPerChannel;
 import com.mozilla.telemetry.republisher.RepublishPerDocType;
+import com.mozilla.telemetry.republisher.RepublishPerDocTypeEnabled;
 import com.mozilla.telemetry.republisher.RepublishPerNamespace;
 import com.mozilla.telemetry.republisher.RepublisherOptions;
 import com.mozilla.telemetry.transforms.PubsubConstraints;
@@ -90,6 +91,9 @@ public class Republisher extends Sink {
 
     // Republish to per-docType destinations.
     if (options.getPerDocTypeEnabledList() != null) {
+      // todo: this will be removed
+      decoded.apply(RepublishPerDocTypeEnabled.of(options));
+    } else if (options.getPerDocTypeDestinations() != null) {
       decoded.apply(RepublishPerDocType.of(options));
     }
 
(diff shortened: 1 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublishPerDocTypeEnabled.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.options.ValueProvider.StaticValueProvider;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.Partition;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.PCollectionList;
+import org.apache.beam.sdk.values.PDone;
+
+// Handles deprecated --perDocTypeDestination and --perDocTypeEnabledList options
+public class RepublishPerDocTypeEnabled extends PTransform<PCollection<PubsubMessage>, PDone> {
```
scholtzan: This class is the old `RepublishPerDocType`, I just renamed it for backwards compatibility. Once we decide to remove the `--perDocTypeDestination` and `--perDocTypeEnabledList` options entirely this class can simply be removed.

ingestion-beam/src/main/java/com/mozilla/telemetry/Republisher.java
```diff
@@ -90,6 +91,9 @@ public static PipelineResult run(RepublisherOptions.Parsed options) {
 
     // Republish to per-docType destinations.
     if (options.getPerDocTypeEnabledList() != null) {
+      // todo: this will be removed
+      decoded.apply(RepublishPerDocTypeEnabled.of(options));
```
scholtzan: This will be removed once `--perDocTypeDestination` and `--perDocTypeEnabledList` are removed.

ingestion-beam/src/main/java/com/mozilla/telemetry/republisher/RepublishPerDocType.java
```diff
@@ -22,9 +22,11 @@ public static RepublishPerDocType of(RepublisherOptions baseOptions) {
 
   @Override
   public PDone expand(PCollection<PubsubMessage> input) {
-    List<Destination> destinations = baseOptions.getPerDocTypeEnabledList().stream()//
-        .map(Destination::new) //
+    List<Destination> destinations = baseOptions.getPerDocTypeDestinations().entrySet().stream()
+        .flatMap(
+            entry -> entry.getValue().stream().map(value -> new Destination(entry.getKey(), value)))
```
jklukas: This looks correct and in line with the existing implementation. Reading through this class again, though, I realize we should probably be doing something more efficient.

Currently, we create a separate pcollection for each doctype appearing in the values; that would mean potentially we would be creating multiple sinks to the same Pub/Sub topic for multiple different docTypes. Ideally, we should be creating one pcollection per output topic. I'm also realizing we should probably support allowing a single docType to appear in multiple output topics, which would mean not using `Partition`.

These are out of scope here. If you'd be willing to write these up in a follow-up issue, though, I'd appreciate it. This work does not seem worth prioritizing until we have use cases where it actually matters.

PR #1222, opened 2020-03-30
Mark more namespaces and non-Firefox apps as unwanted data

scholtzan:
This marks `default-browser-agent` and `META-INF` namespaces as well as some non-Firefox apps as unwanted data:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1626020
* https://bugzilla.mozilla.org/show_bug.cgi?id=1626022
* https://bugzilla.mozilla.org/show_bug.cgi?id=1592010

There is a [query to get all non-Firefox apps](https://bugzilla.mozilla.org/show_bug.cgi?id=1592010) which returns quite a long list. I only added the ones with more than 2000 pings in the past 10 days. Not sure what the threshold should be here.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -8,8 +8,11 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Streams;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
+import java.util.Arrays;
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Stream;
 import org.apache.beam.sdk.metrics.Metrics;
 
@@ -26,26 +29,6 @@ public class MessageScrubber {
       .put("org-mozilla-vrbrowser-dev", "1614410") //
       .put("org-mozilla-fenix-performancetest", "1614412") //
       .put("org-mozilla-vrbrowser-wavevr", "1614411") //
-      .put("default-browser-agent", "1626020") //
-      .put("META-INF", "1626022") //
-      .build();
-
-  private static final Map<String, String> IGNORED_APPS = ImmutableMap.<String, String>builder()
-      .put("FirefoxOS", "1618684") //
-      .put("Ordissimo", "1592010") //
-      .put("adloops", "1592010") //
-      .put("agendissimo", "1592010") //
-      .put("ZeroWeb", "1592010") //
-      .put("CoreApp", "1592010") //
-      .put("webissimo3", "1592010") //
-      .put("Bitcentral%20Core", "1592010") //
-      .put("Zotero", "1592010") //
-      .put("adloops", "1592010") //
-      .put("facebook", "1592010") //
-      .put("vkb_browser", "1592010") //
-      .put("maps", "1592010") //
-      .put("gmail", "1592010") //
-      .put("selfcheckKiosk", "1592010") //
       .build();
 
   /**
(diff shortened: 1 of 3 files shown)


fbertsch:
Hold on, we don't want to merge `default-browser-agent` until we know what it is.

jklukas:
Also, do we know where `META-INF` is coming from? I'm used to that string as something linked to Java package structure.

scholtzan:
I'll do some investigation

fbertsch:
> There is a [query to get all non-Firefox apps](https://bugzilla.mozilla.org/show_bug.cgi?id=1592010) which returns quite a long list. I only added the ones with more than 2000 pings in the past 10 days. Not sure what the threshold should be here.

Perhaps we should add a monitor or alert for this.

relud:
> > There is a [query to get all non-Firefox apps](https://bugzilla.mozilla.org/show_bug.cgi?id=1592010) which returns quite a long list. I only added the ones with more than 2000 pings in the past 10 days. Not sure what the threshold should be here.
> 
> Perhaps we should add a monitor or alert for this.

Perhaps we should make this an allow-list, instead of an ignore-list? i.e.
```java
if (ParseUri.TELEMETRY.equals(namespace)
    && Set("event", "main", "modules").contains(docType)
    && !"Firefox".equals(appName)) {
  throw ...
}
```
Note: `event`, `main`, and `modules` are the only document types aggregated in the linked query.

fbertsch:
> Perhaps we should make this an allow-list, instead of an ignore-list? i.e.

That makes sense. We could do what we do for mobile and make it a normalized field.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -88,8 +82,15 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
...
-      throw new UnwantedDataException(IGNORED_APPS.get(appName));
+    if ("FirefoxOS".equals(appName)) {
+      throw new UnwantedDataException("1618684");
+    }
+
+    // event, main, and modules are the only document types that will get aggregated
+    Set<String> docTypes = new HashSet<>(Arrays.asList("event", "main", "modules"));
+    if (ParseUri.TELEMETRY.equals(namespace) && docTypes.contains(docType)
+        && !"Firefox".equals(appName)) {
+      throw new UnwantedDataException("1592010");
```
scholtzan: > Perhaps we should make this an allow-list, instead of an ignore-list? i.e.

Like this?

scholtzan:
`default-browser-agent` was recently introduced in https://bugzilla.mozilla.org/show_bug.cgi?id=1602463 and lives in the code here: https://searchfox.org/mozilla-central/source/toolkit/mozapps/defaultagent/Telemetry.cpp#28

`WEB-INF` looks like it's a bug. The submission URI is malformed, see https://bugzilla.mozilla.org/show_bug.cgi?id=1626022 I removed it for now from getting ignored.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -88,8 +82,15 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
...
     }
 
-    if (IGNORED_APPS.containsKey(appName)) {
-      throw new UnwantedDataException(IGNORED_APPS.get(appName));
+    if ("FirefoxOS".equals(appName)) {
+      throw new UnwantedDataException("1618684");
+    }
+
+    // event, main, and modules are the only document types that will get aggregated
+    Set<String> docTypes = new HashSet<>(Arrays.asList("event", "main", "modules"));
```
relud: This should move to a static variable, named something like `FIREFOX_ONLY_DOCTYPES`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -88,8 +82,15 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       throw new UnwantedDataException(IGNORED_NAMESPACES.get(namespace));
     }
 
-    if (IGNORED_APPS.containsKey(appName)) {
-      throw new UnwantedDataException(IGNORED_APPS.get(appName));
+    if ("FirefoxOS".equals(appName)) {
+      throw new UnwantedDataException("1618684");
+    }
+
+    // event, main, and modules are the only document types that will get aggregated
```
relud:
```suggestion
    // These document types receive a significant number of pings with malformed `build_id`s due to third-party builds where `appName != "Firefox"`
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
relud: yes 👍

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -31,6 +31,9 @@
       .put("org-mozilla-vrbrowser-wavevr", "1614411") //
       .build();
 
+  private static final Set<String> FIREFOX_ONLY_DOCTYPES = new HashSet<>(
+      Arrays.asList("event", "main", "modules"));
```
jklukas: I'd prefer a Guava ImmutableSet, which would prevent accidental changes and can be constructed a bit more elegantly.

PR #1228, opened 2020-04-02
Allow GC to clean up PubsubMessage before calling AckReplyConsumer

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Pubsub.java
```diff
@@ -41,15 +60,7 @@ private Pubsub() {
...
-              .whenCompleteAsync((result, exception) -> {
-                if (exception == null) {
-                  consumer.ack();
-                } else {
-                  // exception is always a CompletionException caused by the real exception
-                  LOG.warn("Exception while attempting to deliver message", exception.getCause());
-                  consumer.nack();
-                }
-              })))
+              .whenCompleteAsync(asCallback(consumer))))
```
relud, line 63: @akkomar or @acmiyaguchi do you know if having a lambda created in a context where `message` is available would be enough for the jvm to keep it around, or does the body need to contain an actual reference to it?

acmiyaguchi: I did a bit of googling about closure rules in Java. Intuitively, I would think the GC would be able to trace the variables it would need to reference later and disregard the other variables like the message. But in this case, the inner lambda probably holds a reference to the outer lambda to access the consumer, which may keep the message around. This is mostly conjecture, though. Do you have a way of measuring memory usage to see if this works empirically?

* https://stackoverflow.com/questions/47043967/java-8-lambda-closure-and-gc
* https://stackoverflow.com/questions/28446626/do-java8-lambdas-maintain-a-reference-to-their-enclosing-instance-like-anonymous/28447015#28447015

relud: I'm working on setting up a profiler while running, which might give me the information I need to determine if this helps.

akkomar: Memory-wise, I don't see this diff changing anything. I would think message would be available for collection since it's used only in the first stage of CompletableFuture chain. +1 for profiling though.

relud: When I checked with the IntelliJ debugger, it looks like this won't impact memory. closing invalid

PR #1233, opened 2020-04-06
Bug 1625330 - Reduce DEFAULT_BATCH_MAX_DELAY to avoid OOM errors

relud:
see [Bug 1625330 comment 2](https://bugzilla.mozilla.org/show_bug.cgi?id=1625330#c2) for why this is likely worth doing.

I will test this in stage by setting `BATCH_MAX_DELAY=10s` and `BIG_QUERY_OUTPUT_MODE=file_loads`, and seeing if we hit OOM errors.

### DIFF
--- a/ingestion-sink/Dockerfile
+++ b/ingestion-sink/Dockerfile
@@ -14,4 +14,4 @@ RUN mvn package
 
 FROM base
 COPY --from=build /app/ingestion-sink/target/*.jar /app/ingestion-sink/target/
-CMD exec java -jar /app/ingestion-sink/target/*with-dependencies.jar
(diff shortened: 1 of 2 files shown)


relud:
update:

- added `$JAVA_OPTS` to docker command, so that we can do like `-Xmx 3584m` in kubernetes to ensure that java will more of the memory available to it.
- removed `getDefaultMaxOutstandingElementCount` and `getDefaultMaxOutstandingRequestBytes` from gcs and bq file loads, because the default limits are better for avoiding OOM errors

relud:
> we're no longer hitting OOM errors.

as of 23:38 PDT we are hitting OOM errors again. I'm attempting to mitigate this by reducing max outstanding messages.

PR #1235, opened 2020-04-07
Bug 1605199 - Forward deletion pings to amplitude

relud:
This contains all of the boilerplate for setting up a new ingestion-sink output that writes client ids from deletion requests to amplitude.

Search for `TODO` to find all of the spots that need to be filled out with the actual implementation of sending a deletion request to amplitude.


relud:
running this as a streaming sink is overkill, because the job should only take a few minutes per day, so closing this in favor of https://github.com/mozilla/bigquery-etl/pull/1082

PR #1236, opened 2020-04-08
Bug 1626301 Allow "Solaris" to normalize to "Linux"

jklukas:
Glean may report "Solaris" as an OS value whereas we currently only recognize "SunOS" from desktop telemetry.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
@@ -61,7 +61,8 @@ public class NormalizeAttributes
       return WINDOWS;
     } else if (name.startsWith("Darwin")) {
       return MAC;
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/NormalizeAttributes.java
```diff
@@ -61,7 +61,8 @@ public static String normalizeOs(String name) {
       return WINDOWS;
     } else if (name.startsWith("Darwin")) {
       return MAC;
-    } else if (name.contains("Linux") || name.contains("BSD") || name.contains("SunOS")) {
+    } else if (name.contains("Linux") || name.contains("BSD") || name.contains("SunOS")
```
jbeich, line 64: FWIW, DragonFly which was forked from FreeBSD ~15 years ago doesn't contain BSD in system name.

PR #1237, opened 2020-04-08
Close batches from a synchronized context

relud:
to prevent `write` after `close`

Without this we could lose messages, e.g. as seen in [bug 1612371 comment 14](https://bugzilla.mozilla.org/show_bug.cgi?id=1612371#c14), but only when writing to GCS, because that's the only implementation of `BatchWrite` that didn't have `synchronized` on `close`.


relud:
cc @whd

PR #1238, opened 2020-04-09
Limit Gcs.Write to single storage.create api call

relud:
to remove any chance of partial write complications and minimize the number of api calls.

requires #1237 to prevent data loss by ensuring that `content.write` is never called after `content.toByteArray()`

### DIFF
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Gcs.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Gcs.java
@@ -1,6 +1,5 @@
 package com.mozilla.telemetry.ingestion.sink.io;
 
-import com.google.cloud.WriteChannel;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
 import com.google.cloud.storage.Storage;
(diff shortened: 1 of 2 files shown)


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Gcs.java
```diff
@@ -31,6 +30,8 @@ private Gcs() {
 
     public static class Ndjson extends Write {
 
+      private static final byte[] newline = "\n".getBytes(StandardCharsets.UTF_8);
```
jklukas, line 33:
```suggestion
      private static final byte[] NEWLINE = "\n".getBytes(StandardCharsets.UTF_8);
```

relud:
> Am I correct with context here that you've determined the GCS SDK is not releasing memory as we make `.write` calls, so there's no memory benefit to incrementally pushing data into the SDK, so we might as well do the simpler thing of accumulating a byte stream and flushing it to GCS all at once?

yes, but the reason is that with BATCH_MAX_DELAY=10s the sum of in-flight batches is at most around 30MiB uncompressed (not counting any copies lingering from concatenating byte strings).

PR #1239, opened 2020-04-09
Bug 1628539 - Add transform for decrypting pioneer v2 pings

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1628539), [design doc](https://docs.google.com/document/d/1WXrEjnlCNqXsWsd7T5EHrkR4f1HXvSBfvKn23DREs98/edit?usp=sharing)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/PioneerDecryptor.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.beam.sdk.transforms.ProcessFunction;
+import org.apache.beam.sdk.transforms.WithFailures;
+import org.apache.beam.sdk.transforms.WithFailures.Result;
+import org.apache.beam.sdk.values.PCollection;
+import org.apache.beam.sdk.values.TypeDescriptor;
+
+public class PioneerDecryptor extends
+    PTransform<PCollection<PubsubMessage>, Result<PCollection<PubsubMessage>, PubsubMessage>> {
+
+  public static PioneerDecryptor of() {
```
jklukas, line 21: Note on naming that either transforms or their factory methods should be verbs, so this could be `DecryptPioneerPayloads.of` or perhaps `Pioneer.decryptPayloads()` if you ended up wanting to package more things under a single class.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,104 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import org.apache.commons.io.IOUtils;
+import org.everit.json.schema.Schema;
+import org.everit.json.schema.loader.SchemaLoader;
+import org.jose4j.jwk.PublicJsonWebKey;
+import org.jose4j.lang.JoseException;
+import org.json.JSONArray;
+import org.json.JSONObject;
+import org.json.JSONTokener;
```
acmiyaguchi, line 19: @jklukas Do you have insight into the following error?

```
[ERROR] /root/project/ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java:18:1: Illegal import - org.json.JSONObject. [IllegalImport]
```

Is this because `org.json` is not actually actually imported within the beam project? Do you think it's okay to pull in that dependency here?

jklukas: I believe this is coming from checkstyle. We specifically prohibit importing `org.json` so that we don't spread out JSON manipulation logic between two different formats. We should be doing all JSON manipulation with Jackson, and org.json should only come into play specifically for doing schema validation.

See https://github.com/mozilla/gcp-ingestion/blob/master/checkstyle/suppressions.xml#L15-L17 and https://github.com/mozilla/gcp-ingestion/blob/master/checkstyle/checks.xml#L68-L70

acmiyaguchi: I'm using this to load a schema for a configuration file, and then to read from that configuration file into the main data structure. This doesn't fit neatly within core, because of the `jose4j` dependency and the testing resources in the `pioneer` folder.

acmiyaguchi: Found the spot to put the checkstyle suppression -- it'd be necessary no matter what package it's part of.

acmiyaguchi:
I'm going to close this in favor of a new PR that does not come from a forked PR.

PR #1243, opened 2020-04-13
Always provide executor to CompletableFuture::*Async

relud:
and use a non-blocking timeout for BatchWrite.

### DIFF
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
@@ -23,6 +23,8 @@ import java.io.FileInputStream;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.Executor;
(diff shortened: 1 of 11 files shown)


relud:
cc @whd this dramatically speeds up the sink, and appears to be making the batch sink run well enough for 1-minute batches.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -85,7 +91,7 @@
   // to have a total pipeline delay (including edge and decoder) of less than 1 hour, and ideally
   // less than 10 minutes, so this plus DEFAULT_LOAD_MAX_DELAY should be less than 10 minutes.
   // Messages may be kept in memory until they ack or nack, so too much delay can cause OOM errors.
-  private static final String DEFAULT_BATCH_MAX_DELAY = "10s"; // 10 seconds
+  private static final String DEFAULT_BATCH_MAX_DELAY = "1m"; // 1 minute
```
relud, line 94: @whd for this to work the terraform config needs to be updated to set `JAVA_OPTS=-Xmx1536m` so that we have enough memory.

PR #1244, opened 2020-04-14
Remove nulls from arrays in beam PubsubMessageToTableRow

jklukas:
In our reproduction of the pipeline instability incident in stage,
Google support found that we had some batch load jobs continually failing
due to loading nulls into REPEATED fields.

We were able to trace down a load job into crash_v4 where the field
`payload.strack_traces.modules` had some entries that were literal `null`.

In this PR, we strip out nulls from the JSON array. We need to change how we
handle additional_properties, though, in order to maintain order in the array.
Previously, we used `null` as a placeholder value for entries where the full
content was parsed into fields; now, we use `{}` instead so that we can
distinguish that case from literal `null` values appearing in the input array.

It appears that ingestion-sink must already be correctly stripping these out.
We should make sure we fully understand the behavior there along with this
change.

cc @whd

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
@@ -328,15 +328,6 @@ public class PubsubMessageToTableRow implements Serializable {
       List<Object> records = value.filter(List.class::isInstance).map(List.class::cast)
           .orElse(ImmutableList.of());
       List<Object> repeatedAdditionalProperties = new ArrayList<>();
-      records.stream().filter(Map.class::isInstance).map(Map.class::cast).forEach(record -> {
-        Map<String, Object> props = additionalProperties == null ? null : new HashMap<>();
-        transformForBqSchema(record, field.getSubFields(), props);
-        if (props != null && !props.isEmpty()) {
-          repeatedAdditionalProperties.add(props);
-        } else {
-          repeatedAdditionalProperties.add(null);
-        }
-      });
       // Arrays of tuples cannot be transformed in place, instead each element of the parent array
       // will need to reference a new transformed object.
       if (records.stream().allMatch(List.class::isInstance)) {
@@ -348,13 +339,32 @@ public class PubsubMessageToTableRow implements Serializable {
           if (props != null && !props.isEmpty()) {
             repeatedAdditionalProperties.add(props);
           } else {
-            repeatedAdditionalProperties.add(null);
+            repeatedAdditionalProperties.add(Collections.emptyMap());
           }
           records.set(i, (Object) m);
         }
+      } else {
+        records.forEach(untypedRecord -> {
+          if (untypedRecord instanceof Map) {
+            Map<String, Object> record = (Map<String, Object>) untypedRecord;
+            Map<String, Object> props = additionalProperties == null ? null : new HashMap<>();
+            transformForBqSchema(record, field.getSubFields(), props);
+            if (props != null && !props.isEmpty()) {
+              repeatedAdditionalProperties.add(props);
+            } else {
+              repeatedAdditionalProperties.add(Collections.emptyMap());
+            }
+          } else {
+            repeatedAdditionalProperties.add(null);
+          }
+        });
+        // BigQuery cannot load null values into an array, so we must filter them out here, but
+        // additionalProperties will have a null in the correct place in the array.
+        parent.put(name,
+            records.stream().filter(Map.class::isInstance).collect(Collectors.toList()));
       }
 
-      if (!repeatedAdditionalProperties.stream().allMatch(Objects::isNull)) {
+      if (!repeatedAdditionalProperties.stream().allMatch(m -> Collections.emptyMap().equals(m))) {
         additionalProperties.put(jsonFieldName, repeatedAdditionalProperties);
       }
       // If we've made it here, we have a basic type or a list of basic types.
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRowTest.java
@@ -394,7 +394,7 @@ public class PubsubMessageToTableRowTest extends TestWithDeterministicJson {
     TRANSFORM.transformForBqSchema(parent, bqFields, additionalProperties);
     assertEquals(expected, Json.asString(parent));
 
-    String expectedAdditional = "{\"payload\":[null,{\"b\":22},null]}";
+    String expectedAdditional = "{\"payload\":[{},{\"b\":22},{}]}";
     assertEquals(expectedAdditional, Json.asString(additionalProperties));
   }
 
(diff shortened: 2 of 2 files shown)


jklukas:
The particular document we've identified as being pathological for the beam sink is:

```
"submission_timestamp":"2020-04-08T06:16:26.493189Z",
"document_id":"92ed0b15-57be-4b38-a71c-d52f7454dd8f"
```

This document made it into stable tables, but it looks like the entirety of `payload.stack_traces` went into additional_properties.

jklukas:
It looks like [`stackTraces` was only added to the crash ping schema in the past month](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/2f99df573d0a520fd32e658f1ab138febd121f64), so it likely was only deployed _after_ the incident was over, and that would explain why we're only seeing this behavior now.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -361,22 +362,30 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
...
             } else {
-              repeatedAdditionalProperties.addNull();
+              repeatedAdditionalProperties.addObject();
             }
           }
         } else {
+          ArrayNode filteredValue = Json.createArrayNode();
           for (JsonNode record : value) {
             final ObjectNode props = additionalProperties == null ? null : Json.createObjectNode();
-            transformForBqSchema((ObjectNode) record, field.getSubFields(), props);
```
jklukas, line 370: Based on my understanding of the existing code, nulls within object arrays should hit this line and throw an exception since `record` will be `NullNode` rather than `ObjectNode`. Is it possible that these are currently going to error output?

whd:
> It looks like [`stackTraces` was only added to the crash ping schema in the past month](https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/2f99df573d0a520fd32e658f1ab138febd121f64), so it likely was only deployed _after_ the incident was over, and that would explain why we're only seeing this behavior now.
> 
> This field only started seeing values on 2020-04-11.

This corresponds to https://github.com/mozilla-services/mozilla-pipeline-schemas/commit/317d115dea0be30999863d52ca7ecaa88a473808. During the instability period we were indeed pinned to `202003140150_467991a` so this was introduced after we had rolled out the new sink, which explicitly handles this case differently: https://bugzilla.mozilla.org/show_bug.cgi?id=1612372#c1.

jklukas:
> this was introduced after we had rolled out the new sink, which explicitly handles this case differently: https://bugzilla.mozilla.org/show_bug.cgi?id=1612372#c1.

I don't _think_ the difference that @relud mentions there is relevant to this particular problem. AFAICT, both sinks are currently vulnerable to producing nulls in arrays.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
relud: there is no error output for the new sink. it will instead nack the message in pubsub, causing the sink to retry indefinitely, unless the error is transient or gets fixed in an ingestion-sink update.

ingestion-sink/src/test/resources/testdata/payload-format-expected.ndjson
```diff
@@ -13,8 +13,9 @@
 {"additional_properties":{"missingField":null}}
 {"additional_properties":{"test_map_without_value":{"key":"value"}},"test_map":[{"key":"key","value":"value"}],"test_map_without_value":[{"key":"key"}]}
 {"additional_properties":{},"test_repeated_record":[{"field":"value"}],"test_record":{"key":"value"}}
-{"additional_properties":{"test_repeated_record":[null,{"reject":"value","alsoReject":null},null]},"test_repeated_record":[{"field":"value"},{},{}]}
-{"additional_properties":{"test_repeated_record":[{"reject":"value"},null,{"reject":"value"}]},"test_repeated_record":[{},{"field":"value"},{}]}
+{"additional_properties":{"test_repeated_record":[{},{"reject":"value","alsoReject":null},{}]},"test_repeated_record":[{"field":"value"},{},{}]}
+{"additional_properties":{"test_repeated_record":[{"reject":"value"},{},{"reject":"value"}]},"test_repeated_record":[{},{"field":"value"},{}]}
+{"additional_properties":{"test_repeated_record":[null,{}]},"test_repeated_record":[{"field":"value"}]}
```
relud, line 18: I don't think we should ever remove an entry from a list. Instead I think we should insert an empty object where possible, maybe an empty array if all of the values are empty objects, and otherwise send the whole list to `additional_properties`.
```suggestion
{"additional_properties":{"test_repeated_record":[null,{}]},"test_repeated_record":[{},{"field":"value"}]}
```

whd:
> I don't _think_ the difference that @relud mentions there is relevant to this particular problem. AFAICT, both sinks are currently vulnerable to producing nulls in arrays.

Do we have a sense of why we aren't seeing this issue with the kubernetes sinks currently?

jklukas:
> Do we have a sense of why we aren't seeing this issue with the kubernetes sinks currently?

I don't have an explanation for that, nor do I have hard evidence it's _not_ happening in the kube sinks. My reading of the current kube sink is that a row that encountered this problem would throw a WriteErrors exception, resulting in a warn-level log message, and then nack the message.

whd:
> My reading of the current kube sink is that a row that encountered this problem would throw a WriteErrors exception, resulting in a warn-level log message, and then nack the message.

I believe NACKing would eventually result in unbounded oldest unacked message age, which we're not seeing.

jklukas:
Without the code changes here, the new test case introduced in this PR causes the ingestion-sink tests to raise:

```
java.lang.ClassCastException: com.fasterxml.jackson.databind.node.NullNode cannot be cast to com.fasterxml.jackson.databind.node.ObjectNode
	at com.mozilla.telemetry.ingestion.sink.transform.PubsubMessageToObjectNodeTest.canFormatAsPayload(PubsubMessageToObjectNodeTest.java:236)
```

I am trying to see if I can fully trace how exceptions are supposed to bubble up here.

relud:
> My reading of the current kube sink is that a row that encountered this problem would throw a WriteErrors exception, resulting in a warn-level log message, and then nack the message.

this is the intended behavior in ingestion-sink

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -361,22 +362,31 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
...
+              if (!Json.isNullOrEmpty(props)) {
+                repeatedAdditionalProperties.add(props);
+              } else {
+                repeatedAdditionalProperties.addObject();
+              }
             } else {
+              // BigQuery cannot load null values into an array, so we insert an empty object
+              // instead.
+              filteredValue.addObject();
               repeatedAdditionalProperties.addNull();
```
relud, line 384: I think this should preserve `record`, in case it's ever anything but `null` here.
```suggestion
              repeatedAdditionalProperties.add(record);
```

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToObjectNode.java
```diff
@@ -361,22 +362,31 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
...
+              filteredValue.add(record);
+              transformForBqSchema((ObjectNode) record, field.getSubFields(), props);
+              if (!Json.isNullOrEmpty(props)) {
+                repeatedAdditionalProperties.add(props);
+              } else {
+                repeatedAdditionalProperties.addObject();
+              }
             } else {
+              // BigQuery cannot load null values into an array, so we insert an empty object
+              // instead.
```
relud, line 382: I think this comment should reflect that we have only tested that record is not an object
```suggestion
              // BigQuery only allows objects in this array, so we insert an empty object instead.
```

PR #1246, opened 2020-04-15
Bug 1628539 - Add transform for decrypting pioneer v2 pings

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1628539), [design doc](https://docs.google.com/document/d/1WXrEjnlCNqXsWsd7T5EHrkR4f1HXvSBfvKn23DREs98/edit?usp=sharing)

This is a non-forked version of #1239 which can run integration tests for KMS.


ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,104 @@
...
+    return keys.size();
+  }
+
+  private void loadAllKeys() throws IOException, ValidationException {
+    final Map<String, PrivateKey> tempKeys = new HashMap<>();
+
+    Schema schema;
+    try (InputStream inputStream = Resources.getResource("keystore-metadata.schema.json")
+        .openStream()) {
+      JSONObject schemaBytes = new JSONObject(new JSONTokener(inputStream));
```
jklukas, line 62: We should factor all json parsing logic into ingestion-beam's `Json.java`. In this case, this call would become `Json.readValue` and we'd add a new variant of that function that takes in an InputStream. That way, we're consistently using Jackson with the JsonOrgModule to create JSONObject instances.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,104 @@
...
+      schema = SchemaLoader.load(schemaBytes);
+    } catch (IOException e) {
+      throw new IOException("Exception thrown while reading metadata file");
```
jklukas, line 65:
```suggestion
      throw new IOException("Exception thrown while reading metadata file", e);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,104 @@
...
+      String privateKeyUri = entry.getString("private_key_uri");
+      try (InputStream inputStream = BeamFileInputStream.open(privateKeyUri)) {
+        String serializedKey = new String(IOUtils.toByteArray(inputStream));
+        PublicJsonWebKey key = PublicJsonWebKey.Factory.newPublicJwk(serializedKey);
+        tempKeys.put(namespace, key.getPrivateKey());
+      } catch (IOException e) {
+        // TODO: interpolate specific information from the metadata.
+        throw new IOException("Exception thrown while reading key specified by metadata.");
+      } catch (JoseException e) {
+        throw new RuntimeException(e);
```
jklukas, line 88: UncheckedIOException would be more descriptive here.

ingestion-beam/pom.xml
```diff
@@ -60,6 +60,11 @@
             <artifactId>beam-sdks-java-io-google-cloud-platform</artifactId>
             <version>${beam.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-kms</artifactId>
+            <version>0.81.0-beta</version>
```
jklukas, line 66: Is it possible to keep this in step with `${google-cloud.version}`? If not, we should add a comment to explain why.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -0,0 +1,76 @@
...
+                ee.element(), //
+                ee.exception());
+          }
+        }));
+  }
+
+  private class Fn implements ProcessFunction<PubsubMessage, Iterable<PubsubMessage>> {
+
+    @Override
+    public Iterable<PubsubMessage> apply(PubsubMessage message) throws IOException, JoseException {
```
jklukas, line 57: I believe JoseException is redundant here since it inherits from IOException

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -0,0 +1,76 @@
...
+      message = PubsubConstraints.ensureNonNull(message);
+
+      if (keyStore == null) {
+        keyStore = KeyStore.of(metadataLocation.get());
+      }
+
+      // TODO: count per doctype errors
+      ObjectNode json = Json.readObjectNode(message.getPayload());
```
jklukas, line 65: We should probably makes ParsePayload's `parseTimed` method into a static method so that we can call that here.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -0,0 +1,76 @@
...
+
+      PrivateKey key = keyStore.getKey(message.getAttribute("document_namespace"));
```
jklukas, line 67:
```suggestion
      PrivateKey key = keyStore.getKey(message.getAttribute(Attributes.DOCUMENT_NAMESPACE));
```

ingestion-beam/pom.xml
acmiyaguchi: The latest [stable version is 1.38.0](https://cloud.google.com/kms/docs/reference/libraries#installing_the_client_library), and the `google-cloud.version` is pinned to the beam dependency's version of 1.61.0. I ran into authentication issues ([in a similar fashion to this](https://github.com/GoogleCloudPlatform/java-docs-samples/issues/1377)) that were resolved by changing to the beta releases which are more up to date.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
acmiyaguchi: The `parsedTimed` method makes references to private, non-static variables, so it probably doesn't make sense in this context.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
acmiyaguchi: Trying to validate an ArrayNode is frustrating, it looks like there's a mismatch between the ArrayNodes and JSONArrays.

acmiyaguchi: I've factored this out, but there are some extra conversions needed to maintain the separation of libraries.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
acmiyaguchi: This is not the case, it inherits directly from Exception. https://www.javadoc.io/doc/org.bitbucket.b_c/jose4j/latest/org/jose4j/lang/JoseException.html

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
acmiyaguchi: UncheckedIOException doesn't work since JoseException is not an IOException.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,104 @@
...
+    // required to validate Jackson objects
+    JsonValidator validator = new JsonValidator();
+
+    ArrayNode metadata;
+    try (InputStream inputStream = BeamFileInputStream.open(this.metadataLocation)) {
+      byte[] data = IOUtils.toByteArray(inputStream);
+      metadata = Json.readArrayNode(data);
+      validator.validate(schema, metadata);
+    } catch (IOException e) {
+      throw new IOException("Exception thrown while reading keystore metadata schema.");
```
acmiyaguchi, line 75:
```suggestion
      throw new IOException("Exception thrown while reading keystore metadata schema.", e);
```

acmiyaguchi:
@jklukas This is ready for review now. One thing I'd like to do is measure the cost of decryption/decompression with respect to the rest of the job, but it's something that could be done later.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/JsonValidator.java
```diff
@@ -21,4 +26,16 @@ public void validate(Schema schema, ObjectNode json) throws JsonProcessingExcept
...
+  /**
+   * Validate a Jackson array.
+   *
+   * <p>Note this requires serializing the array back to a JSON string because
+   * the Jackson mapper cannot convert an ArrayNode to an org.json.JSONArray
+   * type.
+   */
+  public void validate(Schema schema, ArrayNode json) throws JsonProcessingException, IOException {
+    InputStream serialized = IOUtils.toInputStream(json.toString(), "UTF-8");
+    JSONArray array = Json.readJsonArray(serialized);
```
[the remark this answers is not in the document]
acmiyaguchi: I saw variations of this, but it requires instantiating a new mapper object. I suppose this is better than serializing to a string though.

ingestion-beam/src/main/resources/keystore-metadata.schema.json
```diff
@@ -0,0 +1,23 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "metadata for the key store",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "document_namespace": {
```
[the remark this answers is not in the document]
acmiyaguchi: Yep, this is totally doable.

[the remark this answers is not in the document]
acmiyaguchi: This has been renamed `private_key_id` to match `private_key_uri`.

acmiyaguchi:
This will require another round of significant changes before it can actually be deployed because this assumes a very simple schema for the envelope. This would work if the pings were being send directly to structured ingestion. In reality, the Pioneer pings are sending being send via Telemetry, and therefore the ping will need to include metadata to ensure it's being routed to the correct location.

This PR focuses is focused on implementing the transform and the key management -- a smaller followup can deal with the particulars of decrypting a differently shaped enveloped. I'm going to update the design documentation to reflect what changes will need to be made before implementing the next part.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -37,6 +38,34 @@
...
+      + " private keys, Cloud KMS resource ids for decrypting those keys, and their corresponding"
+      + " document namespaces; leave null to disable.")
+  ValueProvider<String> getPioneerMetadataLocation();
+
+  void setPioneerMetadataLocation(ValueProvider<String> value);
+
+  @Description("If set to true, assume that all private keys are encrypted with the associated"
+      + " KMS resourceId. Otherwise ignore KMS and assume all private keys are stored in plaintext."
+      + " This may be used for debugging.")
+  @Default.Boolean(true)
```
jklukas, line 57: Last I checked, default values were ignored for value providers in practice, although this is not documented. To provide a default value for a ValueProvider, you have to check at time of use and inject a default value there. We should remove this annotation as it has no effect.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -37,6 +38,34 @@
...
+  ValueProvider<Boolean> getPioneerKmsEnabled();
+
+  void setPioneerKmsEnabled(ValueProvider<Boolean> value);
+
+  @Hidden
+  @Description("Decompress pioneer pings.")
+  @Default.Boolean(true)
```
jklukas, line 64: Last I checked, default values were ignored for value providers in practice, although this is not documented. To provide a default value for a ValueProvider, you have to check at time of use and inject a default value there. We should remove this annotation as it has no effect.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/JsonValidator.java
```diff
@@ -21,4 +27,18 @@ public void validate(Schema schema, ObjectNode json) throws JsonProcessingExcept
...
+   * the Jackson mapper cannot convert an ArrayNode to an org.json.JSONArray
+   * type.
+   */
+  public void validate(Schema schema, ArrayNode json) throws JsonProcessingException, IOException {
+    ObjectMapper mapper = new ObjectMapper();
+    mapper.registerModule(new JsonOrgModule());
+    JsonNode root = mapper.valueToTree(json);
+    JSONArray jsonArray = mapper.treeToValue(root, JSONArray.class);
+    validator.performValidation(schema, jsonArray);
+  }
```
jklukas: See https://github.com/mozilla/gcp-ingestion/pull/1260 for an implementation of `validate` for ArrayNode.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,122 @@
+package com.mozilla.telemetry.util;
+
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.node.ArrayNode;
+import com.google.api.pathtemplate.ValidationException;
```
jklukas, line 5: I think this is importing the wrong class.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
+      PCollection<PubsubMessage> messages) {
+    return messages.apply(FlatMapElements.into(TypeDescriptor.of(PubsubMessage.class)) //
+        .via(new Fn()) //
+        .exceptionsInto(TypeDescriptor.of(PubsubMessage.class)) //
+        .exceptionsVia((WithFailures.ExceptionElement<PubsubMessage> ee) -> {
+          try {
+            throw ee.exception();
+          } catch (IOException e) {
```
jklukas, line 51: Do we also want to catch JoseException here and route those to error output?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
acmiyaguchi: This does seem to work though, based on the main decoder test where I don't have to include the flag when launching the pipeline. I would like this value to always be true, but the decompression stage does require a `ValueProvider<boolean>` which needs to be statically defined (or provided before runtime).

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
acmiyaguchi: If a JoseException were thrown here, would it kill the entire pipeline?

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
acmiyaguchi: Interesting that this doesn't fail, since I do check for the ValidationException in the unit tests.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
jklukas: This is how `DecompressPayload` handles the value:

```
      if (enabled.isAccessible() && !enabled.get()) {
        // Compression has been explicitly turned off in options, so return unchanged message.
```

So `DecompressPayload` is essentially enforcing `true` as the default.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
jklukas: > If a JoseException were thrown here, would it kill the entire pipeline?

It would bubble up and kill the worker, causing a new worker to get spun up. If it's a transient error, that's a good thing, because the message should succeed on retry. If it's a problem with the message itself, it will eventually cause the entire pipeline to fail after 4 worker failures.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -0,0 +1,87 @@
...
+      message = PubsubConstraints.ensureNonNull(message);
+
+      if (keyStore == null) {
+        keyStore = KeyStore.of(metadataLocation.get(), kmsEnabled.get());
+      }
+
+      ObjectNode json = Json.readObjectNode(message.getPayload());
+      PrivateKey key = keyStore.getKey(message.getAttribute(Attribute.DOCUMENT_NAMESPACE));
+
+      byte[] decrypted = decrypt(key, json.get("payload").asText());
```
jklukas, line 83:
```suggestion
      byte[] decrypted = decrypt(key, json.get(FieldName.PAYLOAD).asText());
```

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
```diff
@@ -83,6 +87,20 @@ public static String asString(Object data) throws IOException {
...
+    return new JSONObject(new JSONTokener(inputStream));
+  }
+
+  /**
+   * Read an org.json array for JSON validation.
+   */
+  public static JSONArray readJsonArray(InputStream inputStream) {
+    return new JSONArray(new JSONTokener(inputStream));
+  }
+
```
jklukas, line 103: These can be removed now, correct?

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
jklukas: This is because everit's ValidationException extends IOException, so we successfully catch it anyway without explicitly listing it in the signature.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,122 @@
...
+
+    keys = tempKeys;
+  }
+
+  private void ensureKeysLoaded() {
+    if (keys == null) {
+      try {
+        loadAllKeys();
+      } catch (IOException e) {
+        throw new UncheckedIOException("unexpected error while loading keys", e);
```
jklukas, line 118: It looks like we already explicitly catch and wrap all exception cases in `loadAllKeys`. I think I'd prefer to make sure those are all runtime exceptions (UncheckedIOException in particular where it makes sense) with helpful explanation messages and avoid an additional layer in the stack trace here.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,122 @@
...
+
+        PublicJsonWebKey key;
+        if (kmsEnabled) {
+          try (KeyManagementServiceClient client = KeyManagementServiceClient.create()) {
+            DecryptResponse response = client.decrypt(kmsResourceId, ByteString.copyFrom(keyData));
+            key = PublicJsonWebKey.Factory
+                .newPublicJwk(new String(response.getPlaintext().toByteArray()));
+          }
+        } else {
+          key = PublicJsonWebKey.Factory.newPublicJwk(new String(keyData));
```
jklukas, line 99: We should always specify `StandardCharsets.UTF_8` when converting from string to bytes or vice versa.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
```diff
@@ -0,0 +1,122 @@
...
+
+      try (InputStream inputStream = BeamFileInputStream.open(privateKeyUri)) {
+        byte[] keyData = IOUtils.toByteArray(inputStream);
+
+        PublicJsonWebKey key;
+        if (kmsEnabled) {
+          try (KeyManagementServiceClient client = KeyManagementServiceClient.create()) {
+            DecryptResponse response = client.decrypt(kmsResourceId, ByteString.copyFrom(keyData));
+            key = PublicJsonWebKey.Factory
+                .newPublicJwk(new String(response.getPlaintext().toByteArray()));
```
jklukas, line 96: We should always specify `StandardCharsets.UTF_8` when converting from string to bytes or vice versa.

ingestion-beam/src/main/resources/keystore-metadata.schema.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "metadata for the key store",
+    "type": "array",
+    "items": {
+        "type": "object",
+        "properties": {
+            "private_key_id": {
```
jklukas, line 8: This looks good. Thanks for making that name change.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
acmiyaguchi: `readJsonObject` is still necessary for loading the schema into the validator object. `readJsonArray` can be removed.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
acmiyaguchi: Okay, that makes sense. My misunderstanding comes from the `kmsEnabled` option, where I've only tested the non-default value and assumed that defaults actually worked.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/Json.java
jklukas: I want to avoid all use of `org.json`'s facilities for reading Json. I would prefer that we remove `readJsonObject` and instead materialize `inputStream` to a byte array and use `Json.readValue(bytes, JSONArray.class)` in analogy to what we do in `JSONSchemaStore`.

acmiyaguchi: Right, I see. I forgot about casting, although JSONObject does need to be imported into scope somewhere so Jackson can cast to the instance type.

jklukas: How about this: we add a public static method `JSONSchemaStore.readSchema(bytes)` that calls `Json.readValue(bytes, JSONObject.class)`. That way we don't increase the surface area of classes that see JSONObject.

acmiyaguchi: It's quite a bit of indirection to avoid the use of `org.json`, but making the JSONSchemaStore handle parsing schemas seems like a logical fit.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/KeyStore.java
acmiyaguchi: I'll leave out the message here, but wrap the IOExceptions in the unchecked wrapper.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
acmiyaguchi: Do you know what module this comes from?

jklukas: `com.mozilla.telemetry.ingestion.core.Constant.FieldName`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
acmiyaguchi: I've added a test to check for the default behavior, it seems to work as expected. The ValueProvider does set the value to true.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
acmiyaguchi: I've verified the behavior of the `@Default.Boolean` annotation in a unit-test for `kmsEnabled`, let me know what your thoughts are here.

jklukas: It's certainly possible this has been addressed in a recent Beam release. It's also possible that this test only works because the option is provided at compile time. This may be a case where we get a different result if we publish the pipeline as a template and then only provide the option at runtime. We won't be able to know without testing. I'm fine going with this as written here and we'll see what happens when we start testing.

acmiyaguchi: It would be nice if this were fixed in a recent release, since this seems like it should be the correct behavior. Otherwise, perhaps a bug in beam? I haven't looked too closely at this to understand the technical limitations behind providing a default value to a value provider, so if it does need to be changed, it doesn't seem like it should be too large of a patch.

PR #1117, opened 2020-02-06
ISP Lookup

scholtzan:
I'm working on https://bugzilla.mozilla.org/show_bug.cgi?id=1505491 to perform ISP lookups before IP addresses get discarded and add that information to metadata. This information can be used to identify clients related to cloud automation.

For now, I added a separate step for looking up ISP information which relies on the MaxMind Geo ISP database (cloudops has a license for the database and the security team already uses it for [some of their stuff](https://github.com/mozilla-services/cloudops-maxmind-image)). 

As far as I understand, besides adding this step in the decoder there are a few more changes necessary to get this working:
* the Geo-ISP needs to be made accessible alongside the Geo-City database, probably here:     https://github.com/mozilla-services/cloudops-infra/blob/afcb895873fdbba63d2c714ba7080c7b66da5ab9/projects/beam/tf/modules/resources/dataflow_jobs/decoder.tf#L30
* the [schema template for the metadata](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/462b5dd515fe2147c6a281e1422565a991778ca2/templates/include/metadata/ingestionCommonMetadata.1.schema.json) needs to be updated: See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/501

* queries in [bigquery-etl](https://github.com/mozilla/bigquery-etl) need to be updated to include the ISP information

But before making all these changes I wanted to check in if that is the right way to go and if I'm missing something.

PR #1247, opened 2020-04-16
Add ISP to structured metadata

relud:
This is needed in addition to #1117 for ISP to be preserved anywhere but `payload_bytes_raw`


ingestion-beam/pom.xml
```diff
@@ -17,6 +17,21 @@
...
+            <dependencies>
+                <dependency>
+                    <groupId>${project.groupId}</groupId>
+                    <artifactId>ingestion-core</artifactId>
+                    <version>${project.version}</version>
+                </dependency>
+            </dependencies>
+        </profile>
+    </profiles>
+
```
jklukas: Is this block intended as part of this PR?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoIspLookup.java
```diff
@@ -0,0 +1,167 @@
...
+  public class Fn extends SimpleFunction<PubsubMessage, PubsubMessage> {
+
+    private transient DatabaseReader ispReader;
+
+    private final Counter foundIp = Metrics.counter(GeoCityLookup.Fn.class, "found_ip");
+    private final Counter countIspAlreadyApplied = Metrics.counter(Fn.class, "isp_already_applied");
+    private final Counter foundIsp = Metrics.counter(Fn.class, "found_isp");
+    private final Counter countIpForwarded = Metrics.counter(GeoCityLookup.Fn.class,
+        "ip_from_x_forwarded_for");
+    private final Counter countIpRemoteAddr = Metrics.counter(GeoCityLookup.Fn.class,
```
jklukas, line 85:
```suggestion
    private final Counter countIpRemoteAddr = Metrics.counter(GeoIspLookup.Fn.class,
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoIspLookup.java
```diff
@@ -0,0 +1,167 @@
...
+
+  @VisibleForTesting
+  public class Fn extends SimpleFunction<PubsubMessage, PubsubMessage> {
+
+    private transient DatabaseReader ispReader;
+
+    private final Counter foundIp = Metrics.counter(GeoCityLookup.Fn.class, "found_ip");
+    private final Counter countIspAlreadyApplied = Metrics.counter(Fn.class, "isp_already_applied");
+    private final Counter foundIsp = Metrics.counter(Fn.class, "found_isp");
+    private final Counter countIpForwarded = Metrics.counter(GeoCityLookup.Fn.class,
```
jklukas, line 83:
```suggestion
    private final Counter countIpForwarded = Metrics.counter(GeoIspLookup.Fn.class,
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoIspLookup.java
```diff
@@ -0,0 +1,167 @@
...
+          //
+          // In practice, many of the "first" addresses are bogus or internal,
+          // so we target the immediate sending client IP.
+          String[] ips = xff.split("\\s*,\\s*");
+          ip = ips[Math.max(ips.length - 2, 0)];
+          countIpForwarded.inc();
+        } else {
+          ip = attributes.getOrDefault(Attribute.REMOTE_ADDR, "");
+          countIpRemoteAddr.inc();
+        }
```
jklukas, line 123: I'd really like to avoid repeating this code for determining IP. We should break this up into:

```
            .apply(options.getInputType().read(options)) //
            .apply(ParseProxy.of()) // This would also now have responsibility for outputting an "ip" attribute
            .apply(GeoIspLookup.of(options.getGeoIspDatabase())) //
            .apply(GeoCityLookup.of(options.getGeoCityDatabase(), options.getGeoCityFilter())) // This would no longer strip IP address attributes
            .apply(DropIP.of()) // new transform that drops all IP address attributes
```

That probably should be a separate PR, though. This is fine for now.

ingestion-beam/pom.xml
relud: nope, it got rebased out, i'm fixing tests and found that i was going to create hard-to-decipher conflicts with #1117

relud:
copied @jklukas comments over to #1117, from which the relevant code originates.

closing in favor of https://github.com/scholtzan/gcp-ingestion/pull/1 which will merge this into #1117

relud:
This change got lost from #1117, probably during a rebase somewhere, so I'm merging it based on existing r+'s

PR #1254, opened 2020-04-20
Git ignore vscode files

relud:


.gitignore
```diff
@@ -2,5 +2,9 @@
 *.swo
 *.swp
 .DS_Store
+.classpath
+.factorypath
 .idea/
+.project
+.settings/
 target/
```
acmiyaguchi, line 10: `.vscode/` should also be added here.

PR #1255, opened 2020-04-21
Fail storage.create if the blob already exists

relud:


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Gcs.java
```diff
@@ -103,7 +104,8 @@ private Batch(String bucket, String keyPrefix) {
 
       @Override
       protected CompletableFuture<Void> close() {
-        return batchCloseHook.apply(storage.create(blobInfo, content.toByteArray()));
+        return batchCloseHook.apply(
```
jklukas, line 107: It would be worth commenting here that we expect this to throw a particular exception type if the blob exists.

PR #1259, opened 2020-04-22
Detect errors on bq load in ingestion-sink

relud:
without this the `failsOnMissingBlob` test showed that missing blob errors were completely ignored.

cc @whd


relud:
> without this the `failsOnMissingBlob` test showed that missing blob errors were completely ignored.

well then. in CI apparently they still are -_-

relud:
okay, CI fixed, presumably because I forcibly skipped the `~/.m2` cache for a build.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/BigQuery.java
```diff
@@ -1,10 +1,12 @@
 package com.mozilla.telemetry.ingestion.sink.io;
 
+import avro.shaded.com.google.common.collect.ImmutableList;
```
jklukas, line 3: We should avoid using avro's shaded guava.

relud: >_< nice catch

PR #1261, opened 2020-04-22
Check if blob exists in BigQuery.Load

relud:
This prevents the entire load operation from failing due to a single missing uri, because blobs may take longer to become available in GCS than it takes for them to reach `BigQuery.Load`.

This also mitigates an issue in stage where missing blobs are blocking existing blobs from being loaded.


jklukas:
My understanding was that GCS provides strong consistency for read-after-write, so I expected it wouldn't be possible for this situation to occur. Unless we're publishing the object info before the write succeeds.

relud:
> My understanding was that GCS provides strong consistency for read-after-write, so I expected it wouldn't be possible for this situation to occur.

indeed it does: https://cloud.google.com/storage/docs/consistency

The actual issue here is that the subscription for the loader contains messages for gcs objects that don't exist (and stay that way), and dropping those messages (to a dead letter topic) doesn't cause lost documents. My current suspicion is that either these are duplicate messages, or they were successfully loaded and deleted but then nacked.

BigQuery load operations still suffer from occasional transient errors about objects not being present, but this didn't mitigate those errors, so presumably that's some sort of internal failure case for load jobs.

Separately, this revealed that the loader has very limited throughput, which I think is due to long batch times causing input rate limiting.

PR #1264, opened 2020-04-24
Draft docs v2 for Account Ecosystem Telemetry

jklukas:
This is an updated design compared to #1109

We now expect ecosystem_anon_id values in the payload rather than as headers.
We have also determined that it's feasible to route a variety of pings to
the AET-specific topics at the edge, so we should be able to use the desktop
telemetry and Glean libraries without modification. More to come on specific
schemas.


jklukas:
Updated diagram looks like:

image

It's the extra branch between Edge Service and Raw Topics that is new.

docs/architecture/overview.md
```diff
@@ -37,6 +45,25 @@ This document specifies the architecture for GCP Ingestion as a whole.
...
+
+- Must have access restricted to a limited set of operators
+- Must have access to a set of encrypted private keys stored in
+  the source repository or on GCS
+- Must have access to KMS to decrypt the private keys on startup,
+  holding them in memory
+- Must throw away any payload that does not match expected structure
+  - The payload must contain valid JSON
+  - The payload may be compressed via `gzip`
+  - The payload must contain a top-level field called `ecosystem_anon_id`
```
jklukas, line 58: This can be true for desktop telemetry pings and for pings sent from FxA, but probably isn't possible from Glean. We can modify this to accommodate a good place for Glean to record these values.

docs/ingestion-beam/account-ecosystem-decryptor-job.md
```diff
@@ -0,0 +1,54 @@
+# Account Ecosystem Decryptor Job
+
+A job for decrypting identifiers used in Account Ecosystem Telemetry.
+Defined in the `com.mozilla.telemetry.EcosystemDecryptor` class ([source](https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/src/main/java/com/mozilla/telemetry/EcosystemDecryptor.java)).
```
jklukas, line 4: Does not exist yet :)

ingestion-beam/src/main/resources/account-ecosystem-ping.schema.json
```diff
@@ -0,0 +1,33 @@
...
+                "type": "array",
+                "items": {
+                    "type": "string"
+                }
+            }
+        },
+        "required": [
+            "ecosystem_anon_id"
+        ],
+        "not": {
```
jklukas, line 23: Validation will fail if fields are present with any of the following names. They're all misspellings or they're names for the decrypted fields rather than the expected input encrypted fields.

docs/architecture/overview.md
```diff
@@ -37,6 +45,25 @@ This document specifies the architecture for GCP Ingestion as a whole.
...
+    and must _not_ contain top-level fields `ecosystem_user_id` or
+    `previous_ecosystem_user_ids` (which indicate a client not conforming
+    to Account Ecosystem Telemetry contracts)
+- Must replace field `ecosystem_anon_id` with the decrypted value in a
+  field called `ecosystem_user_id`
+- Must replace array-type field `previous_ecosystem_anon_ids` if it exists
+  with the decrypted values in a field called `previous_ecosystem_user_ids`
```
jklukas, line 65: This `previous_ecosystem_anon_ids` array gives us the flexibility to implement either a "ID changed ping" where the array would have a single value, or a deletion request ping with the whole history of values that fxa knows about.

ingestion-edge/docker-compose.yml
```diff
@@ -13,6 +13,15 @@ services:
           "projects/test/topics/stub_installer",
           ["GET"]
         ],
+        # ## Proposed routes for account ecosystem telemetry
+        # [
+        #   "/submit/telemetry/<document_id:string>/<suffix:account-ecosystem.*>",
+        #   "projects/test/topics/telemetry-ecosystem"
+        # ],
+        # [
+        #   "/submit/<namespace:string>/<suffix:account-ecosystem.*>",
```
jklukas, line 22: These proposed routes use a regex where we match any ping with document type starting with "account-ecosystem", so if we have multiple pings, they could be named "account-ecosystem-deletion-request", etc. and still be routed through the decryptor.

docs/architecture/overview.md
rfk: I like this FWIW, feels like it gives us some good optionality

ingestion-edge/docker-compose.yml
rfk: The more I think about it, the more I think we're likely to have multiple different pings, so this seems 👍 to me at a high level.

rfk: As one example of where we might have multiple pings, I think we'll probably want to port the [sync ping](https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/sync-ping.html) over to use these new identifiers, and there is a *lot* of sync-specific stuff in there that I don't think would make sense to shoehorn into a generic shared schema.

rfk:
Re: schema for these pings in general, do we expect it to look something like the current "pre-account ping" documented here?

  https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/ecosystem-telemetry.html#pre-account-ping

Copy-pasting and editing for latest terminology, that would make it something like:

```
{
  "type": "account-ecosystem-main",
  ... /// common ping data, incuding...
  "application": {
    "name": <string>, // Unique identifier for this client applicaiton (e.g. "Firefox")
  },
  "environment": { ... }, // restricted subset of the usual environment data
  "payload": {
    "reason": <string>, // Why the ping was submitted
    "ecosystemClientId": <string>, // Specific ecosystem client ID, not associated with regular client ID
    "ecosystemAnonId": <string>, // Encrypted uid as provided by FxA service (only present if signed in)
    "previousEcosystemAnonIds": [<string>], // Previously-seen values of the above
    "duration": <number>, // duration since ping was last sent or since the beginning of the Firefox session in seconds
    "histograms": {...},
    "keyedHistograms": {...},
    "scalars": {...},
    "keyedScalars": {...},
  }
}
```

Some thoughts:

* It seems like the prevailing style in desktop pings is camelCase rather than snake_case, should we try to be consistent here?
* It's my understanding that there will be a single "AET ping" submitted by all Firefox Desktop clients, regardless of whether or not the user is currently signed in to that profile. This affects what's optional and what's not:
  * `application.name`: **Required**, we need this in order to distinguish AET data from different applications in aggregate analysis (e.g. if we combine AET data from the FxA servers and Firefox Desktop, one would have `application.name=Firefox` and one would have `application.name=FxA` and we'd like to be able to filter on it).
  * `payload.ecosystemClientId`: **Required**, not having this kind of defeats the whole point of AET.
  * `payload.ecosystemAnonId`: **Optional**, because if the user isn't signed in then we can't report this value, but we still want to send AET pings with `ecosystemClientId`.

So I think there's a potential mis-match here with the pipeline requiring the `ecosystemAnonId` field, but clients not being able to send that field unless the user is signed in.

On the other hand, if there is no `ecosystemAnonId` then there's no need to route the ping through the special decryptor logic in the first place.

jklukas:
> Re: schema for these pings in general, do we expect it to look something like the current "pre-account ping"

I likely have seen this before, but had forgotten that a pre-account ping was already implemented for desktop. I do think it's reasonable to support the existing structure as it exists there, so we can allow these to be nested under `payload` and use camelCase names.

`pre-account` is a nice pithy name for this ping, but for routing it would be nice to have an unambiguous common prefix `account-ecosystem`. I need to think more about that point.

>  application.name: Required, we need this in order to distinguish AET data from different applications in aggregate analysis (e.g. if we combine AET data from the FxA servers and Firefox Desktop, one would have application.name=Firefox and one would have application.name=FxA and we'd like to be able to filter on it).

Making the source clear is valuable, but that can be handled at the ETL layer since we now have a design where desktop pings will be going to a separate table from FxA pings, and each Glean app will have its own destination table per ping as well. So we can stitch those together with appropriate app names at query time.

> So I think there's a potential mis-match here with the pipeline requiring the ecosystemAnonId field, but clients not being able to send that field unless the user is signed in.

> On the other hand, if there is no ecosystemAnonId then there's no need to route the ping through the special decryptor logic in the first place.

This makes sense to me. I think we would still route the payload to the decryptor based on the URI, but we can allow ecosystemAnonId to be optional. We can require `ecosystemClientId`; if that's in the right place and right format, it would give us some confidence that the ping is reasonably formatted.

rfk:
> pre-account is a nice pithy name for this ping, but for routing it would be nice to have an
> unambiguous common prefix account-ecosystem

I expect we will refactor the `pre-account` ping into a more general `account-ecosystem` ping of some sort, so the common prefix sounds good from my perspective.

docs/architecture/overview.md
jklukas: It sounds like we'll need to have a set of schemas with the anon_id in different places: desktop, fxa, and glean.
I will also relax this such that anon ID is not required to be present.

jklukas:
As I'm working on implementation of the decryptor right now, I'm thinking that we probably do want an error stream that we can send to BigQuery, as it will provide a much better interface for debugging issues compared to logging and metrics.

We would need to sanitize the payloads; we already have precedent for this with `DuplicateIdExceptions` where we specify a null payload in the error output, so we end up with a row in `payload_bytes_error` in BQ that just contains attributes associated with the duplicate message. We could send a null payload for AET errors, or we could try a more sophisticated approach where we redact all string fields from the JSON payload; that would allow us to see the structure of the failing payload, but guarantee any misplaced `anon_id` values don't leak to the error stream.

I need to get @whd's input on how much this would complicate deployment. Would we need to provision new topics and sinks, or could we publish to the existing `telemetry-errors` and `structured-errors` topics?

ingestion-beam/src/main/resources/account-ecosystem-ping.schema.json
jklukas: As I'm implementing the decryptor, it turns out this method doesn't work with multiple fields. Validation will only fail if the payload has _all_ of the listed fields. Instead, I'll be including these field names in the main schema like `"ecosystemAnonId": false` which has no effect if the field is not present, but will always fail if the field _is_ present (since no field matches schema `false`).

whd:
Adding to the top since this turned into an essay: In any case adding error semantics from a deployment perspective is fairly simple if sensitivity isn't an issue: we can re-use the existing error topic for the pipeline family and the deployment will not be significantly more complicated. The only differences would be that the `aet` service account will be granted write access to the `error` topic for its pipeline family and the `AET` job (whether `Decoder` or `AETDecryptor`, see below) would have `errorOutput` set accordingly.

# Essay
As previously spec'd `AET` adds a new input topic to a particular pipeline family with different downstream processing semantics than what I'm going to call the `raw` input, summarized as follows:

1. privileged KMS access for `AETDecryptor` and associated custom decryption logic, accomplished by input topic separation
1. `AETDecryptor` was spec'd as a small "pre-Decoder" that was expected to output its data to the standard `raw` topic, to facilitate "safe" raw output and more importantly use the rest of the downstream processing pipeline
1. differing error semantics during the decode step (previously to drop errors in `AET`, proposed to redact errors instead)

As you point out we already have error processing logic that performs redaction in the standard `Decoder`, among other error cases.

I think there are essentially two paths to choose from:

1. `AETDecryptor` remains a special small pre-Decoder step that writes to raw and errors topics for its pipeline family
2. `AETDecryptor` becomes a compile-time option to the `Decoder`, similar to Pioneer, writing to decoded and errors topics for its pipeline family

I don't have a strong preference here, I'm merely observing that the privilege separation is already accomplished via a separate input topic and service accounts, and depending on the new error semantics the `AETDecryptor` could conceivably be just a `Decoder`.

I would say if we're expecting that for all cases that the `AET` pre-Decoder will actually have full decoder semantics WRT schemas in order to avoid accidental publishing of sensitive data (i.e. there will be no `payload_bytes_error` output from `AET` via the standard decoder), then the only advantage to that style is that `AET` data that is well-formed will wind up in `payload_bytes_raw`. I doubt this is sufficient reason to introduce a new job class when we can get by with decoder config, as we do with Pioneer.

PR #1269, opened 2020-04-28
Use separate input and output executors in ingestion-sink

relud:
Messages are being queued inside `CompletableFuture`s that all share the same executor capacity, so when `Batch::close` takes a few minutes for `BigQuery.Load`, the slow batches claim all of the available threads and prevent messages from reaching other batches in a timely manner, resulting in unnecessarily small batches.

This attempts to fix that by moving as much work as possible into the subscriber threads, and leaving the output executor threads to focus on `Batch::close` calls.

Additionally this makes handling batch results synchronous, so that thread that closes a batch has to ack/nack all of the attached futures before moving on.


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/Env.java
```diff
@@ -37,6 +39,9 @@ public boolean containsKey(String key) {
   }
 
   private Optional<String> optString(String key) {
+    if (!include.contains(key)) {
+      throw new IllegalArgumentException("key missing from include: " + key);
```
relud, line 43: this should be hit by `Sink*IntegrationTest` if we pull any env vars missing from include, which is a problem I hit again recently `>_<`

PR #1270, opened 2020-04-28
Skip ci jobs on gh-pages branch deployed by mkdocs

wlach:
There is no circleci config on this branch, so it will always fail.


wlach:
Haven't tested this but it should work according to the `mkdocs` command line.

PR #1274, opened 2020-04-29
Fix #1273 - Update GeoLite2-City extraction for new directory structure

### LINKED ISSUE #1273: Running decoder locally fails with "com.maxmind.db.InvalidDatabaseException"
While running the integration testing workflow for ingestion-beam, I ran into the following exception:

```
Caused by: com.maxmind.db.InvalidDatabaseException: The MaxMind DB file's search tree is corrupt: contains pointer larger than the database.`
```

This can be reproduced as follows:

```
cd ingestion-beam
./bin/download-cities15000
./bin/download-geolite2
./bin/download-schemas
./bin/download-document-sample


./bin/mvn compile exec:java -Dexec.mainClass=com.mozilla.telemetry.Decoder -Dexec.args="\
    --geoCityDatabase=GeoLite2-City.mmdb \
    --geoCityFilter=cities15000.txt \
    --schemasLocation=schemas.tar.gz \
    --inputType=file \
    --input=document_sample.ndjson \
    --outputType=file \
    --output=output/output  \
    --errorOutputType=file \
    --errorOutput=output/error
"
```

acmiyaguchi:
This fixes #1273. The GeoLite2-Cirty archive now includes several additional files that caused the `download-geolite` script to generate an invalid mmdb binary. This fixes issues with the `GeoCityLookup` step of the beam pipeline when running the decoder locally for integration testing purposes.

### DIFF
--- a/ingestion-beam/bin/download-geolite2
+++ b/ingestion-beam/bin/download-geolite2
@@ -6,13 +6,27 @@
 # from MaxMind for use in development.
 BASE_DIR="$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )")"
 DOWNLOAD_URL="https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key=${MM_LICENSE_KEY}&suffix=tar.gz"
-LOCAL_ARCHIVE=/tmp/GeoLite2-City.mmdb.gz
-LOCAL_EXTRACTED=/tmp/GeoLite2-City.mmdb
 
 set -e
 
+# The GeoLite2-City archive is structured as follows:
+#
+#   x GeoLite2-City_20200428/
+#   x GeoLite2-City_20200428/README.txt
+#   x GeoLite2-City_20200428/COPYRIGHT.txt
+#   x GeoLite2-City_20200428/GeoLite2-City.mmdb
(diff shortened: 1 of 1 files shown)


ingestion-beam/bin/download-geolite2
```diff
@@ -6,13 +6,27 @@
 # from MaxMind for use in development.
 BASE_DIR="$(dirname "$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )")"
 DOWNLOAD_URL="https://download.maxmind.com/app/geoip_download?edition_id=GeoLite2-City&license_key=${MM_LICENSE_KEY}&suffix=tar.gz"
```
scholtzan, line 8: I wonder if it would make sense to have a check for whether `MM_LICENSE_KEY` is set in the script. Right now, even if it's not set it'll still try to download the files without erroring out. 

Having a error message printed out might make it easier for people setting up the project for the first time to figure out what's going wrong. Though there is documentation somewhere for this, so probably not super urgent.

acmiyaguchi:
Added in a small check for the license, and fixed a bug that would keep redownloading the file even when it existed locally. I tested this in the following way:

* `unset MM_LICENSE_KEY`, remove existing mmdb, and run to failure
* `export MM_LICENSE_KEY=...` and check for mmdb
* rerun and check that it terminates quickly

PR #1275, opened 2020-04-30
Add script for benchmarking Pioneer decryption

acmiyaguchi:
This adds scripts for measuring the overhead of the `DecryptPioneerPayloads` step in Dataflow. This particular run contained 58,933 elements, which is 528.53 MB plaintext and      692.03 MB ciphertext. Each job had ~1 minute total clock time on `ParsePayload`. The Pioneer enabled job had 3 minutes on the `DecryptPioneerPayloads` step. The difference in vCPU/hr (0.096 plaintext vs 0.14 ciphertext) doesn't accurately capture the difference above since there is the startup and cleanup overhead of the job.

* [baseline dataflow job](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2020-04-30_12_08_03-13899129925267677039?project=amiyaguchi-dev)
* [pioneer enabled dataflow job](https://console.cloud.google.com/dataflow/jobsDetail/locations/us-central1/jobs/2020-04-30_12_08_43-191213889725123005?project=amiyaguchi-dev)
* [baseline profiling](https://console.cloud.google.com/profiler/decoder-0-0430190751-f2be1cb9/cpu?project=amiyaguchi-dev)
* [pioneer-enabled profiling](https://console.cloud.google.com/profiler/decoder-0-0430190833-f2251679/cpu?project=amiyaguchi-dev)

### DIFF
--- a/ingestion-beam/bin/run-pioneer-benchmark
+++ b/ingestion-beam/bin/run-pioneer-benchmark
@@ -7,13 +7,16 @@
 # Initial findings show that for every 1 minute spent on ParsePayload, there are
 # 3 minutes spend on DecryptPioneerPayloads. Peak throughput in ParsePayload
 # goes from 3360 elements/sec to 2466 elements/sec when adding encryption.
+#
+# This script requires the use of GNU Coreutils. This may be installed on MacOS
+# via homebrew: `brew install coreutils`.
 
 set -ex
 
 export MAVEN_OPTS="-Xms8g -Xmx8g"
 reset=${RESET:-false}
 project=$(gcloud config get-value project)
-bucket="gs://${BUCKET:-$project}"
+bucket="gs://${BUCKET?bucket value must be specified}"
 prefix="ingestion-beam-benchmark"
 staging="benchmark_staging"
 
@@ -24,7 +27,7 @@ if [[ ! -f document_sample.ndjson ]]; then
     exit 1
 fi
 
-if [[ ! -f pioneer_benchmark_data.ndjson ]] || ${reset}; then
+if [[ ! -f pioneer_benchmark_data.ndjson ]] || [[ ${reset} == "true" ]]; then
     # generate the data using the document sample, this may take a while depending on your machine
     ./bin/mvn clean compile exec:java -Dexec.mainClass=com.mozilla.telemetry.PioneerBenchmarkGenerator
 fi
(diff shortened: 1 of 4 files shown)


ingestion-beam/bin/run-pioneer-benchmark
```diff
@@ -0,0 +1,100 @@
...
+staging="benchmark_staging"
+
+cd "$(dirname "$0")/.."
+
+if [[ ! -f document_sample.ndjson ]]; then
+    echo "missing document_sample.ndjson, run download-document-sample"
+    exit 1
+fi
+
+if [[ ! -f pioneer_benchmark_data.ndjson ]] || ${reset}; then
```
jklukas, line 27: Calling the value of `RESET` as an executable seems potentially dangerous. It might be safer to check `[[ "$reset" = "true" ]]`

ingestion-beam/bin/run-pioneer-benchmark
```diff
@@ -0,0 +1,100 @@
...
+# Initial findings show that for every 1 minute spent on ParsePayload, there are
+# 3 minutes spend on DecryptPioneerPayloads. Peak throughput in ParsePayload
+# goes from 3360 elements/sec to 2466 elements/sec when adding encryption.
+
+set -ex
+
+export MAVEN_OPTS="-Xms8g -Xmx8g"
+reset=${RESET:-false}
+project=$(gcloud config get-value project)
+bucket="gs://${BUCKET:-$project}"
```
jklukas, line 16: This could be surprising. I think it would be reasonable to require BUCKET be specified.

ingestion-beam/bin/run-pioneer-benchmark
```diff
@@ -0,0 +1,100 @@
...
+# generate a new folder and sync it to gcs, assumes a bucket value
+if [[ ! -d $staging ]]; then
+    plaintext=$staging/input/plaintext
+    ciphertext=$staging/input/ciphertext
+    metadata=$staging/metadata
+    mkdir -p $plaintext
+    mkdir -p $ciphertext
+    mkdir -p $metadata
+    # shuffle to avoid data skew, and to prepare for file splitting if necessary
+    shuf document_sample.ndjson > $plaintext/part-0.ndjson
```
jklukas, line 44: Is `shuf` from gnu core-utils? It would be worth mentioning at the top that this script assumes gnu core-utils are installed.

acmiyaguchi: Ah it is from coreutils, that's not ideal. I'll add a note to the top of the script then.

PR #1276, opened 2020-05-01
Update Pioneer transform to handle `pioneer-study` envelope

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1628539)

This finishes up the relevant decoder work for Pioneer v2.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -81,11 +91,36 @@ private DecryptPioneerPayloads(ValueProvider<String> metadataLocation,
...
+        validator = new JsonValidator();
+        byte[] data = Resources
+            .toByteArray(Resources.getResource("telemetry.pioneer-study.4.schema.json"));
+        envelopeSchema = JSONSchemaStore.readSchema(data);
+      }
+
       ObjectNode json = Json.readObjectNode(message.getPayload());
-      PrivateKey key = keyStore.getKey(message.getAttribute(Attribute.DOCUMENT_NAMESPACE));
+      validator.validate(envelopeSchema, json);
+      JsonNode payload = json.get("payload");
```
jklukas, line 103:
```suggestion
      JsonNode payload = json.get(FieldName.PAYLOAD);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -81,11 +91,36 @@ private DecryptPioneerPayloads(ValueProvider<String> metadataLocation,
...
+
+      String encryptionKeyId = payload.get("encryptionKeyId").asText();
```
jklukas, line 105: We should probably make string constants in this class for all these literal string field names.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -81,11 +91,36 @@ private DecryptPioneerPayloads(ValueProvider<String> metadataLocation,
...
+      PrivateKey key = keyStore.getKey(encryptionKeyId);
+      if (key == null) {
+        // Is this really an IOException?
+        throw new IOException(String.format("encryptionKeyId not found: %s", encryptionKeyId));
+      }
+
+      // NOTE: there could be a method for handling merging metadata like
```
jklukas, line 112: I'm not entirely sure what need is being described here, but see `AddMetadata.java` where we insert metadata at the beginning of a json payload without parsing it.

acmiyaguchi: There's the `application` field at the top level part of the ping (as part of the common ping format). It could be appended to the actual payload so it can be available to analysis downstream. The problem noted here is that the payload could be compressed before encryption, which would make it difficult to concatenate unless it were decompressed or somehow done in the post-decompression step.

PR #1277, opened 2020-05-05
Implement pubsub reads using pull API

relud:
which will replace `com.google.cloud.pubsub.v1.Subscriber` which uses the StreamingPull API. `Subscriber` isn't able to handle 2 million messages, each in flight for upwards of 10 minutes, because it eats unnecessary amounts of ram and eventually stalls. Somewhere around 200K messages was the upper limit in python, which closely follows the java implementation.

### DIFF
--- a/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
+++ b/ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
@@ -82,10 +82,10 @@ public abstract class BatchWrite<InputT, EncodedT, BatchKeyT, BatchResultT>
 
   protected final PubsubMessageToTemplatedString batchKeyTemplate;
   protected final Executor executor;
+  protected final Duration maxDelay;
 
   private final long maxBytes;
   private final int maxMessages;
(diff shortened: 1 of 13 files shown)


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
```diff
@@ -165,17 +165,21 @@ private void recordBatchMetrics(long bytes, int messages, long delayMillis) {
...
-    // wait for init then mark this batch full after maxDelay
     @VisibleForTesting
-    final CompletableFuture<Void> full = init.thenComposeAsync(v -> new TimedFuture(maxDelay));
+    public final CompletableFuture<Void> full = init
+        .thenComposeAsync(v -> new TimedFuture(maxDelay));
 
-    // wait for full then synchronize and close
-    private final CompletableFuture<BatchResultT> result = full
+    @VisibleForTesting
+    public final CompletableFuture<BatchResultT> result = full
```
relud, line 182: this chunk needs to be reverted

relud:
I was able to get the sdk subscriber to work better, so this is no longer needed.

PR #1281, opened 2020-05-07
Ensure batch-level exceptions aren't re-logged for every message

relud:


ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/io/PubsubReadIntegrationTest.java
```diff
@@ -26,7 +26,7 @@
   public final SinglePubsubTopic pubsub = new SinglePubsubTopic();
 
   @Rule
-  public final LoggerContextRule logs = new LoggerContextRule("log4j2-test.yaml");
+  public final LoggerContextRule logs = new LoggerContextRule("log4j2-list.yaml");
```
relud, line 29: this was renamed because it was impacting all tests

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
```diff
@@ -176,7 +178,13 @@ private void recordBatchMetrics(long bytes, int messages, long delayMillis) {
 
     // wait for full then synchronize and close
     private final CompletableFuture<BatchResultT> result = full
-        .thenComposeAsync(this::synchronousClose, executor);
+        .thenComposeAsync(this::synchronousClose, executor).exceptionally(exception -> {
+          // with one input rethrow exception, otherwise use LoggedException to only log it once
+          throw this.size == 1 ? (RuntimeException) exception
```
relud, line 183: preserving the original exception for single-message batches prevents logging the exception here when it may still be caught and fixed, e.g. when the mixed output attempts to stream very large messages and then redirects them to file loads.

ingestion-sink/src/test/java/com/mozilla/telemetry/ingestion/sink/util/BoundedSink.java
```diff
@@ -29,31 +30,17 @@
...
-      future.cancel(true);
-    }
-    try {
-      future.join();
-    } catch (CancellationException e) {
-      throw new RuntimeException("Failed to deliver messages in " + timeout + " seconds", e);
+    CompletableFuture<Void> future = runAsync(messageCount);
+    CompletableFuture.anyOf(future, new TimedFuture(Duration.ofSeconds(timeout))).join();
+    if (future.cancel(true)) {
+      throw new CancellationException("Failed to deliver messages in " + timeout + " seconds");
```
relud, line 43: this is included here because the timeout wasn't working

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Pubsub.java
```diff
@@ -46,8 +47,16 @@ private Pubsub() {
...
+                  // exception is always a CompletionException caused by another exception
+                  if (exception.getCause() instanceof BatchException) {
+                    // only log batch exception once
+                    ((BatchException) exception.getCause()).handle((batchExc) ->
+                        LOG.error(String.format("failed to deliver %d messages", batchExc.size),
+                            batchExc.getCause()));
+                  } else {
+                    // log exception specific to this message
+                    LOG.error("failed to deliver message", exception.getCause());
+                  }
```
relud, line 59: > I do have some concerns with this approach.

re-r? now that i've updated this per our discussion on friday

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
```diff
@@ -86,6 +88,8 @@ private static Distribution exponentialDistribution(double base, double expStart
   private final long maxBytes;
   private final int maxMessages;
   private final Duration maxDelay;
+  private final String shortClassName = this.getClass().getName().replaceAll(".*[.]", "");
+  private final Logger logger = LoggerFactory.getLogger(this.getClass());
```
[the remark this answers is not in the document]
jklukas: Or is the point that we want to catch subclass names?

For a logger, I'm used to the logger being static and the log messages refer to the superclass where it was logged rather than the concrete subclass. I think I expect the same for metrics, but I'm open to thoughts there.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/io/Pubsub.java
```diff
@@ -46,8 +47,16 @@ private Pubsub() {
...
                   consumer.ack();
                 } else {
-                  // exception is always a CompletionException caused by the real exception
-                  LOG.warn("Exception while attempting to deliver message", exception.getCause());
+                  // exception is always a CompletionException caused by another exception
+                  if (exception.getCause() instanceof BatchException) {
+                    // only log batch exception once
+                    ((BatchException) exception.getCause()).handle((batchExc) -> LOG.error(
+                        String.format("failed to deliver %d messages", batchExc.size),
+                        batchExc.getCause()));
```
jklukas, line 55: This does indeed seem easier to follow now.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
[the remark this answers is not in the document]
relud: this was actually leftover from the previous method. I removed this change.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/util/BatchWrite.java
```diff
@@ -86,6 +88,8 @@ private static Distribution exponentialDistribution(double base, double expStart
   private final long maxBytes;
   private final int maxMessages;
   private final Duration maxDelay;
+  private final String shortClassName = this.getClass().getName().replaceAll(".*[.]", "");
```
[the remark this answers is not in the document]
relud: this was actually leftover from the previous method. I removed this change.

PR #1283, opened 2020-05-08
Use jar with classpath instead of fat jar

relud:
and expose ParallelPullCount from pubsub subscriber sdk

### DIFF
--- a/ingestion-sink/Dockerfile
+++ b/ingestion-sink/Dockerfile
@@ -1,17 +1,20 @@
 FROM maven:3-jdk-8 AS base
 WORKDIR /app/ingestion-sink
+
+FROM base AS build
 COPY LICENSE /app/LICENSE
 COPY pom.xml /app/pom.xml
 COPY ingestion-sink/pom.xml /app/ingestion-sink/pom.xml
 COPY checkstyle /app/checkstyle
 COPY ingestion-sink/checkstyle /app/ingestion-sink/checkstyle
+RUN mvn prepare-package
 COPY ingestion-core/src /app/ingestion-core/src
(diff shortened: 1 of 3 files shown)


ingestion-sink/Dockerfile
```diff
@@ -1,17 +1,17 @@
...
-COPY ingestion-sink/pom.xml /app/ingestion-sink/pom.xml
-COPY checkstyle /app/checkstyle
-COPY ingestion-sink/checkstyle /app/ingestion-sink/checkstyle
+
+FROM base AS build
+COPY LICENSE checkstyle pom.xml /app/
+COPY ingestion-sink/pom.xml ingestion-sink/checkstyle /app/ingestion-sink/
+RUN mvn prepare-package
 COPY ingestion-core/src /app/ingestion-core/src
 COPY ingestion-sink/src /app/ingestion-sink/src
```
relud, line 9: this now downloads the dependencies before copying `src` directories, so that when building the docker image locally the dependencies are cached, rather than re-downloaded on every source change.

pom.xml
```diff
@@ -224,6 +207,35 @@
...
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addClasspath>true</addClasspath>
+                            <classpathPrefix>${project.build.finalName}.lib/</classpathPrefix>
```
relud, line 233: Dependencies have been moved to jars included via this classpath prefix to fix an issue where grpc was breaking because a `jar-with-dependencies` isn't able to include `META-INF/services` from every dependency.

PR #1284, opened 2020-05-08
Bug 1636317 - Add envelope metadata to Pioneer payload

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1636317)

Depends on https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/544 and https://github.com/mozilla/mozilla-schema-generator/pull/126

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
@@ -63,8 +63,14 @@ public class AddMetadata {
   public static MapWithFailures<PubsubMessage, PubsubMessage, PubsubMessage> of() {
     return MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via((PubsubMessage msg) -> {
       msg = PubsubConstraints.ensureNonNull(msg);
-      ObjectNode metadata = attributesToMetadataPayload(msg.getAttributeMap());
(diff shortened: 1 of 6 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -108,6 +78,40 @@
         });
   }
 
+  /** Merge a JSON byte payload with a ObjectNode. */
+  public static byte[] merge(byte[] payload, ObjectNode metadataObject)
+      throws UncheckedIOException {
```
jklukas, line 83: This signature feels unbalanced. I think I'd prefer that both arguments be byte arrays and it's expected that the caller has to transform the ObjectNode into bytes. I don't have a strong rationale for this, though.

And I think I'd call this `mergedPayload` to indicate what it returns.

acmiyaguchi: A consistent interface is rationale enough. I went from `mergePayloadWithMetadata` to `merge`, but it seems like I missed out on the happy medium.

acmiyaguchi: I renamed the function `mergePayload`, which seems to fit the naming convention a bit better.

PR #1289, opened 2020-05-12
Bug 1614409 - Ignore fenix-debug namespace

fbertsch:
I'm wondering if we can start not adding tests for all affected namespaces, considering the codepath is definitely "proven". Happy to add one if anyone objects.


jklukas:
I got an itchy trigger finger and merged this before I realized what I was doing. Should be fine...

> I'm wondering if we can start not adding tests for all affected namespaces, considering the codepath is definitely "proven". Happy to add one if anyone objects.

I agree that tests per ignored namespace seem unnecessary at this point.

PR #1292, opened 2020-05-13
Bug 1637784 - Add script to generate Pioneer v2 keys

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1637784)

```bash
$ ./generate-jwk-pair --help                                     
usage: generate-jwk-pair [-h] [-k KEY_ID] [-r]

Generate a JWK keypair for use in ingestion-beam.

optional arguments:
  -h, --help            show this help message and exit
  -k KEY_ID, --key-id KEY_ID
                        Key ID (kid) parameter. See https://tools.ietf.org/html/rfc7517#section-4.5
  -r, --random-key-id   Use a random identifier for the Key ID (kid) parameter.
```

```bash
$ ./generate-jwk-pair | jq
{
  "private_key": {
    "crv": "P-256",
    "d": "DvRgbMcaqSLOiWXxXo7Pn7Vsl4L8IZ6_i0Q-pLcSIys",
    "kty": "EC",
    "x": "SyTnTZRynGCu9SbmS4OD56gpzQBYNaJBy1CYDdtEL2o",
    "y": "gHmpuEYXhNRYSUBUKCWqlexVwt5XrWGhfW3MwII2ApU"
  },
  "public_key": {
    "crv": "P-256",
    "kty": "EC",
    "x": "SyTnTZRynGCu9SbmS4OD56gpzQBYNaJBy1CYDdtEL2o",
    "y": "gHmpuEYXhNRYSUBUKCWqlexVwt5XrWGhfW3MwII2ApU"
  }
}

$ ./generate-jwk-pair -r | jq '.private_key.kid'
"il_KHmeGsKnNyYIidpiNkfsDCmVD6IyvLMWe3Kcgcik"

$ ./generate-jwk-pair -k pioneer-debug | jq '.private_key.kid'
"pioneer-debug"
```


acmiyaguchi:
I'm also getting confirmation from :rhelmer since the public key will need to be shipped in the client. @jklukas is this something that could be used for AET, too?

ingestion-beam/bin/generate-pioneer-keys
```diff
@@ -0,0 +1,21 @@
...
+import sys
+import json
+
+try:
+    from jwcrypto import jwk
+except ImportError as exc:
+    print(f"{exc.__class__.__name__}: run `pip install jwcrypto` and try again")
+    sys.exit(1)
+
+key = jwk.JWK.generate(kty="EC")
```
jklukas, line 12: Everything here looks in line with [the example key in the AET anon_id proposal](https://docs.google.com/document/d/1zH3eVVI_28Afg1JXe_McDrW4MTYuWhiJMQR6AQbli8I/edit#heading=h.cda7bwgmtch1) except that the generated keys need to have a `key_id`. This may not be necessary for pioneer, but I also don't think it would hurt to include here.

We can pass a `key_id` param to `generate` here to include a key_id. I think it would make sense to randomly generate that value. I'm not sure how mmiller generated the key_id for his example and I'm not sure what restrictions exist on format for that param, but it looks to be in the same format as the "x", "y", and "d" in the example.

So easiest thing here might be to generate a first key and steal its "x" value to insert into our real key.

```suggestion
key_id = json.loads(jwk.JWK.generate(kty="EC").export_public())['x']
key = jwk.JWK.generate(kty="EC", key_id=key_id)
```

jklukas:
> [bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1637784)
> 
> ```shell
> $ ./bin/generate-pioneer-keys | jq
> {
>   "private_key": {
>     "crv": "P-256",
>     "kty": "EC",
>     "x": "fJD0OoC0WVlNV2HoTVvwiLyCetfGhu4KNKHR6bCDejU",
>     "y": "CogA7ME_pwSf9EL05EMpsfsj7DO2mpYdkRyNwvREFcg"
>   },
>   "public_key": {
>     "crv": "P-256",
>     "d": "ylwWGHdwDaPoCd8ft3eVyaxBUKJNO5o1gZRzAAS7bO4",
>     "kty": "EC",
>     "x": "fJD0OoC0WVlNV2HoTVvwiLyCetfGhu4KNKHR6bCDejU",
>     "y": "CogA7ME_pwSf9EL05EMpsfsj7DO2mpYdkRyNwvREFcg"
>   }
> }
> ```

Note that the example here has public and private keys mixed up. "d" should appear only in the private key. But the code looks correct, so I'm assuming this example was pasted while still drafting the code.

ingestion-beam/bin/generate-pioneer-keys
whd: We've been using _a_ key id which is important to pioneer for routing and we should definitely investigate adding it here. But my assumption has been that the key id doesn't have any effect on encryption and is simply metadata (for routing, see [here](https://github.com/rhelmer/pioneer-v2-example/blob/master/manifest.json#L25) and [here](https://github.com/mozilla-services/cloudops-infra/commit/6c7d7ffe85686190645e728b7772bc00de76281a#diff-b412684ee04868a7b78e063ca9389cac), though they currently don't match). In pioneer the key id we've used is human-readable, and will usually map to the study name since we're using per-study keypairs.

I could be misunderstanding this and we've actually implemented our own key id mechanism separate from the existing one. 

At any rate, if this script spits out a key id in its parameters we can use that as the key id for pioneer too, though it makes some of the [logic I put together](https://github.com/mozilla-services/cloudops-infra/commit/6c7d7ffe85686190645e728b7772bc00de76281a#diff-3b42ed9e701b9a88d1891a516e6af5fe) moot for mapping semantic key ids.

Basically I'm fine with this key id being random, but we should redo our pioneer mechanisms to be like the aet-generated ones (i.e. _not_ having two separate key id mechanisms). I'm also fine with the id not being random if that turns out to make sense.

whd:
We should change the name of this to something generic as we're settling on using it for both AET and pioneer.

ingestion-beam/bin/generate-pioneer-keys
acmiyaguchi: The use of the [`kid` field is optional](https://tools.ietf.org/html/rfc7517#section-4.5) and doesn't follow any particular pattern. After reading the RFC for this particular field, I learned there is a JWK Set that is effectively the `KeyStore` object implemented in the decoder (minus logic for KMS). The `kid` field, if set and applicable, is used to select a key from the JWK Set.

In the case of Pioneer, the key doesn't need to be random, and it doesn't hurt to add the field to the key.

jklukas: It makes sense to me that AET's needs and Pioneer's needs are a bit different here. Pioneer has a concrete rationale for having a set of keys for different purposes, so meaningful names make sense in the Pioneer case. In the AET case, we really only need a single key so there's no meaningful name to give it, and an auto-generated name seems reasonable.

I do think we should embed the `kid` field in the keys so that the payloads clients send also have a `kid` field that we can use in case we end up with a need to disambiguate.

acmiyaguchi:
Re-requesting review from both commenters. I've renamed the script to `generate-jwk-pair`, and added options to add the Key ID to the results.

acmiyaguchi:
:whd pointed out that `key_id` is being written out to the keys instead of `kid` in the spec. I've added an assertion to check that the JWK interface to extract the `key_id` returns a value (which depends on the `kid` field being passed to the constructor).

PR #1295, opened 2020-05-26
Bug 1633525 - reject frecency-update pings as UnwantedData

akkomar:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -115,6 +115,11 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       throw new UnwantedDataException("1631849");
     }
 
+    // First Federated Learning experiment, decommissioned
+    if ("telemetry".equals(namespace) && "frecency-update".equals(docType)) {
```
scholtzan, line 119:
```suggestion
    if (ParseUri.TELEMETRY.equals(namespace) && "frecency-update".equals(docType)) {
```

jklukas: Given that we now have two decommissioned telemetry docTypes we're checking for, it would be good to refactor this similar to how to handle `IGNORED_NAMESPACES`:

```
  private static final Map<String, String> IGNORED_TELEMETRY_DOCTYPES = ImmutableMap
      .<String, String>builder().put("pioneer-study", "1631849")
      .put("frecency-update", "1633525")
      .build();
```

But feel free to defer as out of scope for this PR.

akkomar: Sounds reasonable, I'll push another commit here.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -32,6 +32,11 @@
       .put("org-mozilla-fenix-debug", "1614409") //
       .build();
 
+  private static final Map<String, String> IGNORED_TELEMETRY_DOCTYPES = ImmutableMap
+      .<String, String>builder().put("pioneer-study", "1631849")
+      .put("frecency-update", "1633525")
```
jklukas, line 37:
```suggestion
      .<String, String>builder().put("pioneer-study", "1631849") //
      .put("frecency-update", "1633525") //
```

jklukas:
I went ahead and committed a change to appease spotless, and updated to match master. Hopefully tests will pass now.

PR #1301, opened 2020-06-04
[Bug 1162183] Redact slowSQL for version 41 or older

scholtzan:
Addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1162183

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -131,7 +131,7 @@ public class MessageScrubber {
     if (bug1162183Affected(attributes)) {
       JsonNode payload = json.path("payload");
       if (payload.has("slowSQL")) {
-        ((ObjectNode) payload).replace("slowSQL", NullNode.getInstance());
+        ((ObjectNode) payload).remove("slowSQL");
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -127,6 +127,14 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
         markBugCounter("1602844");
       });
     }
+
+    if (bug1162183Affected(attributes)) {
+      JsonNode payload = json.path("payload");
+      if (payload.has("slowSQL")) {
+        ((ObjectNode) payload).replace("slowSQL", NullNode.getInstance());
```
jklukas, line 134: Will this pass schema validation? [`slowSQL` is defined as type "object"](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/schemas/telemetry/main/main.4.schema.json#L3306-L3349) in the JSON schema. It's allowed to be missing, but may fail to validate if explicitly set to null. Can you check on that behavior?

scholtzan: Just checked, it fails when set to null. Updated the PR to remove the `slowSQL` node instead.

PR #1303, opened 2020-06-08
[Bug 1638902] Reject org-mozilla-fenix-tm namespace

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1638902


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
```diff
@@ -96,6 +96,16 @@ public void testUnwantedDataBug1614412() {
         () -> MessageScrubber.scrub(attributes, Json.createObjectNode()));
   }
 
+  @Test
+  public void testUnwantedDataBug1638902() {
```
fbertsch, line 100: I don't think we need tests for these anymore, but no reason not to keep it.

PR #1304, opened 2020-06-09
Bug 1635893 Implement DecryptAetIdentifiers

jklukas:
Supports Account Ecosystem Telemetry.

This differs somewhat from the design in the open [draft docs PR](https://github.com/mozilla/gcp-ingestion/pull/1264), so those docs will be updated before deploying this.


ingestion-beam/src/main/java/com/mozilla/telemetry/aet/DecryptAetIdentifiers.java
```diff
@@ -0,0 +1,289 @@
...
+      try {
+        if (TELEMETRY_URI_PATTERN.matcher(uri).matches()) {
+          processDesktopTelemetryPayload(json);
+        } else if (false) {
+          // Placeholder condition for handling AET payloads coming from Glean-enabled applications;
+          // design is evolving in https://bugzilla.mozilla.org/show_bug.cgi?id=1634468
+        } else if (STRUCTURED_URI_PATTERN.matcher(uri).matches()) {
+          processStructuredIngestionPayload(json);
+        } else {
+          nonAetDocType.inc();
```
jklukas, line 276: We pass through the message if it doesn't match any of the expected URI patterns. This _shouldn't_ happen if the edge is routing messages correctly. Perhaps it would be better to err on the side of caution here and raise `IllegalAetPayloadException` here so the payload gets sanitized and sent to error. Or I suppose a new `IllegalAetUriException` would be more semantically appropriate.

acmiyaguchi: Sending the message to the error stream may be more actionable to diagnose the problem. Knowing what documents passed through would be useful to determine why the counter goes up. On the other hand, once the document is redacted, it becomes useless for backfill.

ingestion-beam/src/main/java/com/mozilla/telemetry/aet/DecryptAetIdentifiers.java
```diff
@@ -0,0 +1,289 @@
...
+      ObjectNode json;
+      try {
+        json = Json.readObjectNode(message.getPayload());
+      } catch (IOException e) {
+        throw new UnparsableAetPayloadException(e);
+      }
+
+      // Note that we must work with the raw URI because ParseUri can raise errors that would
+      // route payloads containing anon_id values to error output; this transform must be placed
+      // before ParseUri so that we can handle redacting anon_id values.
```
acmiyaguchi, line 265: However, if the values are redacted at this point in the pipeline, it should be fine to pass the uri's into parse uri wholesale?

ingestion-beam/src/main/resources/account-ecosystem/telemetry.schema.json
```diff
@@ -0,0 +1,47 @@
...
+          "description": "Previous Account Ecosystem Telemetry anon_id values associated with this user",
+          "items": {
+            "type": "string"
+          },
+          "type": "array"
+        }
+      },
+      "required": [
+        "ecosystemClientId",
+        "ecosystemDeviceId"
```
rfk, line 29: @jklukas I'm lightly embarrassed to be asking this question, but...what's the difference between "client id" and "device id" here?

I couldn't find a definition of "device id" in [the "identifiers" spec doc](https://docs.google.com/document/d/1rR3uJG8sVtow4plYu6M5Jp5e5zOSrLzdkUx8BdO8kOM), but it does confusing talk about the device being identified by a "client id".

jklukas: It's not obvious to me what the difference is here. I'll try to track down the history of docs to see where these two terms were introduced. I'm definitely open to removing ecosystemDeviceId if it's not meaningfully different from client id.

rfk: I think we should have just `client_id` and remove `device_id`, if that makes sense to you. That's certainly the only such value I had intended on submitting with the first pings from Firefox desktop :-)

jklukas: I have not been able to track down where ecosystem_device_id came from, so I'm inclined to go ahead and remove it. We can easily mark it as non-required immediately, and then I can follow up about recreating the tables to remove it from the schema completely.

rfk: Thanks, SGTM 👍

PR #1306, opened 2020-06-11
Require "payload" in AET telemetry json schema

jklukas:
This will avoid a pipeline error that @whd reported when sending a message with content `{}`.

Was:
```
Error message from worker: java.lang.ClassCastException: com.fasterxml.jackson.databind.node.MissingNode cannot be cast to com.fasterxml.jackson.databind.node.ObjectNode com.mozilla.telemetry.aet.DecryptAetIdentifiers$Fn.processDesktopTelemetryPayload(DecryptAetIdentifiers.java:241)
```

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/aet/DecryptAetIdentifiersTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/aet/DecryptAetIdentifiersTest.java
@@ -228,11 +228,33 @@ public class DecryptAetIdentifiersTest extends TestWithDeterministicJson {
         .newProvider(Resources.getResource("account-ecosystem/metadata-local.json").getPath());
     ValueProvider<Boolean> kmsEnabled = pipeline.newProvider(false);
 
+    String userId = "2ef67665ec7369ba0fab7f802a5e1e04";
+    String anonId = encryptWithTestPublicKey(userId);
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/aet/DecryptAetIdentifiersTest.java
```diff
@@ -221,4 +221,32 @@ public void testErrorOutput() throws Exception {
...
 
+  @Test
+  public void testTelemetryErrorOutput() throws Exception {
+    // minimal test for throughput of a single document
+    ValueProvider<String> metadataLocation = pipeline
+        .newProvider(Resources.getResource("account-ecosystem/metadata-local.json").getPath());
+    ValueProvider<Boolean> kmsEnabled = pipeline.newProvider(false);
+
+    List<String> input = ImmutableList.of("{}", "{\"payload\":{}}",
+        "{\"payload\":{\"ecosystemClientId\":\"foo\",\"ecosystemDeviceId\":\"bar\"}}");
```
acmiyaguchi, line 232: Another test case might be a non-empty payload with a disallowed field like `{"payload": {"ecosystemAnonId": "foobar"}}` which is sent to the error stream with a redacted value.

jklukas: Good call. Adding this case exposed an error in the schema for telemetry which is now fixed in this PR.

PR #1309, opened 2020-06-18
Move PubsubMessageToObjectNode to core and use it from Beam

relud:


ingestion-core/src/test/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNodeBeamTest.java
```diff
@@ -0,0 +1,539 @@
...
+import com.google.cloud.bigquery.LegacySQLTypeName;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.mozilla.telemetry.ingestion.core.util.Json;
+import java.nio.charset.StandardCharsets;
+import java.util.List;
+import java.util.Map;
+import org.junit.Test;
+
+public class PubsubMessageToObjectNodeBeamTest {
```
relud, line 17: formerly `PubsubMessageToTableRowTest`, but it didn't hit anything specific to ingestion-beam. I moved it to core so that it could access methods visible for testing via package scope rather than public scope.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -39,14 +38,16 @@
...
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 /**
- * Transform a {@link PubsubMessage} into an {@link ObjectNode}.
+ * Transform attributes and data into an {@link ObjectNode}.
  */
-public abstract class PubsubMessageToObjectNode implements Function<PubsubMessage, ObjectNode> {
+public abstract class PubsubMessageToObjectNode
+    implements BiFunction<Map<String, String>, byte[], ObjectNode> {
```
relud, line 50: To avoid choosing between beam and google cloud java sdk implementations of `PubsubMessage`, I modified this to accept attributes and data, instead of message.

ingestion-core/src/test/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNodeBeamTest.java
```diff
@@ -0,0 +1,539 @@
...
+  }
+
+  @Test
+  public void testNullRecord() throws Exception {
+    ObjectNode parent = Json.createObjectNode().putNull("record");
+    ObjectNode additionalProperties = Json.createObjectNode();
+    List<Field> bqFields = ImmutableList.of(
+        Field.of("record", LegacySQLTypeName.RECORD, Field.of("field", LegacySQLTypeName.STRING)));
+    TRANSFORM.transformForBqSchema(parent, bqFields, additionalProperties);
+    assertEquals(ImmutableMap.of(), Json.asMap(parent));
```
relud, line 73: this is the only expected test result that changed. Beam would have previously encoded this as `"{"record":null}`, rather than `{}`.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -58,536 +23,91 @@
...
-   *
-   * <p>We include all attributes as fields. It is up to the configured schema for the destination
-   * table to determine which of those actually appear as fields; some of the attributes will be
-   * thrown away.
-   */
-  @VisibleForTesting
-  static TableRow rawTableRow(PubsubMessage message) {
-    TableRow tableRow = new TableRow();
-    message.getAttributeMap().forEach(tableRow::set);
-    tableRow.set(PAYLOAD, message.getPayload());
```
relud, line 148: because this PR converts `ObjectNode` to `TableRow`, the value for `payload` will change from `byte[]` to base64 encoded `String`, and I still need test that this doesn't change the resulting value in BigQuery.

ingestion-beam/src/main/java/com/mozilla/telemetry/util/GzipUtil.java
```diff
@@ -14,14 +13,11 @@
    */
   public static byte[] maybeDecompress(byte[] bytes) {
     try (ByteArrayInputStream payloadStream = new ByteArrayInputStream(bytes);
-        GZIPInputStream gzipStream = new GZIPInputStream(payloadStream);
-        ByteArrayOutputStream decompressedStream = new ByteArrayOutputStream();) {
-      IOUtils.copy(gzipStream, decompressedStream);
-      return decompressedStream.toByteArray();
+        GZIPInputStream gzipStream = new GZIPInputStream(payloadStream)) {
+      return IOUtils.toByteArray(gzipStream);
```
relud, line 17: this is effectively the same, just more consistent within the code base

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
```diff
@@ -58,536 +28,97 @@
...
-   */
-  @VisibleForTesting
-  static String convertNameForBq(String name) {
-    StringBuilder sb = new StringBuilder();
-    if (name.length() > 0 && Character.isDigit(name.charAt(0))) {
-      sb.append('_');
+    @Override
+    public ObjectNode apply(TableId tableId, Map<String, String> attributes, byte[] data) {
+      // attempt to decompress data that may have been compressed to minimize shuffle size
+      return super.apply(tableId, attributes, GzipUtil.maybeDecompress(data));
```
relud, line 121: > This is looking great so far. Integration tests make it look like we're spilling raw bytes, so we must be missing a translation either in the tests or in the implementation here.

we were missing a gzip decompress for payload format, fixed here.

ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -361,7 +361,7 @@ private static PubsubMessageToObjectNode getFormat(Env env) {
         return PubsubMessageToObjectNode.Decoded.of();
       case "payload":
         return PubsubMessageToObjectNode.Payload.of(env.getStrings(STRICT_SCHEMA_DOCTYPES, null),
-            env.getString(SCHEMAS_LOCATION), FileInputStream::new).withOpenCensusMetrics();
+            env.getString(SCHEMAS_LOCATION, null), FileInputStream::new).withOpenCensusMetrics();
```
relud, line 364: ingestion-sink has been updated to support reading table schemas from BigQuery when `schemas_location` is not provided.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubMessageToTableRow.java
relud: confirmed that tests cover this in https://github.com/mozilla/gcp-ingestion/pull/1309/commits/21614b68001c78b2177d4e924668cfe0e15f3393

PR #1310, opened 2020-06-22
[Bug 1642386] Redact payload.syncs.engines.outgoing for old sync pings

scholtzan:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1642386

To reduce the number of validation errors, this adds an additional redaction case to the MessageScrubber. The fields `syncs[...].engines[...].outgoing` get removed from the payload from sync v4 pings sent from clients with FF versions 25 and 26 since they are incompatible with the current schema.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -221,7 +221,7 @@ public class MessageScrubber {
         && "Lockbox".equals(attributes.get(Attribute.APP_NAME))
         && attributes.get(Attribute.APP_VERSION) != null
         && ("1.7.0".equals(attributes.get(Attribute.APP_VERSION))
-            || attributes.get(Attribute.APP_VERSION).matches("1\\.[0-6][0-9.]*"));
+            || attributes.get(Attribute.APP_VERSION).matches("^1\\.[0-6][0-9.]*"));
   }
 
   // See bug 1162183 for discussion of affected versions, etc.
@@ -233,14 +233,15 @@ public class MessageScrubber {
     return ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
         && affectedDocumentTypes.contains(attributes.get(Attribute.DOCUMENT_TYPE))
         && attributes.get(Attribute.APP_VERSION) != null
-        && attributes.get(Attribute.APP_VERSION).matches("([0-3][0-9]|4[0-1])\\..*"); // 41 or older
+        && attributes.get(Attribute.APP_VERSION).matches("^([0-3][0-9]|4[0-1])\\..*"); // >= 41
   }
 
   // See bug 1642386 for discussion of affected versions, etc.
+  @VisibleForTesting
   static boolean bug1642386Affected(Map<String, String> attributes) {
     return ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
         && attributes.get(Attribute.DOCUMENT_TYPE).equals("sync")
         && attributes.get(Attribute.APP_VERSION) != null
-        && attributes.get(Attribute.APP_VERSION).matches("(25|26)\\..*"); // 25 or 26
+        && attributes.get(Attribute.APP_VERSION).matches("^(25|26)\\..*"); // 25 or 26
   }
 }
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
```diff
@@ -271,6 +271,20 @@ public void testBug1162183Affected() throws Exception {
...
+    Map<String, String> baseAttributes = ImmutableMap.<String, String>builder()
+        .put(Attribute.DOCUMENT_NAMESPACE, ParseUri.TELEMETRY).put(Attribute.DOCUMENT_TYPE, "sync")
+        .build();
+
+    assertFalse(MessageScrubber.bug1642386Affected(ImmutableMap.<String, String>builder()
+        .putAll(baseAttributes).put(Attribute.APP_VERSION, "60.6.1").build()));
+    assertTrue(MessageScrubber.bug1642386Affected(ImmutableMap.<String, String>builder()
+        .putAll(baseAttributes).put(Attribute.APP_VERSION, "25.0.1").build()));
+    assertTrue(MessageScrubber.bug1642386Affected(ImmutableMap.<String, String>builder()
+        .putAll(baseAttributes).put(Attribute.APP_VERSION, "26.1").build()));
```
jklukas, line 285: Can you add a test case for "10.25.1"? I don't know that we ever have minor or patch versions as high as 25, but I want to make sure the regex wouldn't accidentally match it.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -226,4 +235,12 @@ static boolean bug1162183Affected(Map<String, String> attributes) {
         && attributes.get(Attribute.APP_VERSION) != null
         && attributes.get(Attribute.APP_VERSION).matches("([0-3][0-9]|4[0-1])\\..*"); // 41 or older
   }
+
+  // See bug 1642386 for discussion of affected versions, etc.
+  static boolean bug1642386Affected(Map<String, String> attributes) {
+    return ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
+        && attributes.get(Attribute.DOCUMENT_TYPE).equals("sync")
+        && attributes.get(Attribute.APP_VERSION) != null
+        && attributes.get(Attribute.APP_VERSION).matches("(25|26)\\..*"); // 25 or 26
```
jklukas, line 244:
```suggestion
        && attributes.get(Attribute.APP_VERSION).matches("^(25|26)\\..*"); // 25 or 26
```

I think `matches` will look for the pattern anywhere in the string, so we might need a `^` to explicitly match the beginning of the string.

scholtzan: `matches` checks for the entire string to match. So the regex as it was wouldn't match something like 10.25.1. But it's probably best to be more explicit to avoid any confusion.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -224,6 +233,15 @@ static boolean bug1162183Affected(Map<String, String> attributes) {
...
+        && attributes.get(Attribute.APP_VERSION).matches("^([0-3][0-9]|4[0-1])\\..*"); // >= 41
+  }
+
+  // See bug 1642386 for discussion of affected versions, etc.
+  @VisibleForTesting
+  static boolean bug1642386Affected(Map<String, String> attributes) {
+    return ParseUri.TELEMETRY.equals(attributes.get(Attribute.DOCUMENT_NAMESPACE))
+        && attributes.get(Attribute.DOCUMENT_TYPE).equals("sync")
+        && attributes.get(Attribute.APP_VERSION) != null
+        && attributes.get(Attribute.APP_VERSION).matches("^(25|26)\\..*"); // 25 or 26
```
acmiyaguchi, line 245: We may want to change so that it catches all versions less than 26, based on the results from https://sql.telemetry.mozilla.org/queries/71851/source#180308. 22, 23, and 24 are the next highest sources of errors. This will take us most of the way there though.

scholtzan: Ah I see. That seems reasonable, I'll open a separate PR for applying this to earlier versions too

PR #1290, opened 2020-05-12
Transform histograms to compact encoding

jklukas:
See [Proposal doc](https://docs.google.com/document/d/1k_ji_1DB6htgtXnPpMpa7gX0klm-DGV5NMY7KkvVB00/edit#)

PR #1313, opened 2020-07-01
Bug 1646825 Compact encoding for use counter histograms

jklukas:
See https://docs.google.com/document/d/1k_ji_1DB6htgtXnPpMpa7gX0klm-DGV5NMY7KkvVB00/edit#

Replaces earlier draft #1290

### DIFF
--- a/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
+++ b/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
@@ -142,6 +142,9 @@ public abstract class PubsubMessageToObjectNode {
       private static final MeasureLong INVALID_HISTOGRAM_SUM = MeasureLong.create(
           "invalid_histogram_sum", //
           "The number of parsed histograms with an invalid sum field", "1");
+      private static final MeasureLong INVALID_HISTOGRAM_USE_COUNTER = MeasureLong.create(
+          "invalid_histogram_use_counter", //
+          "The number of parsed histograms failing use counter invariants", "1");
       private static final MeasureLong INVALID_HISTOGRAM_RANGE = MeasureLong.create(
           "invalid_histogram_range", //
           "The number of parsed histograms with an invalid bucket_count or range field", "1");
@@ -190,14 +193,18 @@ public abstract class PubsubMessageToObjectNode {
       @Override
       protected void incrementInvalidHistogramSum() {
         STATS_RECORDER.newMeasureMap().put(INVALID_HISTOGRAM_SUM, 1).record();
+      }
 
+      /** measure rate of InvalidHistogramUseCounter. */
+      @Override
+      protected void incrementInvalidHistogramUseCounter() {
+        STATS_RECORDER.newMeasureMap().put(INVALID_HISTOGRAM_USE_COUNTER, 1).record();
       }
 
       /** measure rate of InvalidHistogramRange. */
       @Override
       protected void incrementInvalidHistogramRange() {
         STATS_RECORDER.newMeasureMap().put(INVALID_HISTOGRAM_RANGE, 1).record();
-
       }
 
     }
@@ -255,17 +262,18 @@ public abstract class PubsubMessageToObjectNode {
 
     /** measure rate of InvalidHistogramType. */
     protected void incrementInvalidHistogramType() {
-
     }
 
     /** measure rate of InvalidHistogramSum. */
     protected void incrementInvalidHistogramSum() {
+    }
 
+    /** measure rate of InvalidHistogramUseCounter. */
+    protected void incrementInvalidHistogramUseCounter() {
     }
 
     /** measure rate of InvalidHistogramRange. */
     protected void incrementInvalidHistogramRange() {
-
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -595,6 +642,59 @@ private String getAndCacheBqName(String name) {
...
+      JsonNode values = o.path("values");
+      if (histogramType == 4) {
+        // Type 4 histograms are "count" histograms representing only a single value.
+        return TextNode.valueOf(Long.toString(sum));
+      } else if (histogramType == 2 && fieldName.startsWith("use_counter2")) {
+        // "use counters" represent the bulk of histogram data sent in main pings, so we provide
+        // additional optimization for this case; although they are encoded as "boolean" histograms
+        // (type 2), the "0" (false) bucket is never populated, meaning they only encode a single
+        // value.
+        return TextNode.valueOf(Long.toString(sum));
```
relud, line 677: I'd rather not have a duplicate `TextNode.valueOf(Long.toString(sum))` here:
```suggestion
      if (histogramType == 4 || (histogramType == 2 && fieldName.startsWith("use_counter2"))) {
        // Type 4 histograms are "count" histograms representing only a single value.
        // "use counters" represent the bulk of histogram data sent in main pings, so we provide
        // additional optimization for this case; although they are encoded as "boolean" histograms
        // (type 2), the "0" (false) bucket is never populated, meaning they only encode a single
        // value.
        return TextNode.valueOf(Long.toString(sum));
```

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -595,6 +642,59 @@ private String getAndCacheBqName(String name) {
...
+      int histogramType = o.path("histogram_type").asInt(-1);
+      if (histogramType < 0 || histogramType > 5) {
+        incrementInvalidHistogramType();
+        return jsonEncoding;
+      }
+      long sum = o.path("sum").asLong(-1);
+      if (sum < 0) {
+        incrementInvalidHistogramSum();
+        return jsonEncoding;
+      }
```
relud, line 667: no-op: I wish that `jsonEncoding` could be computed lazily here, but I don't think the performance gain would be worth the reduction in readability.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -595,6 +642,59 @@ private String getAndCacheBqName(String name) {
...
+     * <p>We maintain a BigQuery persistent UDF that can parse a variety of formats, so any format
+     * used here must first be supported in:
+     * https://github.com/mozilla/bigquery-etl/blob/master/mozfun/hist/extract/udf.sql
+     *
+     * <p>Schema validation should generally ensure that all histograms are well-formed, but we
+     * perform some light sanity checking on the values and fall back to encoding the histogram
+     * as a JSON string if anything seems off.
+     */
+    private TextNode compactHistogramEncoding(JsonNode o, String fieldName) {
+      TextNode jsonEncoding = TextNode.valueOf(Json.asString(o));
```
relud, line 657: all of the defined variables in this function should be explicitly `final`

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -595,6 +642,59 @@ private String getAndCacheBqName(String name) {
...
+        long rangeLo = o.path("range").path(0).asLong(-1);
+        long rangeHi = o.path("range").path(1).asLong(-1);
+        String valString = Json.asString(values).replace("{", "").replace("}", "").replace("\"",
+            "");
+        if (bucketCount <= 0 || rangeLo < 0 || rangeHi < 0) {
+          incrementInvalidHistogramRange();
+          return jsonEncoding;
+        }
+        return TextNode.valueOf(String.format("%d;%d;%d;%d,%d;%s", //
+            bucketCount, histogramType, sum, rangeLo, rangeHi, valString));
```
relud, line 694: I think we should put `valString` first in this list, rather than last, so that the position in the semicolon separated list consistent.

It's fine as-is if it's too late for this change because we don't want to deviate from the proposal.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
jklukas: I refactored `jsonEncoding` to a function `jsonHistogramEncoding` so that we can avoid computing it unless necessary.

jklukas:
I've run a sample batch beam job on this branch to process one hour of data. None of the new counters were populated, so I'm not expecting any data issues.

I've also kicked off a control using the content of `main` so we can get a specific number for size reduction with this final implementation.

jklukas:
## Size comparison

I ran a control (using the `master` branch) and a treatment (using the content of this PR) batch Dataflow job to process one hour of main ping data (2020-06-30 00:00 to 01:00). The control result table is 421.89 GB and the treatment result is 151.45 GB, so the treatment is 36% the size of the control.

jklukas:
This PR will remain in the draft state until known histogram use cases are adapted to tolerate the new encodings. See dependencies of https://bugzilla.mozilla.org/show_bug.cgi?id=1646825

jklukas:
This PR is now out of draft and ready for final review. I've added a new commit that comments out much of the implementation here so that we can focus on just use counters for "phase 1" of this change. Use counters give us the bulk of savings while impacting a minimal number of analysis use cases.

With the scaled-down version of this PR, we produce a table of size 247 GB compared to 422 GB for the control, so we achieve a size reduction of 42%. Also note that we see 6,217 instances of "invalid" use counters where the sum does not equal the content of the "1" bucket for the sample used in this test.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -595,6 +650,76 @@ private String getAndCacheBqName(String name) {
...
+        // Histograms named as "use_counter" are reported as type 2 (boolean), but only ever have
+        // a non-zero value in the "1" (true) bucket (and the "sum" field should match this count).
+        // They can be encoded as a textual representation of that single number without any loss
+        // of information, and since use counters make up the majority of histogram data volume,
+        // this optimization case is the most important one.
+        // See https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/collection/use-counters.html
+        if (sum == values.path("1").asLong(-1)) {
+          return TextNode.valueOf(Long.toString(sum));
+        } else {
+          incrementInvalidHistogramUseCounter();
```
jklukas, line 686: I was able to track down that all the "invalid" use counters we found were ones that had sum = 0 and an empty values map. These invalids all go away if we catch this case and write out 0 instead. But I think a use counter with value 0 is semantically equivalent to null, so I am considering returning a null node in the case of `sum = 0`.

jklukas: I've pushed an additional commit so that we now encode these as "0".

PR #1318, opened 2020-07-06
Allow complex types to be null when transforming to payload format

### LINKED ISSUE #1317: ingestion-sink undeliverable message: java.lang.ClassCastException: null
/CC @relud, there appears to be a weak correlation between US long weekends and ingestion sink failures...

This is once again telemetry (live) specific.

```
{
 insertId: "of720zlfa2jcu0aps"  
 
jsonPayload: {
  endOfBatch: false   
  
instant: {
   epochSecond: 1593985615    
   nanoOfSecond: 828000000    
  }
  level: "ERROR"   
  loggerFqcn: "org.apache.logging.slf4j.Log4jLogger"   
  loggerName: "com.mozilla.telemetry.ingestion.sink.io.Pubsub$Read"   
  message: "failed to deliver message"   
  thread: "Gax-3"   
  threadId: 26   
  threadPriority: 5   
  
thrown: {
   commonElementCount: 0    
   extendedStackTrace: "java.lang.ClassCastException: null
"    
   name: "java.lang.ClassCastException"    
  }
 }
 
...
 severity: "INFO"  
 timestamp: "2020-07-05T21:46:55.829748951Z"  
}
```

relud:
fixes #1317

### DIFF
--- a/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
+++ b/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
@@ -330,8 +330,12 @@ public abstract class PubsubMessageToObjectNode {
         ObjectNode additionalProperties) {
       final String bqFieldName = field.getName();
 
-      // A record of key and value indicates we need to transformForBqSchema a map to an array.
-      if (isMapType(field)) {
+      // null is valid for any type except an element of a list
+      if (value.isNull()) {
+        updateParent(parent, jsonFieldName, bqFieldName, value);
+
+        // A record of key and value indicates we need to transformForBqSchema a map to an array.
+      } else if (isMapType(field)) {
         expandMapType(jsonFieldName, (ObjectNode) value, field, parent, additionalProperties);
 
         // A record with a single "list" field and a list value should be expanded appropriately.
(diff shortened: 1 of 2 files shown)


ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java
```diff
@@ -330,8 +330,12 @@ private void processField(String jsonFieldName, Field field, JsonNode value, Obj
...
 
-      // A record of key and value indicates we need to transformForBqSchema a map to an array.
-      if (isMapType(field)) {
+      // null is valid for any type except an element of a list
+      if (value.isNull()) {
+        updateParent(parent, jsonFieldName, bqFieldName, value);
+
+        // A record of key and value indicates we need to transformForBqSchema a map to an array.
+      } else if (isMapType(field)) {
         expandMapType(jsonFieldName, (ObjectNode) value, field, parent, additionalProperties);
```
relud, line 339: the exception in #1317 is thrown by this line trying to cast a literal `null` value to `ObjectNode` here.

This PR fixes that by allowing `null` for any `value` passed to `processField`, which I believe is called on every type that is not an element of a list.

The resulting behavior omits the field, which BigQuery will interpret as an empty list if necessary.

PR #1322, opened 2020-07-21
edge disk cleanup service

relud:


ingestion-edge/tests/flush_manager/test_flush_manager.py
```diff
@@ -0,0 +1,174 @@
...
+    print("setting up race condition: attached pvc is also deleted")
+    delete_pvcs(api)
+    print("setting up race condition: pod unschedulable due to missing pvc")
+    with pytest.raises(ApiException) as excinfo:
+        restart_web_pods(api)
+    assert excinfo.value.reason == "Conflict"
+    print("posting message 1")
+    with pytest.raises(requests.exceptions.ConnectionError):
+        requests.post(web, headers={"host": "web"}, json={"id": 1}).raise_for_status()
+
```
acmiyaguchi, line 122: If the flush manager were not run and the pod were recreated, what is the behavior that's undesirable? Resending of a message that has already been flushed once?

relud: if you are asking about prod: messages that are queued to disk and not delivered before shutdown will sit on unattached disks until the stateful set is scaled back up, at which time it will re-attach the disk and flush the messages. this can take days, resulting in very-late delivery of messages.

if you asking about this test: the web pods are configured with an invalid `PUBSUB_EMULATOR_HOST` so that messages are always queued, and only delivered by the jobs the flush manager creates. if the flush manager is not run then the messages are not flushed at all, and the only undesirable behavior I can think of that you might be asking about is that the test would not cover the full behavior of the flush manager.

relud:
> This looks good to me. Is the primary goal to ensure persistent disks aren't reused between pod restarts or to reap disks without an associated job?

reap disks without an associated ~job~ pod.

> And what changes are there in the deployment of the edge service?

the edge will be deployed the same. the flush manager will be a separate kube deployment in the same cluster, plus a handful of other kube resources to grant permissions. I'll be working with whd to get that deployed after this is merged.

In a separate PR the edge will be updated to write immediately to queue, instead of trying pubsub first, and that PR will be blocked on thorough load testing. I'm not including that in this PR, so that we can get this into stage sooner and I can use that to triple check for unexpected behavior in the flush manager, because that's part of this that feels highest risk to me right now.

PR #1323, opened 2020-07-22
Bug 1654558 Implement RerouteDocuments for xfocsp-error-report

jklukas:
Relies on https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/581

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
@@ -11,6 +11,7 @@ import com.mozilla.telemetry.decoder.ParsePayload;
 import com.mozilla.telemetry.decoder.ParseProxy;
 import com.mozilla.telemetry.decoder.ParseUri;
 import com.mozilla.telemetry.decoder.ParseUserAgent;
+import com.mozilla.telemetry.decoder.RerouteDocuments;
 import com.mozilla.telemetry.transforms.DecompressPayload;
 import com.mozilla.telemetry.transforms.LimitPayloadSize;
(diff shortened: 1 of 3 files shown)


whd:
My expectation (for the future mechanism, not for this PR) was that the json schema from the original namespace would be used for decode, and the decoder would apply rerouting based on metadata only on output; separately MSG would only generate the bq schema for the rerouted dataset. At least to me this seems like the least sanity-questioning approach. It occurs to me now that that might break some metadata extraction in the bq provisioning pipeline (that assumes certain conventions about pathing between bq and json schemas), but in any case this PR appears to do the right thing to me, if not in the way I was expecting.

jklukas:
> My expectation (for the future mechanism, not for this PR) was that the json schema from the original namespace would be used for decode, and the decoder would apply rerouting based on metadata only on output; separately MSG would only generate the bq schema for the rerouted dataset. At least to me this seems like the least sanity-questioning approach. It occurs to me now that that might break some metadata extraction in the bq provisioning pipeline (that assumes certain conventions about pathing between bq and json schemas), but in any case this PR appears to do the right thing to me, if not in the way I was expecting.

I agree it would be better to reroute only at the BQ sink step. That, however, would require more complex logic changes since BQ loading can happen either via ingestion-sink for the streaming pipeline or via ingestion-sink in the batch reprocessing case.

I also had imagined that once we have schema metadata in place, we would set `bq_namespace`, etc. as attributes when we get to `ParsePayload` in the decoder, since that's the point in the pipeline where we are already looking up JSON schemas. It would be the cleaner solution code-wise, but misses out on the advantage of applying the routing logic as late as possible.

PR #1328, opened 2020-08-03
Bug 1655477 Support X-Source-Tags header

jklukas:
Will also need a corresponding change in mozilla-pipeline-schemas (see https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/587).

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
@@ -42,7 +42,8 @@ public class AddMetadata {
 
   private static final String HEADER = "header";
   private static final List<String> HEADER_ATTRIBUTES = ImmutableList //
-      .of(Attribute.DATE, Attribute.DNT, Attribute.X_PINGSENDER_VERSION, Attribute.X_DEBUG_ID);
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -42,7 +42,8 @@
 
   private static final String HEADER = "header";
   private static final List<String> HEADER_ATTRIBUTES = ImmutableList //
-      .of(Attribute.DATE, Attribute.DNT, Attribute.X_PINGSENDER_VERSION, Attribute.X_DEBUG_ID);
+      .of(Attribute.DATE, Attribute.DNT, Attribute.X_PINGSENDER_VERSION, Attribute.X_DEBUG_ID,
+          Attribute.X_SOURCE_TAGS);
```
jklukas, line 46: I'm assuming here that we'll include x_debug_id as a simple string field in BigQuery, and potentially update user-facing views to transform the string to a list. This seems cleaner than trying to parse the field to a list within the pipeline.

whd:
For my convenience also update https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-edge/ingestion_edge/config.py#L55

PR #1329, opened 2020-08-03
Add X-Source-Tags to core AddMetadata headers list

jklukas:
Fixup to #1328 since it looks like we have this logic duplicated in core and in beam.

### DIFF
--- a/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
+++ b/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
@@ -31,7 +31,8 @@ public class AddMetadata {
 
   private static final String HEADER = "header";
   private static final List<String> HEADER_ATTRIBUTES = ImmutableList //
-      .of(Attribute.DATE, Attribute.DNT, Attribute.X_DEBUG_ID, Attribute.X_PINGSENDER_VERSION);
+      .of(Attribute.DATE, Attribute.DNT, Attribute.X_DEBUG_ID, Attribute.X_PINGSENDER_VERSION,
+          Attribute.X_SOURCE_TAGS);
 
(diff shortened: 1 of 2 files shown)


ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
```diff
@@ -31,7 +31,8 @@ private AddMetadata() {
 
   private static final String HEADER = "header";
   private static final List<String> HEADER_ATTRIBUTES = ImmutableList //
-      .of(Attribute.DATE, Attribute.DNT, Attribute.X_DEBUG_ID, Attribute.X_PINGSENDER_VERSION);
+      .of(Attribute.DATE, Attribute.DNT, Attribute.X_DEBUG_ID, Attribute.X_PINGSENDER_VERSION,
+          Attribute.X_SOURCE_TAGS);
```
jklukas, line 35: @relud: Should we be removing most of the logic from `ingestion-beam`'s `AddMetadata` in favor of the `core` implementation?

relud: yeah, i'll file a PR on that

relud: https://github.com/mozilla/gcp-ingestion/pull/1330

Dragonrebl (commented):
ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
Dragonrebl: You should be leaving my account alone.

PR #1330, opened 2020-08-03
Move beam's AddMetadata to core

relud:


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/AddMetadataTest.java
```diff
@@ -71,178 +63,4 @@ public void testOutput() {
 
     pipeline.run();
   }
-
```
relud, line 74: everything after here was moved without changes to [ingestion-core/src/test/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadataTest.java](https://github.com/mozilla/gcp-ingestion/pull/1330/files#diff-4175c5b7051ee412b8d6a4a0b7c0b342) below

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
```diff
@@ -16,11 +25,6 @@
  */
 public class AddMetadata {
 
-  private AddMetadata() {
-  }
-
-  public static final String METADATA = "metadata";
```
relud, line 22: moved to `Constants.FieldName.METADATA`

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/AddMetadata.java
```diff
@@ -93,43 +126,113 @@ private AddMetadata() {
...
+
+  ////////
+
+  private static void putAttributes(Map<String, String> attributes, ObjectNode metadata,
+      String nestingKey, String prefix) {
+    Optional //
+        .ofNullable(metadata) //
+        .map(m -> m.path(nestingKey).fields()) //
+        .map(Streams::stream).orElseGet(Stream::empty) //
+        .filter(entry -> !entry.getValue().isNull()) //
```
relud, line 234: this was modified from `.filter(entry -> entry.getValue() != null) //` because `Json.createObject().set("value", (JsonNode) null).get("value")` returns a `NullNode`, so we should never get a literal `null`.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/AddMetadata.java
```diff
@@ -70,11 +31,13 @@
       byte[] metadata;
       try {
         // Get attributes as bytes, throws IOException
-        metadata = Json.asBytes(attributesToMetadataPayload(msg.getAttributeMap()));
+        metadata = Json.asBytes(com.mozilla.telemetry.ingestion.core.transform.AddMetadata
```
jklukas, line 34: We may want to choose a different name for this class in `core`. Perhaps `MetadataPayload` or `NestedMetadata`. That would avoid having this fully-qualified class reference and avoid potential confusion between the two classes in the future.

relud: `NestedMetadata` sounds good to me

PR #1337, opened 2020-08-14
ParseLogEntry metrics and support for additional record type

jklukas:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseLogEntry.java
```diff
@@ -43,27 +50,34 @@ public static ParseLogEntry of() {
...
       }
-      Optional.ofNullable(logEntry.path("insertId").textValue())
-          .ifPresent(id -> attributes.put(Attribute.DOCUMENT_ID,
-              UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8)).toString()));
+      JsonNode fields = logEntry.path("jsonPayload").path("Fields");
+      String documentId = Optional.ofNullable(logEntry.path("insertId").textValue())
+          .map(id -> attributes.put(Attribute.DOCUMENT_ID,
+              UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8)).toString()))
+          .orElse(UUID.randomUUID().toString());
+      attributes.put(Attribute.DOCUMENT_ID, documentId);
```
relud, line 58: if i understand correctly, this only needs to put `DOCUMENT_ID` once:
```suggestion
          .map(id -> UUID.nameUUIDFromBytes(id.getBytes(StandardCharsets.UTF_8)).toString())
          .orElse(UUID.randomUUID().toString());
      attributes.put(Attribute.DOCUMENT_ID, documentId);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseLogEntry.java
```diff
@@ -43,27 +50,34 @@ public static ParseLogEntry of() {
...
       Optional.ofNullable(logEntry.path("receiveTimestamp").textValue())
           .ifPresent(v -> attributes.put(Attribute.SUBMISSION_TIMESTAMP, v));
-      Optional.ofNullable(logEntry.path("jsonPayload").path("Fields").path("userAgent").textValue())
+      Optional.ofNullable(fields.path("userAgent").textValue())
           .ifPresent(v -> attributes.put(Attribute.USER_AGENT, v));
-      String ecosystemAnonId = logEntry.path("jsonPayload").path("Fields").path("ecosystemAnonId")
-          .textValue();
-      String event = logEntry.path("jsonPayload").path("Fields").path("event").asText();
+      String event = fields.path("event").asText();
+      String ecosystemAnonId = fields.path("ecosystemAnonId").textValue();
```
relud, line 64: why `.asText()` for `event`, but `.textValue()` for `ecosystemAnonId`?

jklukas: `asText` returns `""` in case the node doesn't exist. `textValue` returns null. It's a matter of whether we're going to inspect the null case or not. There's definite room for improvement here and a more coherent approach to parsing as we better understand the scope of different record types we'll need to ingest and how much of the structure we want to parse.

PR #1338, opened 2020-08-14
Add error output support to ingestion-sink

relud:
this enables configuring a dead-letter-queue for undeliverable messages, which is an important feature for working around pubsub lite nack behavior.


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/config/SinkConfig.java
```diff
@@ -116,23 +121,28 @@
...
   protected static class Output implements Function<PubsubMessage, CompletableFuture<Void>> {
 
-    private final Env env;
     private final OutputType type;
     private final Function<PubsubMessage, CompletableFuture<Void>> write;
 
-    private Output(Env env, OutputType outputType,
-        Function<PubsubMessage, CompletableFuture<Void>> write) {
-      this.env = env;
+    private Output(OutputType outputType, Function<PubsubMessage, CompletableFuture<Void>> write) {
```
relud, line 134: `env` has been split up into input/output/error, and a side-effect is that the `env` used in `getOutput` no longer needs to be kept around for use by `getInput`

PR #1341, opened 2020-08-19
Bug 1646151 - Handle per-study pioneer deletion request

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1646151)

This requires a schema per study named `deletion-request.1` and is the empty schema. See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/588

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
@@ -118,7 +118,7 @@ public class DecryptPioneerPayloads extends
       JsonNode payload = json.get(FieldName.PAYLOAD);
 
       byte[] payloadData;
-      if (payload.get(SCHEMA_NAME).equals(DELETION_REQUEST_SCHEMA_NAME)) {
+      if (payload.get(SCHEMA_NAME).asText().equals(DELETION_REQUEST_SCHEMA_NAME)) {
         // The deletion-request ping is special cased within the decryption
         // pipeline. The Pioneer client requires an encrypted payload in the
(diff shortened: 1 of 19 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecryptPioneerPayloads.java
```diff
@@ -115,21 +117,31 @@ private DecryptPioneerPayloads(ValueProvider<String> metadataLocation,
...
-      if (decompressPayload.get()) {
-        payloadData = GzipUtil.maybeDecompress(decrypted);
+      if (payload.get(SCHEMA_NAME).asText().equals(DELETION_REQUEST_SCHEMA_NAME)) {
+        // The deletion-request ping is special cased within the decryption
+        // pipeline. The Pioneer client requires an encrypted payload in the
+        // payload, and the server (this transform in particular) requires the
+        // payload to contain encrypted data. We only require the injected
+        // metadata from the envelope to shred all data related to a client, so
+        // the encrypted data in the deletion-request ping is thrown away.
+        payloadData = "{}".getBytes(Charsets.UTF_8);
```
akkomar, line 128: Why do we need to clear the payload manually here?

Deletion-request schema specifies an empty document and would be used to validate these pings later on in the transformation chain. IIUC this guarantees an empty payload anyway.

acmiyaguchi: The client is required to encrypt something locally, which ends up being "" with a throwaway key. We need to replace the payload because of this.

akkomar (approved):

PR #604, opened 2019-05-15
Implement user agent parsing with exact aws behavior

relud:
fixes #556

PR #1366, opened 2020-09-01
Remove unused yauaa dependency and fix shaded import

relud:


jklukas (approved):
Looks like this is following up on https://github.com/mozilla/gcp-ingestion/pull/604#issuecomment-492466673

LGTM

PR #1374, opened 2020-09-01
Bug 1648208 Use jwe_mappings from mozPipelineMetadata for AET decryption

jklukas:
This cleans up the code and paves the way for supporting AET from Glean, where apps will declare field mappings as part of Glean metrics configuration.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -140,18 +140,14 @@ private ParsePayload(ValueProvider<String> schemasLocation) {
 
       // If no "document_version" attribute was parsed from the URI, this element must be from the
       // /submit/telemetry endpoint and we now need to grab version from the payload.
-      if (!attributes.containsKey("document_version")) {
```
jklukas, line 143: This logic is refactored to a public function here so that we can reuse it for AET.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/schema/PipelineMetadataStore.java
```diff
@@ -0,0 +1,134 @@
...
+
+/**
+ * This schema store uses the same mozilla-pipeline-schemas tarball as JSONSchemaStore,
+ * but instead of returning Schema instances to be used for validation, it returns
+ * {@link PipelineMetadata} objects parsed from each schema's "mozPipelineMetadata" entry.
+ */
+public class PipelineMetadataStore extends SchemaStore<PipelineMetadataStore.PipelineMetadata> {
+
+  @AutoValue
+  @JsonDeserialize(builder = AutoValue_PipelineMetadataStore_JweMapping.Builder.class)
```
jklukas, line 25: Everywhere else in this repo, we interact with JSON via `JsonNode` and the tree interface, but this `PipelineMetadataStore` uses Jackson databind to instantiate a class defined with AutoValue. It's definitely cleaner to work with this structure, since all the parsing happens centrally. But, it's adding a new concept to the codebase, so I'm open to feedback there.

This structure is particularly convenient because it handles constructing concrete `JsonPointer` instances as part of the cached PipelineMetadata objects, whereas we would likely have to reconstruct those every time or maintain a separate cache if we returned an `ObjectNode` here.

acmiyaguchi: This effectively turns the json node into an object with known fields at compile time?

acmiyaguchi (approved):
r+

It's nice to see minimal test changes during a refactor.

ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/schema/PipelineMetadataStore.java
jklukas: > This effectively turns the json node into an object with known fields at compile time?

Exactly. Using Jackson databind is essentially a validation step and a construction step. Jackson will throw an error if it finds incompatible types between what's present in the JSON and what's defined in the class fields, for example.

PR #1399, opened 2020-09-10
Format markdown with prettier and enforce in CI

relud:


bin/format-markdown
```diff
@@ -0,0 +1,12 @@
...
+
+set -e
+
+cd "$(dirname "$0")/.."
+
+docker run -it --rm \
+    --volume $PWD:/gcp-ingestion \
+    --workdir /gcp-ingestion \
+    node:alpine \
+    sh -c 'npm install -g prettier && npx prettier -w "**/*.md"'
```
wlach, line 12: If we ever have multiple node dependencies for these things, it might be simpler to put a `package.json` in the root to make it simpler to install/use all of them locally. I think this is a fine approach for now though.

wlach (approved):

PR #1462, opened 2020-10-27
Bug 1673489 Remove special routing for xfocsp-error-report

jklukas:
We now have table-level ACLs in place that allow us to write into the normal table under `telemetry_live`.


relud (approved):

whd (approved):
I'll merge this on Monday next week.

PR #1486, opened 2020-11-17
[Bug 1673433] Implement URI based scrubbing for sslreports URIs

scholtzan:
https://bugzilla.mozilla.org/show_bug.cgi?id=1673433

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -167,7 +167,7 @@ public class MessageScrubber {
    *
    * <p>May throw an exception as a signal to route the message to error output or to be dropped.
    */
-  public static void scrubByUri(String uri) throws AffectedByBugException {
+  public static void scrubByUri(String uri) {
     if (IGNORED_URIS.containsKey(uri)) {
       throw new UnwantedDataException(IGNORED_URIS.get(uri));
     }
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -51,6 +51,11 @@ public class MessageScrubber {
   private static final ImmutableSet<String> FIREFOX_ONLY_DOCTYPES = ImmutableSet.of("event", "main",
       "modules");
 
+  private static final Map<String, String> IGNORED_URIS = ImmutableMap.<String, String>builder()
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -51,6 +51,11 @@
   private static final ImmutableSet<String> FIREFOX_ONLY_DOCTYPES = ImmutableSet.of("event", "main",
       "modules");
 
+  private static final Map<String, String> IGNORED_URIS = ImmutableMap.<String, String>builder()
+      .put("/submit/sslreports", "1585144") //
+      .put("/submit/sslreports/", "1585144") //
```
scholtzan, line 56: See https://bugzilla.mozilla.org/show_bug.cgi?id=1667827

Currently `/submit/sslreports` is dropped and `/submit/sslreports/` isn't, but it's probably best to scrub both just in case.

jklukas: I wonder if URI prefixes would be the more generally interesting target rather than specific URIs, in which case this would reduce to just "/submit/sslreports". But it would be easy to refactor this in the future if we end up with additional cases where we want to scrub based on prefix, so this looks fine as-is.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -157,6 +162,17 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
     }
   }
 
+  /**
+   * Check the message URI against known URIs of messages with potentially harmful data.
+   *
+   * <p>May throw an exception as a signal to route the message to error output or to be dropped.
+   */
+  public static void scrubByUri(String uri) throws AffectedByBugException {
```
jklukas, line 170:
```suggestion
  public static void scrubByUri(String uri) {
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -76,6 +77,8 @@
 
       // parse uri based on prefix
       final String uri = attributes.get(Attribute.URI);
+      MessageScrubber.scrubByUri(uri);
```
jklukas, line 80:
```suggestion

      // Throws an exception on certain URI patterns to signal that the message should be rejected.
      MessageScrubber.scrubByUri(uri);
```

jklukas (approved):
r+wc

Thanks for grabbing this.

PR #1490, opened 2020-11-18
[Bug 1673433] Scrub v2/old pings

scholtzan:
This is a follow-up to https://bugzilla.mozilla.org/show_bug.cgi?id=1673433. Some of the tests need to be updated, but I wanted to make sure that the general approach will actually work.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -167,9 +167,14 @@ public class MessageScrubber {
    *
    * <p>May throw an exception as a signal to route the message to error output or to be dropped.
    */
-  public static void scrubByUri(String uri) {
+  public static void scrubByUri(String uri, Map<String, String> attributes) {
     if (IGNORED_URIS.containsKey(uri)) {
       throw new UnwantedDataException(IGNORED_URIS.get(uri));
+    }
+
+    // Ignore v2 / other old ping buckets
+    if (bug1673433Affected(uri, attributes)) {
+      throw new UnwantedDataException("1673433");
     }
   }
 
@@ -268,5 +273,31 @@ public class MessageScrubber {
         && attributes.get(Attribute.DOCUMENT_TYPE).equals("sync")
         && attributes.get(Attribute.APP_VERSION) != null
         && attributes.get(Attribute.APP_VERSION).matches("^([0-9]|[0-2][0-7])\\..*"); // <= 27
+  }
+
+  // See bug 1673433 for discussion of affected URIs
+  private static boolean bug1673433Affected(String uri, Map<String, String> attributes) {
+    // See: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/
+    // k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41
+    if (uri == null) {
+      return false;
+    }
+
+    final ImmutableSet<String> allowedDocTypes = ImmutableSet.of("/main/", "/core/", "/sync/",
+        "/sslreports", "/FirefoxOS/");
+
+    boolean dropPing = true;
+    for (String docType : allowedDocTypes) {
+      if (uri.contains(docType)) {
+        dropPing = false;
+        break;
+      }
+    }
+
+    if (attributes.get(Attribute.ARGS).contains("v=4")) {
+      dropPing = false;
+    }
+
+    return dropPing;
   }
 }
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
@@ -79,7 +79,7 @@ public class ParseUri {
       final String uri = attributes.get(Attribute.URI);
 
       // Throws an exception on certain URI patterns to signal that the message should be rejected.
-      MessageScrubber.scrubByUri(uri);
+      MessageScrubber.scrubByUri(uri, attributes);
 
       if (uri == null) {
         // We should only have a missing uri attribute if we're replaying messages from decoded
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
@@ -527,11 +527,34 @@ public class MessageScrubberTest {
 
   @Test
   public void testUnwantedData1585144() {
-    MessageScrubber.scrubByUri(null);
-    MessageScrubber.scrubByUri("/foo/bar");
+    Map<String, String> attributes = ImmutableMap.<String, String>builder()
+        .put(Attribute.ARGS, "v=4").build();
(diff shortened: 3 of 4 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -269,4 +274,30 @@ static boolean bug1642386Affected(Map<String, String> attributes) {
...
+        break;
+      }
+    }
+
+    if (attributes.get(Attribute.ARGS).contains("v=4")) {
+      dropPing = false;
+    }
+
+    return dropPing;
+  }
```
scholtzan, line 302: This tries to re-implement the logic from https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41

Just to make sure, will the `v=4` URI arg be part of `args` in `attributes`?

jklukas: Yes, I believe this is correct, that we end up with an "args" key in `attributes`.

jklukas: This is indeed documented in https://mozilla.github.io/gcp-ingestion/architecture/edge_service_specification/

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -269,4 +274,30 @@ static boolean bug1642386Affected(Map<String, String> attributes) {
...
+
+    boolean dropPing = true;
+    for (String docType : allowedDocTypes) {
+      if (uri.contains(docType)) {
+        dropPing = false;
+        break;
+      }
+    }
+
+    if (attributes.get(Attribute.ARGS).contains("v=4")) {
```
jklukas, line 297: I'd like to see us properly parse the args using a library function build for this purpose. I would not be surprised if there are weird ways this can go wrong. I wouldn't be surprised if we already import a library with a reasonable implementation. Apache HTTP has URLEncodedUtils; not sure if that's available here.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -269,4 +274,30 @@ static boolean bug1642386Affected(Map<String, String> attributes) {
         && attributes.get(Attribute.APP_VERSION) != null
         && attributes.get(Attribute.APP_VERSION).matches("^([0-9]|[0-2][0-7])\\..*"); // <= 27
   }
+
+  // See bug 1673433 for discussion of affected URIs
+  private static boolean bug1673433Affected(String uri, Map<String, String> attributes) {
+    // See: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/
+    // k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41
```
jklukas, line 281: Is it possible to not break this across two lines? I have a strong preference for this being directly copy-pastable.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -269,4 +274,30 @@ static boolean bug1642386Affected(Map<String, String> attributes) {
...
+    }
+
+    final ImmutableSet<String> allowedDocTypes = ImmutableSet.of("/main/", "/core/", "/sync/",
+        "/sslreports", "/FirefoxOS/");
+
+    boolean dropPing = true;
+    for (String docType : allowedDocTypes) {
+      if (uri.contains(docType)) {
+        dropPing = false;
+        break;
```
jklukas, line 293: We have a combination of early return (null case above) and breaking out of loops. I'd prefer to use early returns consistently in this case. So we wouldn't have `dropPing` at all.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -269,4 +274,30 @@ static boolean bug1642386Affected(Map<String, String> attributes) {
...
+  // See bug 1673433 for discussion of affected URIs
+  private static boolean bug1673433Affected(String uri, Map<String, String> attributes) {
+    // See: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/
+    // k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41
+    if (uri == null) {
+      return false;
+    }
+
+    final ImmutableSet<String> allowedDocTypes = ImmutableSet.of("/main/", "/core/", "/sync/",
+        "/sslreports", "/FirefoxOS/");
```
jklukas, line 287: This should be moved to a class-level constant so it doesn't have to be created on every call.

jklukas, line 287: If I recall correctly, we now catch the "sslreports" and "FirefoxOS" cases as unwanted data in other places within MessageScrubber. I worry now that we're adding confusion by explicitly allowing these here, but then catching them separately elsewhere.

It's probably the right thing to separate these out for having the right bug attributed. But I think it warrants some additional commentary in the code. I'd like to see commentary on this object specifically to call out that `sslreports` and `FirefoxOS` are thrown out elsewhere, but allowed here for historical consistency.

jklukas (commented):
This logic looks correct. Verifying it is exhausting. Every time I have read https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41 I have to read it over at least five times to convince myself I've properly parsed the negations and nils.

My comments here are mostly style and documentation. I'd like to read it over again after changes are made for safety.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
whd: > This logic looks correct. Verifying it is exhausting. Every time I have read https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-ingestion/k8s/charts/data-ingestion/templates/filter-configmap.yaml#L14-L41 I have to read it over at least five times to convince myself I've properly parsed the negations and nils.

I [commented](https://bugzilla.mozilla.org/show_bug.cgi?id=1673433#c4) that another option here is to avoid translating this logic wholesale and instead re implement the still relevant pieces (that haven't already been implemented) as-needed by observing the garbage we begin to receive when removing the edge filter.

As-is this logic should be a noop on URIs that made it to `payload_bytes_raw`.

scholtzan: +1 to only re-implementing the pieces we actually need. This will hopefully simplify the logic and keep it more maintainable in the future.

jklukas:
Based on https://bugzilla.mozilla.org/show_bug.cgi?id=1673433#c6 it sounds like we are not going to pursue this. Can we close this PR?

PR #1494, opened 2020-11-21
Add build option to ingestion-sink to create standalone jar

acmiyaguchi:
This generates a standalone jar with all of the dependencies included for use in mozilla-pipeline-schemas for CI. See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/645.

The jars from each build should get uploaded to a public GCS bucket so it can be fetched the the mps docker container. The jar is ~40mb.

```
% ls -alh target/*.jar
-rw-r--r--  1 amiyaguchi  staff    40M Nov 30 14:20 target/ingestion-sink-0.1-SNAPSHOT-jar-with-dependencies.jar
-rw-r--r--  1 amiyaguchi  staff   151K Nov 30 14:20 target/ingestion-sink-0.1-SNAPSHOT.jar
```

### DIFF
--- a/ingestion-sink/pom.xml
+++ b/ingestion-sink/pom.xml
@@ -172,6 +172,26 @@
                     <skip>false</skip>
                 </configuration>
             </plugin>
+            <!-- Create an uber jar for standalone usage. https://stackoverflow.com/a/16222971 -->
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-assembly-plugin</artifactId>
+                <version>3.3.0</version>
+                <configuration>
+                    <descriptorRefs>
+                        <descriptorRef>jar-with-dependencies</descriptorRef>
+                    </descriptorRefs>
+                </configuration>
+                <executions>
+                    <execution>
+                        <id>assemble-all</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
         </plugins>
     </build>
 </project>


ingestion-sink/pom.xml
```diff
@@ -172,6 +172,26 @@
...
+                <executions>
+                    <execution>
+                        <id>assemble-all</id>
+                        <phase>package</phase>
+                        <goals>
+                            <goal>single</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
```
relud, line 194: Uber jars break grpc calls, per [#1283 comment](https://github.com/mozilla/gcp-ingestion/pull/1283/files#r422409954):

> Dependencies have been moved to jars included via this classpath prefix to fix an issue where grpc is breaking because a `jar-with-dependencies` isn't able to include `META-INF/services` from every dependency.

relud (commented):
> The jars from each build should get uploaded to a public GCS bucket so it can be fetched the the mps docker container. The jar is ~40mb.

instead of fetching from GCS, I recommend doing `COPY --from=mozilla/ingestion-sink:latest /app/ingestion-sink/target/* /some/destination/` (only the jars are present in that directory, everything else is removed after building)

That will copy the normal jar and the dependency directory, which is used via a path-relative reference in the normal jar's classpath per [pom.xml#L240](https://github.com/mozilla/gcp-ingestion/blob/master/pom.xml#L240).

acmiyaguchi:
> > The jars from each build should get uploaded to a public GCS bucket so it can be fetched the the mps docker container. The jar is ~40mb.
> 
> instead of fetching from GCS, I recommend doing `COPY --from=mozilla/ingestion-sink:latest /app/ingestion-sink/target/* /some/destination/` (only the jars are present in the `target` directory, everything else is removed from the image after building the jars)
> 
> That will copy the normal jar and the dependency directory, which is used via a path-relative reference in the normal jar's classpath per [pom.xml#L240](https://github.com/mozilla/gcp-ingestion/blob/master/pom.xml#L240).

Ooh, I had no idea that this was a thing that I could do.

acmiyaguchi:
The image doesn't seem to ship with dependencies, but I managed to get this working with the following stage in the docker file:

```dockerfile
FROM mozilla/ingestion-sink:latest as ingestion-sink
RUN mvn dependency:copy-dependencies
COPY --from=mozilla/ingestion-sink:latest /app/ingestion-sink/target /app/target

FROM ...
...
COPY --from=ingestion-sink /app/ingestion-sink/target /app/target
```

Closing this to avoid introducing new infrastructure burdens.

relud:
> The image doesn't seem to ship with dependencies

it has to, or it wouldn't run

acmiyaguchi:
I would think so, but I only see the standalone jar in the application's home directory. If I run `mvn dependency:copy-dependencies`, it copies down quite a few jars into the `target/dependency` folder, which I expect since this depends on things like `google-cloud-bigquery`. 

```
% docker run -it mozilla/ingestion-sink:latest bash
root@061f4e7518c6:/app/ingestion-sink# ls -alh target
total 144K
drwxr-xr-x 3 root root 4.0K Aug 12 21:38 .
drwxr-xr-x 1 root root 4.0K Aug 12 21:38 ..
-rw-r--r-- 1 root root 131K Aug 12 21:38 ingestion-sink-0.1-SNAPSHOT.jar
drwxr-xr-x 2 root root 4.0K Aug 12 21:38 ingestion-sink-0.1-SNAPSHOT.lib
```

relud:
the dependencies are in the `ingestion-sink-0.1-SNAPSHOT.lib` directory there

PR #1527, opened 2021-01-05
Detect unschedulable pods more aggressively in the flush manager

relud:
unblocks [Bug 1666498](https://bugzilla.mozilla.org/show_bug.cgi?id=1666498)


relud:
@whd After this is merged I'd like to see it run in stage through a few deploys to make sure it fixes the observed issue and doesn't introduce a new one.

I expected this to be harder than an `any` clause, but I can't seem to find any evidence of an issue with it and the integration tests work fine.

whd (approved):
I'll run a few redeploys in stage to verify the efficacy of this before rolling out to prod.

PR #1541, opened 2021-01-20
Flush manager fix

relud:
needs tests


relud:
whoops, branch collision

PR #1542, opened 2021-01-20
Add --pubsubIdAttribute beam option for dedup when reading pubsub

relud:
for [Bug 1627758](https://bugzilla.mozilla.org/show_bug.cgi?id=1627758)

~needs tests~

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
@@ -11,6 +11,7 @@ import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nullable;
 import org.apache.avro.generic.GenericRecord;
 import org.apache.beam.sdk.io.TextIO;
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO;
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -35,15 +35,25 @@
...
     @Override
     public PCollection<PubsubMessage> expand(PBegin input) {
+      final PubsubIO.Read<PubsubMessage> readBase = PubsubIO
+          .readMessagesWithAttributesAndMessageId();
+      final PubsubIO.Read<PubsubMessage> read;
+      if (idAttribute == null) {
+        read = readBase;
+      } else {
+        read = readBase.withIdAttribute(idAttribute);
+      }
```
jklukas, line 54: `getIdAttribute` is marked as `@Nullable` in PubsubIO, which means `idAttribute` is allowed to be null. I think this means it's also possible to pass in `null` to `withIdAttribute`, so this logic is hopefully unnecessary.

relud: yep, works as expected when null.

relud:
@whd this is going to be a pre-template option, because the SDK doesn't accept a `ValueProvider` here.

whd:
> @whd this is going to be a pre-template option, because the SDK doesn't accept a `ValueProvider` here.

Can you add it to https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/bin/build-template#L55 so it gets picked up automatically?

relud:
> add it to https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/bin/build-template#L55

done

jklukas (approved):

jklukas:
Up to @whd as to whether we should delay merging right now, or if you'd like this to go to staging right away.

whd:
> Up to @whd as to whether we should delay merging right now, or if you'd like this to go to staging right away.

Let's merge this on Monday. The tentative plan:

1. merge on the 25th, auto-deploy to stage, observe stage
2. deploy prod (telemetry decoder only) before UTC 26th
3. observe performance of the decoder, run until at least UTC 27th
4. inspect dedup performance from e.g. pbe for the 26th
5. decide whether to roll forward the structured decoder or roll back the telemetry decoder
6. assuming roll forward, continue deploys as per usual

whd:
I'm holding off on testing this until an unrelated support case with Google is resolved.

whd:
Looks like the issue is resolved, so we can set this test up next week.

whd:
I'm pushing this test out to next week (UTC Feb 17) due to https://github.com/mozilla/gcp-ingestion/issues/1560 and because I'll be out after Wednesday this week.

whd:
I'm planning to merge this after https://github.com/mozilla/gcp-ingestion/pull/1566 is deployed.

Issue #1534, opened 2021-01-14
Use Dataflow Flex Templates and remove use of ValueProvider


jklukas:
[Dataflow Flex templates](https://cloud.google.com/dataflow/docs/concepts/dataflow-templates) are now GA. They allow publishing a docker container that compiles the pipeline at runtime so that any configuration parameter can be specified at runtime, not just those that are marked as ValueProvider.

This would allow us to clean up the code a bit, and would provide more flexibility in how we define new parameters in the future. There are a few [documented downsides](https://cloud.google.com/dataflow/docs/concepts/dataflow-templates#comparing-templated-jobs) which at time of writing are lack of support for FlexRS (which we have not so far used) and lack of support for updating streaming jobs (which we also don't do).

This may not be worth prioritizing until we have some clear motivation to do so. It would require a close collaboration between data eng and ops, since it completely changes the workflow for publishing a template and deploying a job.

cc @whd

relud:
This is going to be needed for Pub/Sub Lite IO, because it doesn't support ValueProvider.

Also, flex templates are a different resource type in terraform: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/dataflow_flex_template_job

whd:
It looks like we already have a use case and flex templates was on :jason's stretch goals to investigate last year.

This does completely change the workflow and will require coordination, but that coordination will be necessary for the use case being investigated anyway (pubsub lite). Of particular note, we're using a custom build of terraform 0.11 for beam and shared-prod right now. Moving to modern terraform is planned for this quarter, but is currently blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1682749 for data-shared. Almost assuredly upstream terraform by now supports the custom options we added to our custom provider for beam, but I will need to go through and verify that as part of the update process here.

As a side note: :jason and I confirmed last year that the standard dataflow resource in modern terraform *does* support updates operations now, and we've used it to effect in the opsec pipeline. For the purposes of the standard ingestion pipeline on 0.11 however, moving to flex templates should be equivalent to the current system.

jklukas:
Also note that Beam supports Java 11 as of a recent release (I believe 2.29). It may make sense to roll this into our change to flex template deploys, using the Java 11 base image.

whd:
This is being tracked in [DSRE-79](https://mozilla-hub.atlassian.net/browse/DSRE-79).

whd:
[...]

PR #1549, opened 2021-01-26
Add support for publishing Dataflow flex templates

relud:
cc @whd

~needs automated tests, but it is currently working~ edit: all code changes were removed


ingestion-beam/src/main/java/com/mozilla/telemetry/Sink.java
```diff
@@ -67,4 +77,60 @@ static void registerOptions() {
...
+          final Default.Integer defaultAnnotation = method.getAnnotation(Default.Integer.class);
+          defaultValue = defaultAnnotation == null ? null
+              : Integer.toString(defaultAnnotation.value());
+        } else if (long.class.equals(returnType) || Long.class.equals(returnType)) {
+          regex = "[0-9]+";
+          final Default.Long defaultAnnotation = method.getAnnotation(Default.Long.class);
+          defaultValue = defaultAnnotation == null ? null
+              : Long.toString(defaultAnnotation.value());
+        } else {
+          regex = ".*";
```
relud, line 119: parameters that use `ValueProvider` get this regex instead of checking their subtype, but I figure that's fine because #1534 includes removing `ValueProvider` stuff.

ingestion-beam/src/main/java/com/mozilla/telemetry/Sink.java
```diff
@@ -67,4 +77,60 @@ static void registerOptions() {
...
 
+  static class FlexTemplateMetadata {
+
+    public static void main(String[] args) {
+      run(SinkOptions.class);
+    }
+
+    static void run(Class<? extends PipelineOptions> optionsClass) {
+      final ArrayNode parameters = Json.createArrayNode();
+      for (Method method : optionsClass.getMethods()) {
```
relud, line 88: I wasn't sure about using reflection here, but I looked up how the help page is generated and all the methods were private and use reflection, so I figure that's just what we get.

jklukas: Can you add a comment pointing to that help page code?

jklukas:
I'll need to set aside some time to read through the flex templates docs carefully to fully review here. It may not be until next week. Is that fine?

relud:
> I'll need to set aside some time to read through the flex templates docs carefully to fully review here. It may not be until next week. Is that fine?

yeah, no rush here

ingestion-beam/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+COPY pom.xml /app/pom.xml
+COPY ingestion-beam/pom.xml /app/ingestion-beam/pom.xml
+COPY checkstyle /app/checkstyle
+COPY ingestion-beam/checkstyle /app/ingestion-sink/checkstyle
+RUN mvn prepare-package
+COPY ingestion-core/src /app/ingestion-core/src
+COPY ingestion-beam/src /app/ingestion-beam/src
+RUN mvn package -Dmaven.test.skip=true
+
+FROM gcr.io/dataflow-templates-base/java8-template-launcher-base:latest
```
jklukas, line 13:
```suggestion
# This is the base image for the final image we're building.
FROM gcr.io/dataflow-templates-base/java8-template-launcher-base:20210120_RC00
```

It is specifically recommended not to choose latest: https://cloud.google.com/dataflow/docs/guides/templates/configuring-flex-templates

ingestion-beam/Dockerfile
```diff
@@ -0,0 +1,20 @@
+
```
jklukas, line 1:
```suggestion
# We use a maven image to build the jar that will be included via COPY later.
```

ingestion-beam/bin/build-flex-templates
```diff
@@ -0,0 +1,27 @@
...
+  export IMAGE=gcr.io/$PROJECT/ingestion-beam/$JOB_TYPE:$BUILD_ID
+  docker-compose build --build-arg FLEX_TEMPLATE_JAVA_MAIN_CLASS=com.mozilla.telemetry.${JOB_CLASS}
+  docker-compose push
+
+  METADATA_FILE=metadata.$JOB_TYPE.json
+  docker-compose run --rm flex-template > $METADATA_FILE
+
+  gcloud dataflow flex-template build \
+    gs://$TEMPLATE_BUCKET/$JOB_TYPE/flex-templates/$BUILD_ID.json \
+    --image $IMAGE \
```
jklukas, line 24:
```suggestion
    --image-gcr-path $IMAGE \
```

This is `image-gcr-path` in the tutorial. I'm not sure whether there's a difference here.

ingestion-beam/bin/build-flex-templates
```diff
@@ -0,0 +1,27 @@
...
+    --sdk-language JAVA \
+    --metadata-file $METADATA_FILE
```
jklukas, line 26:
```suggestion
    --flex-template-base-image JAVA8 \
    --metadata-file $METADATA_FILE
```

It would be good to make this explicit.

jklukas (commented):
Leaving a round of comments. There are a bunch of new concepts to wrap my head around to understand what the flow for publishing a flex template, but so far this very good. I'll hopefully to do final review later this afternoon.

It's disappointing that the metadata.json file is necessary; it seems like extra overhead to understand. I think there's something similar with legacy templates, but somehow we were able to avoid it.

jklukas:
This is looking reasonable. It's not clear if `metadata.json` is necessary and I was going to suggest we cut out that piece due to the reflection needs, but the amount of code necessary is pretty small, so probably fine to keep. It's not totally clear to me what benefit we get from the `metadata.json`, though. It seems mostly just relevant for Google's published templates being more user-friendly.

This definitely needs some more documentation. We should have some basic Javadoc for the new classes, and we should have README updates to discuss how to use the docker-compose and build flex templates locally.

ingestion-beam/bin/build-flex-templates
relud: `gcloud dataflow flex-template build` requires either `--image=IMAGE` or `--env=[ENV,..] --flex-template-base-image=FLEX_TEMPLATE_BASE_IMAGE --image-gcr-path=IMAGE_GCR_PATH --jar=[JAR,...]`, not both. The image is built in a previous command, so this provides `--image`.

relud: the image is pre-built because the auto-built image copies dependency jars one at a time, which in turn exceeds the maximum number of docker image layers.

relud:
> It's not totally clear to me what benefit we get from the `metadata.json`

It's required to have *a* `metadata.json`, and I had assumed that only parameters specified there were allowed. I haven't tested that, though, and if we don't have to specify parameters then that sounds better, especially if we don't need to.

relud:
> > It's not totally clear to me what benefit we get from the `metadata.json`
> 
> It's required to have _a_ `metadata.json`, and I had assumed that only parameters specified there were allowed. I haven't tested that, though, and if we don't have to specify parameters then that sounds better.

we don't have to specify parameters

ingestion-beam/src/main/java/com/mozilla/telemetry/Sink.java
relud: the reflection stuff was removed because it's not required

relud:
> It's required to have _a_ `metadata.json`, and I had assumed that only parameters specified there were allowed. I haven't tested that, though, and if we don't have to specify parameters then that sounds better.

I don't know when it changed, or if I just missed it before, but apparently it is **not** required to have a `metadata.json`, and in fact if we don't want to specify parameters, then it **must not** be specified when creating the template.

docs/architecture/pain_points.md
```diff
@@ -37,10 +37,11 @@ in batch mode.
...
 Does not support [`NestedValueProvider`] for destinations in streaming mode on
-Dataflow, which is needed to create templates that accept a mapping of document
-type to a predetermined number of destinations. This is because Dataflow moves
-the implementation into the shuffler to improve performance. Current workaround
-is to specify mapping at template creation time.
+Dataflow, which is needed to create classic templates that accept a mapping of
+document type to a predetermined number of destinations. This is because
+Dataflow moves the implementation into the shuffler to improve performance.
+Current workaround is to specify mapping at classic template creation time, or
+use flex templates.
```
jklukas, line 44:
```suggestion
use Flex Templates.
```

I'm being a bit pedantic here, but GCP docs consistently use title case for Flex Templates and I do find it a bit confusing to see "flex template" which readers might think is referring to some generally known concept rather than a Dataflow-specific product feature.

docs/ingestion-beam/decoder-job.md
```diff
@@ -64,6 +64,8 @@ Decoder jobs are executed the [same way as sink jobs](../sink-job/#executing)
 but with a few extra flags:
 
 - `-Dexec.mainClass=com.mozilla.telemetry.Decoder`
+  - For Dataflow flex templates, change the `docker-compose` build argument to
```
jklukas, line 67:
```suggestion
  - For Dataflow Flex Templates, change the `docker-compose` build argument to
```

docs/ingestion-beam/republisher-job.md
```diff
@@ -36,6 +36,8 @@ Republisher jobs are executed the [same way as sink jobs](../sink-job/#executing
 but with a few differences in flags. You'll need to set the `mainClass`:
 
 - `-Dexec.mainClass=com.mozilla.telemetry.Republisher`
+  - For Dataflow flex templates, change the `docker-compose` build argument to
```
jklukas, line 39:
```suggestion
  - For Dataflow Flex Templates, change the `docker-compose` build argument to
```

docs/ingestion-beam/sink-job.md
```diff
@@ -257,14 +257,79 @@ gcloud dataflow jobs list
 gsutil cat $BUCKET/output/*
 ```
 
-### On Dataflow with templates
+### On Dataflow with flex templates
```
jklukas, line 260:
```suggestion
### On Dataflow with Flex Templates
```

docs/ingestion-beam/sink-job.md
```diff
@@ -257,14 +257,79 @@ gcloud dataflow jobs list
...
 
-Dataflow templates make a distinction between
+The Dataflow templates documentation includes [a section explaining the benefits of flex
+templates](https://cloud.google.com/dataflow/docs/concepts/dataflow-templates#evaluating-which-template-type-to-use)
```
jklukas, line 263:
```suggestion
The Dataflow templates documentation includes [a section explaining the benefits of Flex
Templates](https://cloud.google.com/dataflow/docs/concepts/dataflow-templates#evaluating-which-template-type-to-use)
```

ingestion-beam/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+COPY ingestion-beam/pom.xml /app/ingestion-beam/pom.xml
+COPY checkstyle /app/checkstyle
+COPY ingestion-beam/checkstyle /app/ingestion-sink/checkstyle
+RUN mvn prepare-package
+COPY ingestion-core/src /app/ingestion-core/src
+COPY ingestion-beam/src /app/ingestion-beam/src
+RUN mvn package -Dmaven.test.skip=true
+
+# This is the base image for the final image we're building.
+FROM gcr.io/dataflow-templates-base/java8-template-launcher-base:20210120_RC00
```
jklukas, line 14: Can we add a comment here linking to the docs that list the current recommended base images?

jklukas (approved):
Just a few nits about documentation.

PR #1557, opened 2021-02-04
Bug 1684980 - Ignore non-Glean user agents from firefox-desktop

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1684980)


acmiyaguchi:
Do you have any clues into why the tests would suddenly start failing. I don't see a connection between the two failing tests, other than the testing attributeMap possibly missing fields.

acmiyaguchi:
I think this might have been an issue with the order of `.equals`. It seems to work after rebasing and slimming down the logic.

scholtzan (approved):
lgtm

PR #1576, opened 2021-02-25
Remove decoder uri dedup option from top-level build script

whd:


relud (approved):
lgtm as long as we are setting it via DECODER_OPTS, because this option has to be set at template creation.

whd:
Yeah I'm working on the infra PR that puts this in the infra build script in the right place, this'll only be merged after that.

PR #1579, opened 2021-02-25
Create contextual services pipeline

BenWu:
This my first draft outline with some missing parts.  This runs locally, I'll try dataflow later.

TODO:
- ~~metric logging~~
- ~~verify allowlist and url query param logic~~
- ~~tests~~
- ~~make http request~~
- ~~decide what job should output~~ just errors to bigquery


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -0,0 +1,248 @@
...
+            }
+
+            if (singletonOsToUserAgentMapping == null) {
+              loadOsUserAgentMapping();
+            }
+          } catch (IOException e) {
+            throw new UncheckedIOException(e);
+          }
+
+          String reportingUrl = json.path("reporting_url").asText();
```
jklukas, line 98: Can use the constant here rather than literal

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -0,0 +1,248 @@
...
+          Map<String, String> queryParams = Arrays.stream(urlObj.getQuery().split("&"))
+              .map(param -> param.split("="))
+              .collect(Collectors.toMap(item -> item[0], item -> item[1]));
+
+          String ipParam = singletonCountryToIpMapping
+              .get(attributes.get(Attribute.NORMALIZED_COUNTRY_CODE));
+          if (ipParam == null) {
+            throw new IllegalArgumentException("Could not get ip value: Unrecognized country "
+                + Attribute.NORMALIZED_COUNTRY_CODE);
+          }
```
jklukas, line 125: Have we specified that we want to reject messages that don't have known IP or OS values? Will the adM API tolerate having these missing? We may want to check with Nan.

jklukas:
I've read through once, and the overall structure here looks great. It's consistent with the existing codebase and it's clear where the missing pieces are that still need to be filled in.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SendRequest.java
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  @Override
+  public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
+      PCollection<PubsubMessage> messages) {
+    return messages.apply(
+        MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via((PubsubMessage message) -> {
+          message = PubsubConstraints.ensureNonNull(message);
+
+          // TODO: HTTP GET
```
jklukas, line 32: I did a little review of Java HTTP clients. If we could use Java 11, I'd recommend using the new built-in HttpClient, but we're not there yet.

https://square.github.io/okhttp/ looks to be pretty widely used and the simplest interface.

The more conservative choice would be Apache HttpComponents classic, which is probably reasonably easy to use: http://hc.apache.org/httpcomponents-client-5.0.x/index.html

ingestion-beam/src/main/java/com/mozilla/telemetry/ContextualServicesReporter.java
```diff
@@ -0,0 +1,76 @@
...
+
+  /**
+   * Execute an Apache Beam pipeline and return the {@code PipelineResult}.
+   */
+  public static PipelineResult run(ContextualServicesReporterOptions.Parsed options) {
+    final Pipeline pipeline = Pipeline.create(options);
+    final List<PCollection<PubsubMessage>> errorCollections = new ArrayList<>();
+
+    // We wrap pipeline in Optional for more convenience in chaining together transforms.
+    Optional.of(pipeline) // TODO: can remove optional?
```
jklukas, line 54: Yes, we can remove `Optional`. That exists in other pipelines to allow conditionally adding transforms based on compile-time parameters. We don't have that need here.

ingestion-beam/src/main/java/com/mozilla/telemetry/ContextualServicesReporter.java
```diff
@@ -0,0 +1,76 @@
...
+            .apply(options.getInputType().read(options)) //
+            // TODO: could use subscription filter instead
+            .apply(FilterByDocType.of(options.getAllowedDocTypes())) //
+            .apply(DecompressPayload.enabled(options.getDecompressInputPayloads())) //
+            .apply(ParseReportingUrl.of(options.getUrlAllowList(), options.getCountryIpList(), //
+                options.getOsUserAgentList()))
+            .failuresTo(errorCollections) //
+            .apply(SendRequest.of()).failuresTo(errorCollections) //
+            // TODO: should we output anything at all?
+            .apply(options.getOutputType().write(options)).failuresTo(errorCollections)); //
```
jklukas, line 65: I'm of the opinion that we can rely on counters, and that we don't need to record successful output messages. We can always add that later if we see a need. I expect that output to a BQ table with `payload_bytes_decoded` schema would make sense should we need to log these.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ContextualServicesReporterOptions.java
```diff
@@ -0,0 +1,50 @@
...
+  void setCountryIpList(ValueProvider<String> value);
+
+  @Description("Path (local or gs://) to a CSV containing mappings of operating system to "
+      + "user agent used for reporting (expected format is os,ua with no header)")
+  ValueProvider<String> getOsUserAgentList();
+
+  void setOsUserAgentList(ValueProvider<String> value);
+
+  @Description("List of doc types for which to send reporting requests")
+  ValueProvider<List<String>> getAllowedDocTypes();
```
jklukas, line 31: I assume we'd want this list to be bare doctypes that aren't qualified with namespace (we assume the namespace is contextual-services), but should specify in the description.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SendRequest.java
```diff
@@ -0,0 +1,46 @@
...
+          System.out.println(message.getAttribute(Attribute.REPORTING_URL));
+
+          return message;
+        }).exceptionsInto(TypeDescriptor.of(PubsubMessage.class))
+            .exceptionsVia((ExceptionElement<PubsubMessage> ee) -> {
+              try {
+                throw ee.exception();
```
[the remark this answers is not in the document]
BenWu: I'm just starting the sending part now so it's still uncertain.  I was thinking of retrying a few times with exponential backoff and if it repeatedly fails we can write the failed messages to bigquery where we have a mechanism to manually retry from.

Do you have thoughts on the retry logic?  Like how many times to retry, time between retries, and which error codes should be retried.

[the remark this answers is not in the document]
BenWu: More specifically, I'm not sure if we should retry for 400's if we son't expect 400's in a valid request.  But 500's seem like they should be retried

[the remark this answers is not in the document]
jklukas: 5xx codes should generally indicate a transient condition. I think it makes sense to include retry logic with exponential backoff for 5xx. Anything else that's not a 2xx can flow to errors.

But I'd also be happy to keep this super-simple for the initial draft and carve out 5xx handling to a future PR.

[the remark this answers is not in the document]
ncloudioj: The retry on 5xx errors in here sounds good to me. Would it be possible to log the request errors somewhere? It'll be useful to have that if that doesn't complicate the system for this release.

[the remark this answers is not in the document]
jklukas: > The retry on 5xx errors in here sounds good to me. Would it be possible to log the request errors somewhere? It'll be useful to have that if that doesn't complicate the system for this release.

That sounds like good motivation to avoid all retry logic for the initial release. If we let the 5xx messages flow to the error output, then we'll get the full error messages along with them, which could be helpful for making decisions about specific error handling logic we want to implement.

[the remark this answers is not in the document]
ncloudioj: Having the visual on error rates (even if the retry eventually succeeds) allows us to understand the overall health of the data sharing channel, also it can alert ourselves as well as the partner if the error happens too often.

To be clear, this could be done later. Let's have something deadly simple (such as retry those 5xx errors) for now.

jklukas:
We'll need to add something in https://github.com/mozilla/gcp-ingestion/blob/master/ingestion-beam/bin/build-template to support the new job type

ingestion-beam/bin/build-template
```diff
@@ -65,6 +65,12 @@ elif [[ $JOB_TYPE =~ ^republisher_.*$ ]] ; then
 --outputFileFormat=json ${SINK_OPTS} ${REPUBLISHER_OPTS:?}"
   ERROR_OUTPUT_OPTS="--errorOutputType=file \
 --errorOutputNumShards=${ERROR_NUM_SHARDS}"
+elif [[ $JOB_TYPE =~ ^contextualservices$ ]] ; then
```
BenWu, line 68: @jklukas Can you take a look at this?  Do job specific options like `allowedDocTypes` belong here?

ingestion-beam/bin/build-template
```diff
@@ -65,6 +65,12 @@ elif [[ $JOB_TYPE =~ ^republisher_.*$ ]] ; then
...
+  JOB_CLASS="com.mozilla.telemetry.ContextualServicesReporter"
+  ERROR_OUTPUT_OPTS="--errorOutputType=bigquery \
+--errorBqWriteMethod=streaming"
+${DECODER_OPTS}"
+  ERROR_OUTPUT_OPTS="--errorOutputType=pubsub"
```
jklukas, line 73:
```suggestion
```

This line may have been a typo?

BenWu: Yes and the decoder opts oops

ingestion-beam/src/main/java/com/mozilla/telemetry/metrics/KeyedCounter.java
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Increment the counter associated with the given key by n.
+   */
+  public static void inc(String key, long n) {
+    getOrCreateCounter(key).inc(n);
+  }
+
+  @VisibleForTesting
+  static Counter getOrCreateCounter(String key) {
+    return counters.computeIfAbsent(key, k -> Metrics.counter(KeyedCounter.class, k));
```
jklukas, line 32: Can we pass in class to the KeyedCounter constructor, and then use that here, so that metrics get registered with the class they're being logically called from rather than always being registered to KeyedCounter?

jklukas: Consider this request out of scope now that I realize it's following the existing pattern from PerDoctypeCounter.

ingestion-beam/bin/build-template
jklukas: We only need to specify compile-time options here, since those are necessary at the point we're building a template.

The new options specific to ContextualServicesReporter are all runtime options (using the ValueProvider interface), so those don't need to be specified here. Instead, they'll get specified in deploy-time ops logic.

ingestion-beam/bin/build-template
```diff
@@ -65,6 +67,10 @@ elif [[ $JOB_TYPE =~ ^republisher_.*$ ]] ; then
 --outputFileFormat=json ${SINK_OPTS} ${REPUBLISHER_OPTS:?}"
   ERROR_OUTPUT_OPTS="--errorOutputType=file \
 --errorOutputNumShards=${ERROR_NUM_SHARDS}"
+elif [[ $JOB_TYPE =~ ^contextualservices$ ]] ; then
```
whd, line 70:
```suggestion
elif [[ $JOB_TYPE == "ctxsvc_reporter" ]] ; then
```

I think I'm going to refer to this as `ctxsvc_reporter` in ops infra.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -0,0 +1,302 @@
...
+        singletonCountryToIpMapping.get(DEFAULT_COUNTRY));
+    if (ipParam == null) {
+      throw new IllegalArgumentException("Could not get ip value: Unrecognized country "
+          + "and missing default value: " + Attribute.NORMALIZED_COUNTRY_CODE);
+    }
+    return ipParam;
+  }
+
+  @VisibleForTesting
+  String createUserAgentParam(String os, String clientVersion) throws UnsupportedEncodingException {
```
jklukas, line 205:
```suggestion
  String createUserAgentParam(String userAgentOs, String clientVersion) throws UnsupportedEncodingException {
```

As I was reading this, I was wondering, "why don't we use normalized_os?" because I forgot we only have user agent os. It would be nice to always refer to this non-normalized value as userAgentOs to make that more explicit.

jklukas (commented):
I hit end-of-day before I could fully read through, so leaving the few comments I have so far. I will prioritize finishing review tomorrow morning.

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/ParseReportingUrlTest.java
```diff
@@ -0,0 +1,161 @@
...
+    pipeline.run();
+
+    Map<String, String> mapping = parseReportingUrl.loadOsUserAgentMapping();
+
+    Map<String, String> expected = ImmutableMap.of("Windows", "W {0} {0}", "Macintosh", "M {0} {0}",
+        "Linux", "L {0} {0}");
+
+    Assert.assertEquals(expected, mapping);
+
+    Assert.assertEquals("W+10+10", parseReportingUrl.createUserAgentParam("Windows", "10"));
```
jklukas, line 70:
```suggestion
    // The output of createUserAgentParam should be URL-encoded, so spaces become "+".
    Assert.assertEquals("W+10+10", parseReportingUrl.createUserAgentParam("Windows", "10"));
```

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/ParseReportingUrlTest.java
```diff
@@ -0,0 +1,161 @@
...
+  @Test
+  public void testOsUserAgentLoadAndParam() throws IOException {
+    ParseReportingUrl parseReportingUrl = ParseReportingUrl.of(pipeline.newProvider(URL_ALLOW_LIST),
+        pipeline.newProvider(COUNTRY_IP_LIST), pipeline.newProvider(OS_UA_LIST));
+
+    pipeline.run();
+
+    Map<String, String> mapping = parseReportingUrl.loadOsUserAgentMapping();
+
+    Map<String, String> expected = ImmutableMap.of("Windows", "W {0} {0}", "Macintosh", "M {0} {0}",
```
jklukas, line 65:
```suggestion
    // The templates here are invented for testing and include "{0}" to validate injection of OS version.
    Map<String, String> expected = ImmutableMap.of("Windows", "W {0} {0}", "Macintosh", "M {0} {0}",
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -0,0 +1,302 @@
...
+    } else {
+      normalizedOs = DEFAULT_OS;
+    }
+    String userAgentFormatString = singletonOsToUserAgentMapping.getOrDefault(normalizedOs,
+        singletonOsToUserAgentMapping.get(DEFAULT_COUNTRY));
+    if (userAgentFormatString == null) {
+      throw new IllegalArgumentException(
+          "Could not get user agent value: Unrecognized OS and missing default value: " + os);
+    }
+    String userAgent = MessageFormat.format(userAgentFormatString, clientVersion);
```
jklukas, line 222: I have never interacted with `java.text.MessageFormat` and did not know it existed. Interesting.

Elsewhere in this codebase we use [commons-text StringSubstitutor](https://commons.apache.org/proper/commons-text/apidocs/org/apache/commons/text/StringSubstitutor.html) and I'd have a preference for using that here as well, in which case we'd replace the `{0}` placeholder with `{ua_os_version}` in the templates, which would be more descriptive and more extensible if the UA formatting becomes more complex in the future.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -0,0 +1,302 @@
...
+    return URLEncoder.encode(userAgent, "UTF-8");
```
jklukas, line 223:
```suggestion
    return URLEncoder.encode(userAgent, StandardCharsets.UTF_8.name());
```

jklukas (approved):
r+wc

Nothing blocking here. Mostly suggestions for additional commentary for pieces that took me a while to parse.

The only substantive suggestion here is changing the templating syntax for user agents.

jklukas (approved):
Looks great. 🚢

PR #1583, opened 2021-02-26
Remove support for redis deduplication

relud:
marking draft because this should not be merged until redis is actually disabled.


jklukas (approved):
Looks good. We'll also need to make sure that the deploy logic is no longer referencing these options before merging.

PR #1595, opened 2021-03-10
Change allowedDocTypes to string value

BenWu:
This changes the parameter a bit by erroring if it's missing where previously it was allow all


jklukas (approved):
Looks great!

PR #1596, opened 2021-03-10
Change allowedDocType setter type

BenWu:


whd (approved):
:+1: assuming it passes CI

PR #1600, opened 2021-03-12
Bug 1697602 Drop AET messages

jklukas:
Prerequisite for https://github.com/mozilla/gcp-ingestion/pull/1599 so we make sure that we drop any messages that might contain `ecosystem_anon_id` which was considered sensitive under the deployed AET design.

This is a first preventative step we can take right now for tearing down AET. Further steps will require some coordination with SRE.


acmiyaguchi (approved):

jklukas:
I'll plan to delay merging until Monday, just in case.

PR #1606, opened 2021-03-17
Disable okhttp redirects for ctxsvc

BenWu:


ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/SendRequestTest.java
```diff
@@ -103,4 +103,35 @@ public void testSuccessfulSend() {
...
+    WithFailures.Result<PCollection<PubsubMessage>, PubsubMessage> result = pipeline
+        .apply(Create.of(input)).apply(SendRequest.of(pipeline.newProvider(true)));
+
+    PAssert.that(result.failures()).satisfies(messages -> {
+      Iterator<PubsubMessage> iterator = messages.iterator();
+      int messageCount = 0;
+
+      for (; iterator.hasNext(); iterator.next()) {
+        messageCount++;
+      }
```
jklukas, line 127:
```suggestion
      int messageCount = Iterators.size(messages.iterator());
```

Slight preference for using a guava helper in this case.

jklukas (approved):
Looks good! One suggestion in the test.

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/SendRequestTest.java
BenWu: Wow that's useful

jklukas: Guava has a bunch of helpers like this that can make code prettier. Don't expect any magic performance-wise, though. I'm pretty sure `Iterators.size()` is doing exactly the same thing you already have written out here.

PR #1619, opened 2021-03-23
Use maven.compiler.release to build for java 8 from jdk 9+

relud:
per https://www.baeldung.com/maven-java-version#java9
> To cross-compile correctly, the new `-release` option replaces three flags: `-source`, `-target` and `-bootclasspath`.

Note that `maven.compiler.release` is only valid for jdk 9+, and when set causes maven-compiler-plugin to ignore `source` and `target`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,7 @@ There are currently four components:
 For more information, see [the documentation](https://mozilla.github.io/gcp-ingestion).
 
 Java 11 support is a work in progress for the Beam Java SDK, so this project requires
-Java 8 and will likely fail to compile using newer versions of the JDK.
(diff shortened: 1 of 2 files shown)


relud:
test failures are unrelated, and should be fixed by #1620

jklukas (approved):
LGTM

Am I understanding correctly that we'll still be building our production templates with JDK 8, and this change is at this point simply a convenience for development so that you can run tests, etc. locally without having to make sure you're on JDK 8?

relud:
> Am I understanding correctly that we'll still be building our production templates with JDK 8, and this change is at this point simply a convenience for development so that you can run tests, etc. locally without having to make sure you're on JDK 8?

exactly

PR #1633, opened 2021-04-05
Resume unacked tasks when flushing a queue for flush_manager

relud:


relud:
this isn't necessary, as persistqueue already does this by default

PR #1634, opened 2021-04-05
Work around bug that was breaking q.empty()

relud:
https://github.com/peter-wangxu/persist-queue/pull/154 can cause `q.empty()` to return inaccurate results, which in turn can cause the flush manager to incorrectly think that disks were flushed because this loop wouldn't drain the queue:
https://github.com/mozilla/gcp-ingestion/blob/70b89609350f941d4e5b606e4b8558fe9550fc89/ingestion-edge/ingestion_edge/flush.py#L134

### DIFF
--- a/ingestion-edge/ingestion_edge/publish.py
+++ b/ingestion-edge/ingestion_edge/publish.py
@@ -106,7 +106,11 @@ def get_queue(config: dict) -> SQLiteAckQueue:
         for key, value in config.items()
         if key.startswith("QUEUE_")
     }
-    return SQLiteAckQueue(**queue_config)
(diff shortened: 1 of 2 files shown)


whd (commented):
This does appear to fix the error case we're debugging, but tests are failing and the upstream fix PR is also failing CI.

whd (approved):
I'm slightly worried that we're catching this now instead of in a test case somewhere, but LGTM

relud:
> I'm slightly worried that we're catching this now instead of in a test case somewhere, but LGTM

I filed #1635 to track adding a relevant test

PR #1638, opened 2021-04-07
Fix warning about missing plugin versions

relud:
and consistently omit `<groupId>org.apache.maven.plugins</groupId>`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -163,7 +163,6 @@
     <build>
         <plugins>
             <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


jklukas (approved):
I have no idea what `<groupId>org.apache.maven.plugins</groupId>` does. What was it doing? Why are we removing it?

relud:
> I have no idea what `<groupId>org.apache.maven.plugins</groupId>` does. What was it doing? Why are we removing it?

it works exactly like `groupId` for dependencies, it's a namespace qualifier. in the case of plugins, it's only required if they are third party.

PR #1639, opened 2021-04-07
Bug 1703648 - Assign rally or pioneer id in pioneer decoder

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1703648)


akkomar:
According to [Slack thread](https://mozilla.slack.com/archives/C010NL72MT9/p1617882360142100?thread_ts=1617820542.138600&cid=C010NL72MT9) they're fine with renaming the study so I think we don't need this.

whd:
Unfortunately I already provisioned the infrastructure for this so I think we should still move ahead with this change.

akkomar (approved):

PR #1672, opened 2021-04-29
Bug 1708552 - Do not inject metadata into rally pings

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1708552)

Injecting the rally id into the top level causes the document to fail schema validation. This removes variable injection, which means the rally_id and study_name are now deprecated fields.


akkomar (approved):
LGTM

PR #1686, opened 2021-05-07
Update beam to 2.29 and jackson to 2.12.1

relud:
use the new `beam-sdks-java-google-cloud-platform-bom` to manage beam dependency versions
and update tests because `PubsubMessage` now enforces non-null payload

this does not remove the dependency on `com.google.cloud:pubsublite-beam-io:0.14.1`, because `org.apache.beam:beam-sdks-java-io-google-cloud-platform:2.29.0` still depends on `com.google.api.grpc:grpc-google-cloud-pubsublite-v1:0.7.0`, which isn't new enough to allow messages over 1MB.


ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/PubsubConstraints.java
```diff
@@ -38,14 +38,12 @@ public static String truncateAttributeValue(String value) {
     return truncateToFitUtf8ByteLength(value, MAX_ATTRIBUTE_VALUE_BYTES);
   }
 
-  /** Fills out empty payload and attributes if the message itself or components are null. */
+  /** Fills out empty attributes if message or attributeMap are null. */
   public static PubsubMessage ensureNonNull(PubsubMessage message) {
     if (message == null) {
-      return new PubsubMessage(new byte[] {}, new HashMap<>());
-    } else if (message.getPayload() == null) {
-      return ensureNonNull(new PubsubMessage(new byte[] {}, message.getAttributeMap()));
```
relud, line 46: this condition was removed because it's now unreachable

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/PubsubLiteCompatTest.java
```diff
@@ -12,7 +12,6 @@
 import com.mozilla.telemetry.ingestion.core.util.IOFunction;
 import com.mozilla.telemetry.options.OutputFileFormat;
 import com.mozilla.telemetry.util.Json;
-import com.mozilla.telemetry.util.PubsubWithNullsJsonCoder;
```
relud, line 15: special coder is no longer needed, because it was only needed to code messages with null payload

PR #1689, opened 2021-05-12
 Aggregate topsites impressions by timed window

BenWu:
Logic is mostly in `AggregateImpressions`.  Running this definitely aggregates something and outputs urls but I haven't checked yet if it's correct.  What I'm still trying to wrap my head around is the windowing timestamps and lateness.  I remember seeing a warning about late data being dropped while testing yesterday but I can't find it now.

I'll keep testing things out and read docs more.

It might be useful at this point to output something to bigquery with urls for debugging since the urls aren't 1 to 1 with the pings anymore.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
@@ -1,0 +1,99 @@
+package com.mozilla.telemetry.contextualservices;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.Iterables;
+import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
+import com.mozilla.telemetry.transforms.PubsubConstraints;
+import com.mozilla.telemetry.util.Time;
+import java.util.List;
+import java.util.Map;
+import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
+import org.apache.beam.sdk.options.ValueProvider;
+import org.apache.beam.sdk.transforms.DoFn;
+import org.apache.beam.sdk.transforms.GroupByKey;
+import org.apache.beam.sdk.transforms.PTransform;
+import org.apache.beam.sdk.transforms.ParDo;
+import org.apache.beam.sdk.transforms.windowing.FixedWindows;
+import org.apache.beam.sdk.transforms.windowing.IntervalWindow;
+import org.apache.beam.sdk.transforms.windowing.Window;
+import org.apache.beam.sdk.values.KV;
+import org.apache.beam.sdk.values.PCollection;
+import org.joda.time.Duration;
+
+/**
+ * Count the number of impressions aggregated by reporting URL.
+ */
+public class AggregateImpressions
+    extends PTransform<PCollection<PubsubMessage>, PCollection<PubsubMessage>> {
+
+  private final List<String> aggregationFields = ImmutableList.of(
+      ReportingUrlParser.PARAM_COUNTRY_CODE, ReportingUrlParser.PARAM_REGION_CODE,
+      ReportingUrlParser.PARAM_FORM_FACTOR, ReportingUrlParser.PARAM_OS_FAMILY,
+      ReportingUrlParser.PARAM_ID);
+
+  private final ValueProvider<Integer> aggregationWindowDuration;
+
+  public AggregateImpressions(ValueProvider<Integer> aggregationWindowDuration) {
+    this.aggregationWindowDuration = aggregationWindowDuration;
+  }
+
+  public static AggregateImpressions of(ValueProvider<Integer> aggregationWindowDuration) {
+    return new AggregateImpressions(aggregationWindowDuration);
+  }
+
+  @Override
+  public PCollection<PubsubMessage> expand(PCollection<PubsubMessage> messages) {
+    return messages
+        .apply(
+            Window.into(FixedWindows.of(Duration.standardMinutes(aggregationWindowDuration.get()))))
+        .apply(ParDo.of(new AddAggregationKey())).apply(GroupByKey.create())
+        .apply(ParDo.of(new CountImpressionsPerKey()));
+  }
+
+  private class AddAggregationKey extends DoFn<PubsubMessage, KV<String, PubsubMessage>> {
+
+    @ProcessElement
+    public void processElement(@Element PubsubMessage message,
+        OutputReceiver<KV<String, PubsubMessage>> out) {
+      message = PubsubConstraints.ensureNonNull(message);
+
+      String reportingUrl = message.getAttribute(Attribute.REPORTING_URL);
+
+      ReportingUrlParser urlParser = new ReportingUrlParser(reportingUrl);
+
+      // Rebuild url filtering out unneeded parameters
+      ReportingUrlParser aggregationUrl = new ReportingUrlParser(urlParser.getBaseUrl() + "?");
+      for (String name : aggregationFields) {
+        aggregationUrl.addQueryParam(name, urlParser.getQueryParam(name));
+      }
+
+      out.output(KV.of(aggregationUrl.getReportingUrl().toString(), message));
+    }
+  }
+
+  private class CountImpressionsPerKey
+      extends DoFn<KV<String, Iterable<PubsubMessage>>, PubsubMessage> {
+
+    @ProcessElement
+    public void processElement(@Element KV<String, Iterable<PubsubMessage>> input,
+        OutputReceiver<PubsubMessage> out, IntervalWindow window) {
+      ReportingUrlParser urlParser = new ReportingUrlParser(input.getKey());
+
+      int impressionCount = Iterables.size(input.getValue());
+      long windowStart = window.start().getMillis();
+      long windowEnd = window.end().getMillis();
+
+      urlParser.addQueryParam(ReportingUrlParser.PARAM_IMPRESSIONS,
+          Integer.toString(impressionCount));
+      urlParser.addQueryParam(ReportingUrlParser.PARAM_TIMESTAMP_BEGIN, Long.toString(windowStart));
+      urlParser.addQueryParam(ReportingUrlParser.PARAM_TIMESTAMP_END, Long.toString(windowEnd));
+
+      Map<String, String> attributeMap = ImmutableMap.of(Attribute.REPORTING_URL,
+          urlParser.toString(), Attribute.SUBMISSION_TIMESTAMP,
+          Time.epochNanosToTimestamp(windowEnd / 1000));
+
+      out.output(new PubsubMessage(new byte[] {}, attributeMap));
+    }
+  }
+}
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
@@ -1,10 +1,12 @@
 package com.mozilla.telemetry.contextualservices;
 
+import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Iterables;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
 import com.mozilla.telemetry.transforms.PubsubConstraints;
+import com.mozilla.telemetry.util.Json;
 import com.mozilla.telemetry.util.Time;
 import java.util.List;
 import java.util.Map;
@@ -15,6 +17,7 @@ import org.apache.beam.sdk.transforms.GroupByKey;
 import org.apache.beam.sdk.transforms.PTransform;
 import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.transforms.windowing.FixedWindows;
+import org.apache.beam.sdk.transforms.windowing.GlobalWindows;
 import org.apache.beam.sdk.transforms.windowing.IntervalWindow;
 import org.apache.beam.sdk.transforms.windowing.Window;
 import org.apache.beam.sdk.values.KV;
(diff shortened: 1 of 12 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ContextualServicesReporterOptions.java
```diff
@@ -30,6 +30,12 @@
 
   void setReportingEnabled(ValueProvider<Boolean> value);
 
+  @Description("Size (in minutes) of the window for aggregation of reporting requests.")
```
jklukas, line 33: I'd like to use a Duration here rather than assuming minutes. See this pattern in https://github.com/mozilla/gcp-ingestion/blob/e023a6a4c740633ce64ca726a59ed95e034ddb96/ingestion-beam/src/main/java/com/mozilla/telemetry/options/SinkOptions.java#L226-L232

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,109 @@
...
+    this.getAggregationWindowSize = getAggregationWindowSize;
+  }
+
+  public static AggregateImpressions of(ValueProvider<Integer> getAggregationWindowSize) {
+    return new AggregateImpressions(getAggregationWindowSize);
+  }
+
+  @Override
+  public PCollection<PubsubMessage> expand(PCollection<PubsubMessage> messages) {
+    return messages.apply(ParDo.of(new AddAggregationKey())) //
```
jklukas, line 50: I need to do some more reading through Beam docs, but this looks like a reasonable track. We should probably drop the message at this point and just have a KV with an empty value, so that we don't have to shuffle the value. There's a `Count.perKey()` transform we may want to use here.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,109 @@
...
+          Long.toString(windowStart / 1000));
+      urlParser.addQueryParam(ReportingUrlUtil.PARAM_TIMESTAMP_END,
+          Long.toString(windowEnd / 1000));
+
+      Map<String, String> attributeMap = ImmutableMap.of(Attribute.REPORTING_URL,
+          urlParser.toString());
+
+      ObjectNode json = Json.createObjectNode(); // TODO: for debug output
+      json.put(Attribute.REPORTING_URL, urlParser.toString());
+      json.put(Attribute.SUBMISSION_TIMESTAMP, Time.epochNanosToTimestamp(windowEnd / 1000));
```
jklukas, line 103: I need to think more about this too, but I'm not sure submission_timestamp gives us any real value here, and it may be better to let this job use processing time for windowing. That way, we'll never have to worry about late records, etc.

BenWu: The submission timestamp here is just for the bigquery output for debugging and the timestamp is just to have something easy to look at when looking at the table.  I think the windowing is using processing time but I was confused because I saw that late data was being dropped.

This is what shows up in the counters:
Screen Shot 2021-05-12 at 5 28 42 PM
This is from yesterday and I haven't been able to recreate it today.

jklukas:
Did some more reading through Beam docs this morning. If we decide we need to use submission_timestamp, we'd need to make sure our windowing is using that value. See https://beam.apache.org/documentation/programming-guide/#adding-timestamps-to-a-pcollections-elements for how to apply custom timestamps to elements.

But I think at least for a first pass, we should ignore submission_timestamp and instead use the default timestamps on the records, applied by Beam when reading records from PubSub, which will basically make the concern of late-arriving data go away.

Here's a basic structure of how counting would work based on higher-level transforms in the Beam SDK:

```
    pipeline.apply(Create.of("a", "b", "a"))
        .apply(Window.into(FixedWindows.of(Duration.standardMinutes(10))))
        .apply(WithKeys.of(elem -> elem)).setCoder(KvCoder.of(StringUtf8Coder.of(), StringUtf8Coder.of()))
        .apply(Count.perKey()));
```

For the `WithKeys`, we can assign a key based on URI similar to what's already in this PR, but not include timestamps in the URIs at that point.

After applying the count, we can have a final `DoFn` that adds the begin and end timestamps based on the window context:

```
    @ProcessElement
    public void processElement(@Element String uri,
        OutputReceiver<KV<String, PubsubMessage>> out, BoundedWindow window) {
      Instant windowEnd = window.maxTimestamp();
      ...
```

There are a bunch of magic parameters you can add in to a `processElement` declaration to get various context about the execution and it's difficult to discover those through the Beam Javadoc. BoundedWindow appears to be useful here based on https://stackoverflow.com/questions/40366910/how-to-get-the-max-timestamp-of-the-current-sliding-window

BenWu:
Updated to look a little cleaner.  As far as I can tell this is working.  I output both unaggregated and aggregated urls to a table and the total counts are close (unaggregated is slightly higher but I think that's due to cutoff when I stop the job).  It's hard to exactly test because of submission timestamps vs. window timestamps but it looks reasonably close.

e.g. sample queries
aggregated https://console.cloud.google.com/bigquery?sq=638854937685:4cde5c2e06df476aa4debc7998b11907
unaggregated https://console.cloud.google.com/bigquery?sq=638854937685:ab5738899cc746e6863a513de307688b

I'm seeing some droppedDueToLateness but only after the job is drained or cancelled.  Possibly because of leftover elements?  I'm not entirely sure why there would be late data but in my limited testing it might not be a problem in practice.

I'm also seeing `RuntimeException: ManagedChannel allocation site` for the write to bigquery step with this message in the worker logs
```
*~*~*~ Channel ManagedChannelImpl{logId=11745, target=bigquerystorage.googleapis.com:443} was not shutdown properly!!! ~*~*~* Make sure to call shutdown()/shutdownNow() and wait until awaitTermination() returns true.
```
which I haven't figured out yet.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,109 @@
...
+        .apply(WithKeys.of((SerializableFunction<PubsubMessage, String>) this::getAggregationKey)) //
+        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of())) //
+        .apply(WithTimestamps.of(message -> new Instant())) //
+        .apply(Window.into(FixedWindows.of(Time.parseDuration(getAggregationWindowDuration.get())))) //
+        .apply(Count.perKey()) //
+        .apply(ParDo.of(new BuildAggregateUrl())) //
+        .apply(Window.into(new GlobalWindows()));
+  }
+
+  private String getAggregationKey(PubsubMessage message) {
```
jklukas, line 65: Can this method be static?

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,109 @@
...
+    return new AggregateImpressions(getAggregationWindowSize);
+  }
+
+  @Override
+  public PCollection<PubsubMessage> expand(PCollection<PubsubMessage> messages) {
+    return messages //
+        .apply(WithKeys.of((SerializableFunction<PubsubMessage, String>) this::getAggregationKey)) //
+        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of())) //
+        .apply(WithTimestamps.of(message -> new Instant())) //
+        .apply(Window.into(FixedWindows.of(Time.parseDuration(getAggregationWindowDuration.get())))) //
```
jklukas, line 59: I'm realizing that we may not want to set a window at all, and instead use GlobalWindows with a processing time trigger: https://beam.apache.org/documentation/programming-guide/#processing-time-triggers

It would be something like:

```
        .apply(Window.configure().triggering(
          AfterProcessingTime.pastFirstElementInPane().alignedTo(
            Duration.standardMinutes(10))))
```

I think that eliminates the possibility of dropped messages, as this would still drain remaining messages when the pipeline is drained.

Problems with this: it appears that `AfterProcessingTime` doesn't accept ValueProvider. Also, I haven't fully been able to figure out syntax here to get global windows with this trigger.

It would be interesting to at least try out this approach and see if we see the dropped message problem go away.

BenWu: After some reading and tinkering this is what I came up with before my EOD:
```java
return messages //
    .apply(WithKeys.of((SerializableFunction<PubsubMessage, String>) this::getAggregationKey)) //
    .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of())) //
    .apply(Window.<KV<String, PubsubMessage>>configure()
        .triggering(Repeatedly.forever(AfterProcessingTime.pastFirstElementInPane()
            .alignedTo(Time.parseDuration(aggregationWindowDuration.get()))))
        .discardingFiredPanes()) //
    .apply(Count.perKey()) //
    .apply(ParDo.of(new BuildAggregateUrl())) //
    .apply(Window.<PubsubMessage>into(new GlobalWindows()).triggering(DefaultTrigger.of()));
```

Setting default trigger after is needed because of `An exception occured while executing the Java class. Inputs to Flatten had incompatible triggers: Repeatedly.forever(AfterSynchronizedProcessingTime.pastFirstElementInPane()), DefaultTrigger` from the write output stage.  I'm not sure if this is right thing to do.

This uses a global window instead of an interval window which means we can't get a nice start and end of the aggregation period without some extra steps.

This seems to do the right thing.  Urls with aggregated counts are being output but I haven't validated it.  Also I don't see any droppedDueToLateness after stopping the job.

jklukas:
> This explains the dropped due to lateness: https://cloud.google.com/dataflow/docs/guides/stopping-a-pipeline#effects_of_draining_a_job
> 
> > When you issue the Drain command, Dataflow immediately closes any in-process windows and fires all triggers. The system does not wait for any outstanding time-based windows to finish.
> 
> So the trigger doesn't fire again after draining and everything that comes is "late"

It follows this with a note about forcing windows closed:

> Dataflow causes open windows to close by advancing the system watermark to infinity

I'm guessing this doesn't work for `GlobalWindows`. So perhaps a workaround would be to use `FixedWindows` of arbitrarily large but finite size (something like `FixedWindows.of(100 years)` so that it covers 1970 to 2070). This would effectively be a global window, but hopefully the advancing of the watermark to infinity would cause it to flush. We'd still have the 10-minute trigger otherwise.

BenWu:
To update, I don't see any `droppedDueToLateness` after a couple tests with the long fixed window.  I'll commit the changes for that.  It does feel like a hack but it seems to work.

I also did a rough comparison of the number of pings vs the number counted in the each aggregation period and it looks close.  The timestamps don't line up so this is just an estimate.  query here: https://console.cloud.google.com/bigquery?sq=638854937685:86b830f8fbd64e1c96bb5129401ba213
The intervals are very close other than a couple where there were no individual pings written.  My guess is that this is due to the `RuntimeException: ManagedChannel allocation site` error.  I'll change to write to the error stream and see what happens.  Other than that, I'm reasonably confident this is doing the right thing.

jklukas:
>  My guess is that this is due to the RuntimeException: ManagedChannel allocation site error

I've looked a bit more into this, and it looks like this bubbles up to `BigQueryWriteClient`, which I'm fairly certain is the cient for the new grpc-based BigQuery Storage Write API. Previously, when we did streaming in Dataflow jobs, BigQueryIO was using the old streaming inserts API instead.

This gives me confidence that we will not be seeing `ManagedChannel` errors if we move to sending these messages to PubSub via error output.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,119 @@
...
+        // .apply(WithTimestamps.of(message -> new Instant())) //
+        .apply(Window.<KV<String, PubsubMessage>>into(FixedWindows.of(Duration.standardDays(36500)))
+            .triggering(Repeatedly.forever(AfterProcessingTime.pastFirstElementInPane()
+                .alignedTo(Time.parseDuration(aggregationWindowDuration.get()))))
+            .withAllowedLateness(Duration.ZERO).discardingFiredPanes()) //
+        // .apply(Window.<KV<String,
+        // PubsubMessage>>into(FixedWindows.of(Time.parseDuration(aggregationWindowDuration.get()))))
+        .apply(Count.perKey()) //
+        .apply(ParDo.of(new BuildAggregateUrl())) //
+        .apply(Window.<PubsubMessage>into(new GlobalWindows()).triggering(DefaultTrigger.of()));
```
BenWu, line 71: While testing writing reporting urls to errors I got more dropped due to lateness after draining.  I'm not sure why it happened now but now the couple of times I tried yesterday.  

If the watermark advances to infinity when draining starts then that would be past the end of any window so any data entering a window would be late since it's older than the watermark.  So I think this makes sense.  I'm still confused about this and it doesn't seem like setting a large window fixed it consistently.

jklukas: I've taken this question to the beam user mailing list: https://lists.apache.org/thread.html/raca525048383ad7d4c428d6924e21708c05fea2147be7aa39f37c4e0%40%3Cuser.beam.apache.org%3E

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,113 @@
...
+    return messages //
+        // Add reporting url as key, pubsub message as value
+        .apply(WithKeys.of((SerializableFunction<PubsubMessage, String>) //
+        AggregateImpressions::getAggregationKey)) //
+        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of())) //
+        // Set timestamp to current time
+        .apply(WithTimestamps.of(message -> new Instant())) //
+        // Group impressions into timed windows
+        // TODO: Use valueprovider
+        .apply("IntervalWindow", Window.into(FixedWindows.of(Time.parseDuration("10m")))) //
```
BenWu, line 64: For this line if I do `Time.parseDuration(aggregationWindowDuration.get())`, I get this error 
```
Value only available at runtime, but accessed from a non-runtime context: RuntimeValueProvider{propertyName=providerRuntimeValues, default=null}
```
 in the unit test that has `pipeline.apply(AggregateImpressions.of(pipeline.newProvider("10m")))` but not in the real pipeline.  I understand the error but what would be a good workaround here?

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -0,0 +1,113 @@
...
+      long windowStart = window.start().getMillis();
+      long windowEnd = window.end().getMillis();
+
+      urlParser.addQueryParam(ParsedReportingUrl.PARAM_IMPRESSIONS, Long.toString(impressionCount));
+      urlParser.addQueryParam(ParsedReportingUrl.PARAM_TIMESTAMP_BEGIN,
+          Long.toString(windowStart / 1000));
+      urlParser.addQueryParam(ParsedReportingUrl.PARAM_TIMESTAMP_END,
+          Long.toString(windowEnd / 1000));
+
+      Map<String, String> attributeMap = ImmutableMap.of(Attribute.REPORTING_URL,
```
BenWu, line 106: submission_timestamp is added for the output written to error output.  Payload and attributes are empty otherwise

BenWu (commented):
Wrote some unit tests and switched back to fixed windowing.  Only issue is the runtime value for testing which I don't have time to fix until tomorrow.  I left an inline comment.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
jklukas: It looks like FixedWindows doesn't give us any hook to use a ValueProvider, so it expects to know window configuration at compile time. I don't see any way around that, which means `aggregationWindowDuration` needs to become a compile-time parameter.

This whole problem goes away once we're able to move to deploys based on Flex Templates.

jklukas (approved):
r+wc

Looks good to go once the aggregation window duration is changed to compile-time.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ContextualServicesReporterOptions.java
```diff
@@ -30,6 +30,19 @@
 
   void setReportingEnabled(ValueProvider<Boolean> value);
 
+  @Description("Duration window for aggregation of reporting requests.")
+  @Default.String("10m")
+  String getAggregationWindowDuration();
+
+  void setAggregationWindowDuration(String value);
```
BenWu, line 37: @whd this is being added as a compile-time parameter.  Is there anything else that needs to be done?  We can use the default value for now

whd: If we need to set this in ops logic we can use https://github.com/mozilla-services/cloudops-infra/blob/master/projects/beam/Jenkinsfile.build#L136, but if we're using the default value I don't think anything else needs to change.

If we need to change this at some point without a corresponding code change it will require replacing the existing template since we don't currently have versioning on compile time parameters. That or we can just update the code default every time we want to change this.

PR #1699, opened 2021-05-18
Bug 1711706 Support X-Telemetry-Agent header

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1711706

This will require changing config for ingestion-edge and redeploying to allow `X-Telemetry-Agent` to be passed to pubsub.

The relevant fields are added to schemas in https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/676

I don't _think_ the order in which these are merged/applied particularly matters. We should be tolerant at each step of the new field existing or not. If we start processing the field before schemas have a relevant field, the value for the new field will simply get dropped until the schemas are updated.

### DIFF
--- a/ingestion-edge/ingestion_edge/config.py
+++ b/ingestion-edge/ingestion_edge/config.py
@@ -62,6 +62,7 @@ DEFAULT_METADATA_HEADERS = [
     "X-Pipeline-Proxy",
     "X-Debug-ID",
     "X-Source-Tags",
+    "X-Telemetry-Agent",
 ]
 
 METADATA_HEADERS = {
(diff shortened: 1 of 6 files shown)


relud (approved):

ingestion-edge/ingestion_edge/config.py
```diff
@@ -62,6 +62,7 @@ class Route:
     "X-Pipeline-Proxy",
     "X-Debug-ID",
     "X-Source-Tags",
+    "X-Telemetry-Agent",
```
whd, line 65: We don't set this in ops config so a redeploy to pick up the code change should suffice.

whd (approved):

PR #1711, opened 2021-06-08
Revert "Update beam to 2.29 and jackson to 2.12.1"

jklukas:
Reverts mozilla/gcp-ingestion#1686

We need to downgrade to beam 2.28 because we are being hit by https://issues.apache.org/jira/browse/BEAM-12356 in the Contextual Services sender job. It sounds like the relevant fix will be included in 2.31, which won't be available for ~5 weeks, at which point we can restore these changes.


BenWu (approved):
I'll approve because I know this will work for my use case but I'm not certain about the effects on other pipelines

relud (approved):

jklukas:
I've staged https://github.com/mozilla/gcp-ingestion/pull/1712 as a draft for restoring these changes when 2.31 is available.

PR #1712, opened 2021-06-08
Update beam to 2.31 and jackson to 2.12.1

jklukas:
Reverts mozilla/gcp-ingestion#1711 to restore the changes from mozilla/gcp-ingestion#1686

Beam 2.31 should be released mid-July and will contain fixes to BigQueryIO that will allow streaming inserts to run smoothly. Until then, we need to stick with beam 2.28.


pom.xml
```diff
@@ -62,13 +62,12 @@
         <!-- The classPath setting is a workaround for https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=911925 -->
         <argLine>-Xmx1024m -XX:MaxPermSize=256m -Djdk.net.URLClassPath.disableClassPathURLCheck=true</argLine>
 
-        <beam.version>2.28.0</beam.version>
+        <beam.version>2.29.0</beam.version>
```
jklukas, line 65:
```suggestion
        <beam.version>2.31.0</beam.version>
```

Once 2.31 is released.

jklukas:
Beam 2.31.0 was released today, so we should be able to proceed with finalizing this.

whd (approved):

PR #1739, opened 2021-06-24
Add WithCurrentTimestamp to avoid allowedSkew errors

jklukas:
~~Needs tests.~~


jklukas:
cc @whd

jklukas:
If we're able to get this reviewed today, I'll plan to not merge until we're back from wellness week. Just in case.

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/WithCurrentTimestampTest.java
```diff
@@ -0,0 +1,21 @@
...
+import org.joda.time.Instant;
+import org.junit.Test;
+
+public class WithCurrentTimestampTest {
+
+  @Test
+  public void testAdjustedTimestamp() {
+    Instant startTimestamp = new Instant();
+    assertEquals(1.0 * startTimestamp.getMillis(),
+        1.0 * WithCurrentTimestamp.adjustedTimestamp(new Instant(0)).getMillis(), 100.0);
```
BenWu, line 14: Is the 100 for the delta?  Shouldn't we expect no delta since the instant should be the same?

BenWu (approved):
This looks fine to me

ingestion-beam/src/test/java/com/mozilla/telemetry/transforms/WithCurrentTimestampTest.java
jklukas: I should have added a comment here. I would expect the transform to create its own `new Instant()` which will probably be a few millis after `startTimestamp`.

PR #1740, opened 2021-06-25
Bug 1716665 - Add metadata checking filters for ctxsvc

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1716665


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -218,23 +230,33 @@ private String getOsParam(String userAgentOs) {
   @VisibleForTesting
   List<Set<String>> loadAllowedUrls() throws IOException {
     if (singletonAllowedImpressionUrls == null || singletonAllowedClickUrls == null) {
-      singletonAllowedImpressionUrls = new HashSet<>();
-      singletonAllowedClickUrls = new HashSet<>();
+      Set<String> allowedImpressionUrls = new HashSet<>();
+      Set<String> allowedClickUrls = new HashSet<>();
```
BenWu, line 234: Having the assignment here seemed to be causing some concurrency issues in the new test where some inputs would return from this method with the empty allow list set and the message would be rejected even if the url was valid.  Looking at the `InvalidUrlException`'s in the prod errors table this look like it happened to 6 messages in the last 14 days.  This change still isn't thread-safe but it seems to work.  I'm not sure if any synchronization is needed or how to reason about that in beam

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -218,23 +230,33 @@ private String getOsParam(String userAgentOs) {
...
         }
       }
+
+      singletonAllowedImpressionUrls = allowedImpressionUrls;
+      singletonAllowedClickUrls = allowedClickUrls;
     }
     return Arrays.asList(singletonAllowedClickUrls, singletonAllowedImpressionUrls);
   }
+
+  private void checkIfParamPresent(ParsedReportingUrl reportingUrl, String paramName) {
```
jklukas, line 256: The name here should make it more apparent that an exception will be thrown. Probably `requireParamPresent`.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
jklukas: I think this change is sufficient. It ensures that the singletons will only ever be set to null or to a complete set. If a thread sees the singleton as null, it will be guaranteed to go through the loading process again. If a thread sees a non-null value, it's guaranteed to be a complete set.

In the previous version of the code, it makes sense that a thread could end up reading a non-null but incomplete value. In GeoCityLookup, we do have `getOrCreate*` methods that are marked as synchronized which ensures we don't create those large maps multiple times in memory.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/VerifyMetadata.java
```diff
@@ -0,0 +1,87 @@
...
+
+          if (!docTypesToVerify.isAccessible()
+              || !docTypesToVerify.get().contains(attributes.get(Attribute.DOCUMENT_TYPE))) {
+            return message;
+          }
+
+          // Message must be gzip compressed
+          if (!GzipUtil.isGzip(message.getPayload())) {
+            throw new RejectedMessageException("Payload must be gzip compressed", "gzip");
+          }
```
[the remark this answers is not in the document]
jklukas: See https://github.com/mozilla/gcp-ingestion/pull/1751

jklukas (approved):
Changes look great. Looks like CI failure is just checkstyle at this point.

BenWu:
I removed the option to choose which doctypes should go through the VerifyMetadata transform because I don't see a reason to not verify everything.  This also means we don't need to update the cloudops deploy config which is nice.  @jklukas do you have any objections to this?

jklukas:
> I removed the option to choose which doctypes should go through the VerifyMetadata transform because I don't see a reason to not verify everything. This also means we don't need to update the cloudops deploy config which is nice. @jklukas do you have any objections to this?

No objections. This PR includes sufficient metric coverage to help us understand if something is going wild with the filtering.

PR #1745, opened 2021-07-06
Bug 1719353 Support X-Foxsec-IP-Reputation header

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1719353

### DIFF
--- a/docs/architecture/decoder_service_specification.md
+++ b/docs/architecture/decoder_service_specification.md
@@ -82,7 +82,7 @@ required group attributes {
   optional string x_pingsender_version       // header from client
   optional string x_debug_id                 // header from client
   optional string x_source_tags              // header from client
-  optional string x_foxsec_ip_reputation     // header from client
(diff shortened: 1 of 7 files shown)


docs/architecture/decoder_service_specification.md
```diff
@@ -82,6 +82,7 @@ required group attributes {
   optional string x_pingsender_version       // header from client
   optional string x_debug_id                 // header from client
   optional string x_source_tags              // header from client
+  optional string x_foxsec_ip_reputation     // header from client
```
whd, line 85:
```suggestion
  optional string x_foxsec_ip_reputation     // header from iprepd
```

whd (approved):
Look like last time we added a header we changed a different architecture doc: https://github.com/mozilla/gcp-ingestion/pull/1699/files. I'm not sure if those are supposed to be kept in sync. We should probably also update  `docs/ingestion-edge/index.md` with the new default value.

relud (approved):

PR #1748, opened 2021-07-08
Bug 1719724 Add labeling of click spikes for contextual services

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1719724

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
@@ -41,8 +41,8 @@ public class LabelClickSpikes extends
 
   private final Integer maxClicks;
   private final Long windowMillis;
-  private final Counter status64Counter = Metrics.counter(LabelClickSpikes.class,
-      "click_status_64");
+  private final Counter status65Counter = Metrics.counter(LabelClickSpikes.class,
+      "click_status_65");
 
   /**
    * Composite transform that wraps {@code DetectClickSpikes} with keying by {@code context_id}.
@@ -87,8 +87,8 @@ public class LabelClickSpikes extends
   private static void addClickStatusToReportingUrlAttribute(Map<String, String> attributes) {
     String reportingUrl = attributes.get(Attribute.REPORTING_URL);
     ParsedReportingUrl urlParser = new ParsedReportingUrl(reportingUrl);
-    // "64" indicates a spike in clicks per the Conducive API spec.
-    urlParser.addQueryParam("click-status", "64");
+    // "65" indicates a short-term spike in "ghost" clicks per the Conducive API spec.
+    urlParser.addQueryParam("click-status", "65");
     attributes.put(Attribute.REPORTING_URL, urlParser.toString());
   }
 
@@ -117,7 +117,7 @@ public class LabelClickSpikes extends
         PubsubMessage message = element.getValue();
         Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
         addClickStatusToReportingUrlAttribute(attributes);
-        status64Counter.inc();
+        status65Counter.inc();
         out.output(KV.of(element.getKey(), new PubsubMessage(message.getPayload(), attributes)));
       }
     }
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
@@ -36,12 +36,12 @@ import org.joda.time.Instant;
 /**
  * Transform that maintains state per key in order to label suspicious clicks.
  */
-public class DetectClickSpikes extends
+public class LabelClickSpikes extends
     PTransform<PCollection<KV<String, PubsubMessage>>, PCollection<KV<String, PubsubMessage>>> {
 
   private final Integer maxClicks;
   private final Long windowMillis;
-  private final Counter status64Counter = Metrics.counter(DetectClickSpikes.class,
+  private final Counter status64Counter = Metrics.counter(LabelClickSpikes.class,
       "click_status_64");
 
   /**
@@ -53,15 +53,15 @@ public class DetectClickSpikes extends
         .apply(WithKeys.of((message) -> message.getAttribute(Attribute.CONTEXT_ID))) //
         .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of()))
         .apply(WithCurrentTimestamp.of()) //
-        .apply(DetectClickSpikes.of(maxClicks, windowDuration)) //
+        .apply(LabelClickSpikes.of(maxClicks, windowDuration)) //
         .apply(Values.create()));
   }
 
-  public static DetectClickSpikes of(Integer maxClicks, Duration windowDuration) {
-    return new DetectClickSpikes(maxClicks, windowDuration);
+  public static LabelClickSpikes of(Integer maxClicks, Duration windowDuration) {
+    return new LabelClickSpikes(maxClicks, windowDuration);
   }
 
-  private DetectClickSpikes(Integer maxClicks, Duration windowDuration) {
+  private LabelClickSpikes(Integer maxClicks, Duration windowDuration) {
     this.maxClicks = maxClicks;
     this.windowMillis = windowDuration.getMillis();
   }
@@ -76,7 +76,8 @@ public class DetectClickSpikes extends
         .filter(ts -> (currentMillis - ts) < windowMillis)
         // We order with largest timestamps first.
         .sorted(Comparator.reverseOrder())
-        // We conserve memory by keeping only the most recent clicks if we're over the limit.
+        // We conserve memory by keeping only the most recent clicks if we're over the limit;
+        // we allow (maxClicks + 1) elements so size checks can use ">" rather than ">=".
         .limit(maxClicks + 1).collect(Collectors.toList());
     state.write(timestamps);
     return timestamps;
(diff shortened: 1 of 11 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Transform that maintains state per key in order to label suspicious clicks.
+ */
+public class LabelClickSpikes extends
+    PTransform<PCollection<KV<String, PubsubMessage>>, PCollection<KV<String, PubsubMessage>>> {
+
+  private final Integer maxClicks;
+  private final Long windowMillis;
+  private final Counter status64Counter = Metrics.counter(LabelClickSpikes.class,
+      "click_status_64");
```
jklukas, line 45:
```suggestion
  private final Counter status65Counter = Metrics.counter(LabelClickSpikes.class,
      "click_status_65");
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -0,0 +1,138 @@
...
+    state.write(timestamps);
+    return timestamps;
+  }
+
+  /** Updates the passed attribute map, adding click-status to the reporting URL. */
+  private static void addClickStatusToReportingUrlAttribute(Map<String, String> attributes) {
+    String reportingUrl = attributes.get(Attribute.REPORTING_URL);
+    ParsedReportingUrl urlParser = new ParsedReportingUrl(reportingUrl);
+    // "64" indicates a spike in clicks per the Conducive API spec.
+    urlParser.addQueryParam("click-status", "64");
```
jklukas, line 91:
```suggestion
    // "65" indicates a short-term spike in "ghost" clicks per the Conducive API spec.
    urlParser.addQueryParam("click-status", "65");
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -0,0 +1,138 @@
...
+      // it will overwrite this timer value.
+      timer.set(Instant.ofEpochMilli(elementTs.getMillis() + windowMillis));
+
+      if (timestamps.size() <= maxClicks) {
+        out.output(element);
+      } else {
+        PubsubMessage message = element.getValue();
+        Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
+        addClickStatusToReportingUrlAttribute(attributes);
+        status64Counter.inc();
```
jklukas, line 120:
```suggestion
        status65Counter.inc();
```

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    PAssert.that(result).satisfies(iter -> {
+      int size = Iterables.size(iter);
+      assert size == 40 : "Expected 40 messages, but found " + size;
+      return null;
+    });
+
+    PAssert.that(result).satisfies(iter -> {
+      long countWithStatus = StreamSupport.stream(iter.spliterator(), false) //
+          .filter(m -> m.getAttribute(Attribute.REPORTING_URL).contains("click-status=64")) //
```
jklukas, line 67:
```suggestion
          .filter(m -> m.getAttribute(Attribute.REPORTING_URL).contains("click-status=65")) //
```

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+    PAssert.that(result).satisfies(iter -> {
+      int size = Iterables.size(iter);
+      assert size == 16 : "Expected 16 messages, but found " + size;
+      return null;
+    });
+
+    PAssert.that(result).satisfies(iter -> {
+      long countWithStatus = StreamSupport.stream(iter.spliterator(), false) //
+          .filter(m -> m.getAttribute(Attribute.REPORTING_URL).contains("click-status=64")) //
```
jklukas, line 103:
```suggestion
          .filter(m -> m.getAttribute(Attribute.REPORTING_URL).contains("click-status=65")) //
```

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -0,0 +1,113 @@
...
+  @Rule
+  public TestPipeline pipeline = TestPipeline.create();
+
+  @Test
+  public void testSetsClickStatus() {
+    ImmutableMap<String, String> attributes = ImmutableMap.of(Attribute.CONTEXT_ID, "a", //
+        Attribute.REPORTING_URL, "https://test.com");
+    PubsubMessage message = new PubsubMessage(new byte[] {}, attributes);
+    PubsubMessage[] messages = new PubsubMessage[19];
+    Arrays.fill(messages, message);
```
BenWu, line 33: This array isn't used anywhere

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    // Now, we add 20 messages each only a second apart. The first 10 should saturate the timestamp
+    // state, then the final 10 should be marked as suspicious via click-status.
+    for (int i = 1; i <= 20; i++) {
+      eventBuilder = eventBuilder
+          .advanceWatermarkTo(
+              new Instant(0L).plus(Duration.standardMinutes(30)).plus(Duration.standardSeconds(i)))
+          .addElements(message);
+    }
```
BenWu, line 51: I would like to see these as two separate test cases to be explicit about what's be marked with click status

BenWu (commented):
I had to read up on beam states but this looks good to me.  I just have some minor comments and questions.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -0,0 +1,137 @@
...
+    attributes.put(Attribute.REPORTING_URL, urlParser.toString());
+  }
+
+  private class Fn extends DoFn<KV<String, PubsubMessage>, KV<String, PubsubMessage>> {
+
+    // See https://beam.apache.org/documentation/programming-guide/#state-and-timers
+    @StateId("click-state")
+    private final StateSpec<ValueState<List<Long>>> clickState = StateSpecs.value();
+    @TimerId("click-timer")
+    private final TimerSpec clickTimer = TimerSpecs.timer(TimeDomain.EVENT_TIME);
```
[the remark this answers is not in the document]
jklukas: I think you're right here. We're in kind of a squishy space as to what event time means for these clicks; the two concepts are not very different in our case, but processing time is indeed easier to reason about in general. Will change.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -0,0 +1,137 @@
...
+      timer.set(Instant.ofEpochMilli(elementTs.getMillis() + windowMillis));
+
+      if (timestamps.size() <= maxClicks) {
+        out.output(element);
+      } else {
+        PubsubMessage message = element.getValue();
+        Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
+        addClickStatusToReportingUrlAttribute(attributes);
+        ghostClickCounter.inc();
+        out.output(KV.of(element.getKey(), new PubsubMessage(message.getPayload(), attributes)));
```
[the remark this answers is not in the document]
jklukas: Yes, we could. It feels more composable to have the input and output types be identical here, and have the normal case be that the element is passed through unaltered, then strip off the key in a separate step. In reality, it's unlikely we will have any need to use this in a context where we don't strip out the key immediately after the transform. I'll think a bit more about this.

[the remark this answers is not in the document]
jklukas: I think my strongest argument here is that the `LabelClickSpikes.perContextId` factory method ends up having a parallel structure as well. It takes in an unkeyed PCollection, handles wrapping it in a KV, explicitly handles unwrapping it (via `Values.create()`), and returns an unkeyed PCollection.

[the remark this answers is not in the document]
BenWu: Ok sounds reasonable, bit of a nit anyway

BenWu (approved):
Looks good.  Just wondering about the `advanceWatermarkTo` in the test

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -0,0 +1,141 @@
...
+    ImmutableMap<String, String> attributes = ImmutableMap.of(Attribute.CONTEXT_ID, "a", //
+        Attribute.REPORTING_URL, "https://test.com");
+    PubsubMessage message = new PubsubMessage(new byte[] {}, attributes);
+    Builder<PubsubMessage> eventBuilder = TestStream.create(PubsubMessageWithAttributesCoder.of());
+
+    // We add 20 messages each only a second apart. The first 10 should saturate the timestamp
+    // state, then the final 10 should be marked as suspicious via click-status.
+    for (int i = 1; i <= 20; i++) {
+      eventBuilder = eventBuilder
+          .advanceWatermarkTo(new Instant(0L).plus(Duration.standardSeconds(i)))
```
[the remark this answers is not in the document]
jklukas: Good call. I'll update these to `advanceProcessingTime`. It's not clear to me why `advanceWatermarkTo` still works here, but may be an artifact of how `TestStream` is implemented.

PR #1751, opened 2021-07-09
Bug 1716665 Record whether client payload was compressed in Decoder

jklukas:
Relevant to https://bugzilla.mozilla.org/show_bug.cgi?id=1716665 and a dependency for https://github.com/mozilla/gcp-ingestion/pull/1740

When a client payload comes into the decoder compressed, we now add `client_compression=gzip` to the attribute map. If the payload is not compressed, we don't add the attribute. So downstream transforms can check for existence of this attribute to determine whether the original payload was compressed.

Note that we don't currently have `client_compression` present in the payload_bytes_* schemas, so we'll lose this information if we need to replay messages from a BQ source. I think that's fine for now and we can circle back, adding it to schemas if we find this is broadly useful information.


ingestion-beam/src/test/java/com/mozilla/telemetry/DecoderMainTest.java
```diff
@@ -74,28 +74,6 @@ public void testMixedErrorCases() throws Exception {
         matchesInAnyOrder(expectedErrorOutputLines));
   }
 
-  @Test
```
jklukas, line 77: I've removed gzip-specific cases out of the integration tests in favor of having the main `input.ndjson` contain one gzipped document.

BenWu (approved):

jklukas:
cc @whd

PR #1759, opened 2021-07-14
Bug 1698934 - Add example for deletion-request ping

acmiyaguchi:
[bug link](https://bugzilla.mozilla.org/show_bug.cgi?id=1698934)

Added an example in CI to test for the deletion-request example


akkomar (approved):
+1

PR #1789, opened 2021-08-16
Remove support for Dataflow classic templates

### LINKED ISSUE #1534: Use Dataflow Flex Templates and remove use of ValueProvider

relud:
fixes #1534


ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -147,30 +144,27 @@ public FileOutput(ValueProvider<String> outputPrefix, OutputFileFormat format,
...
+              .defaultNaming(pathTemplate.replaceDynamicPart(placeholderValues), format.suffix()));
 
       if (inputType == InputType.pubsub) {
-        // Passing a ValueProvider to withNumShards disables runner-determined sharding, so we
-        // need to be careful to pass this only for streaming input (where runner-determined
-        // sharding is not an option).
+        // withNumShards disables runner-determined sharding, so we need to be careful to pass this
+        // only for streaming input (where runner-determined sharding is not an option).
         write = write.withNumShards(numShards);
       }
```
relud, line 169: I'm unclear on whether putting this in a condition is still necessary, because we aren't using ValueProvider, but we do set a default of `100`.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
```diff
@@ -276,18 +268,17 @@ public void processElement(ProcessContext ctx) {
...
+              .defaultNaming(pathTemplate.replaceDynamicPart(placeholderValues), ".avro"));
 
       if (inputType == InputType.pubsub) {
-        // Passing a ValueProvider to withNumShards disables runner-determined sharding, so we
-        // need to be careful to pass this only for streaming input (where runner-determined
-        // sharding is not an option).
+        // withNumShards disables runner-determined sharding, so we need to be careful to pass this
+        // only for streaming input (where runner-determined sharding is not an option).
         write = write.withNumShards(numShards);
       }
```
relud, line 283: unsure if condition is necessary here, too

bin/mvn
```diff
@@ -45,5 +45,5 @@ docker run $INTERACTIVE_FLAGS --rm \
     -e GOOGLE_APPLICATION_CREDENTIALS \
     $(for e in $PASS_ENV; do echo "-e $e"; done) \
     $MOUNT_CREDENTIALS_FLAGS \
-    maven:3-jdk-8 \
+    maven:3-jdk-11 \
```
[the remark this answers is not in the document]
jklukas: Does this change actually determine which JDK is being used in our deployed Dataflow jobs?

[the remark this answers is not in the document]
relud: this change has no impact on our dataflow jobs, only on devs using this script to run maven

ingestion-beam/Dockerfile
```diff
@@ -8,7 +8,7 @@ COPY ingestion-beam/checkstyle /app/ingestion-sink/checkstyle
 RUN mvn prepare-package
 COPY ingestion-core/src /app/ingestion-core/src
 COPY ingestion-beam/src /app/ingestion-beam/src
-RUN mvn package -Dmaven.test.skip=true
+RUN mvn package -Dmaven.test.skip=true -Dmaven.compiler.release=11
```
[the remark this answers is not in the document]
relud: this change does not determine what JDK is being used, just what JDK the jars are compiled for

docs/architecture/overview.md
```diff
@@ -80,29 +80,23 @@ This document specifies the architecture for GCP Ingestion as a whole.
...
   - Must only deliver messages with configured destinations
 - Must accept optional configuration for sampling telemetry data
-  - Must accept a compile-time parameter defining a topic pattern string
-    (may be promoted to runtime if Dataflow adds support for PubSub topic
-    names defined via `NestedValueProvider`)
-  - Must accept compile-time parameters defining the sampling ratio for
-    each channel (nightly, beta, and release)
+  - Must accept configuration for the destination topic pattern
+  - Must accept configuration for the sampling ratio for each channel (nightly,
+    beta, and release)
```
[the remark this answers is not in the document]
whd: LGTM

jklukas (approved):
LGTM

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
jklukas: The docs for `withNumShards(int numShards)` specify:

> 0 means runner-determined sharding. Specifying a non-zero value may hurt performance, because it will limit the parallelism of writing

So I think the right thing to do now is to change our default to 0 and document that 0 means runner-determined sharding.

jklukas: Actually, I take that back. We should still ignore `numShards` in the batch case. This conditional still makes sense, and it still makes sense to have 100 as the default value that we pass when we have a streaming input.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Write.java
jklukas: I think this conditional still makes sense. We should always use the default runner-determined sharding when in batch mode.

PR #1792, opened 2021-08-18
Bug 1723994 - Add dma-code to topsites urls

BenWu:
https://bugzilla.mozilla.org/show_bug.cgi?id=1723994


jklukas (approved):
Looks great 👍

PR #1793, opened 2021-08-18
Wait for PVCs to remain detached before deleting

relud:


whd (approved):
LGTM, hopefully this does the trick.

PR #1794, opened 2021-08-19
Remove fixed list of query params for impression aggregation

BenWu:
This should fix it but this seems like this can easily happen again and it might be tricky to catch in unit tests.  I think the aggregation fields was used to reduce the number of keys to group by but I'm not sure if it's useful.  I can look at the query params that we're actually receiving and see if this is needed.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
@@ -1,7 +1,6 @@
 package com.mozilla.telemetry.contextualservices;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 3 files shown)


BenWu:
I didn't find any impressions urls that have unexpected query params so having an allow list of params shouldn't be needed.

Redash isn't working for me right now but this is the query I used to check:
```
SELECT
  REGEXP_EXTRACT_ALL(reporting_url, '[\\?&]([^=]+=[^&]*)') AS query_params,
  reporting_url,
FROM
  `moz-fx-data-shared-prod.contextual_services_stable.topsites_impression_v1`
WHERE
  DATE(submission_timestamp) = "2021-08-18"
ORDER BY
  ARRAY_LENGTH(query_params) DESC
```

jklukas (approved):
Removing the list of allowed fields seems reasonable. I think the headache of forgetting to add future fields to the list is probably greater than the risk of an unexpected param showing up here.

BenWu:
Repurposed this pr to remove the fixed list so requesting re-review

jklukas (approved):
Looks great 👍

PR #1811, opened 2021-08-31
Run all maven tests as one matrix and include ingestion-core

relud:


.circleci/config.yml
```diff
@@ -165,113 +165,74 @@ jobs:
...
-        name: Maven Integration Test
-        # Google's client libraries will check for GOOGLE_APPLICATION_CREDENTIALS
-        # and use a file in that location for credentials if present;
-        # See https://cloud.google.com/docs/authentication/production
+          - &cache_key
+            maven-repo-v1-<< parameters.module >>-{{ .Branch }}-{{ checksum "../pom.xml" }}-{{ checksum "../ingestion-core/pom.xml" }}-{{ checksum "pom.xml" }}
+          - maven-repo-v1-<< parameters.module >>-{{ .Branch }}-{{ checksum "../pom.xml" }}-{{ checksum "../ingestion-core/pom.xml" }}-
+          - maven-repo-v1-<< parameters.module >>-{{ .Branch }}-{{ checksum "../pom.xml" }}-
+          - maven-repo-v1-<< parameters.module >>-{{ .Branch }}-
+          - maven-repo-v1-<< parameters.module >>-main-
```
relud, line 191: prevent unusual cache behavior by falling back to the main branch, instead of an arbitrary branch

.circleci/config.yml
```diff
@@ -165,113 +165,74 @@ jobs:
...
+    - when:
+        condition: << parameters.integration >>
+        steps:
+          - run:
+              name: Early return if this build is from a forked PR
+              command: |
+                if [ -n "$CIRCLE_PR_NUMBER" ]; then
+                  echo "Cannot pass creds to forked PRs, so marking this step successful"
+                  circleci step halt
+                fi
```
relud, line 201: This `when` condition ensures that this step only appears for integration tests, but `$CIRCLE_PR_NUMBER` doesn't have an equivalent [pipeline value](https://circleci.com/docs/2.0/pipeline-variables/#pipeline-values), so the bash if statement is still necessary. 

It may be possible to use `<< pipeline.project.git_url >>`, but I dislike that it would hard-code the current url, e.g.:
```suggestion
        condition:
          and:
            - << parameters.integration >>
            - not:
                equal: [https://github.com/mozilla/gcp-ingestion, << pipeline.project.git_url >>]
        steps:
          - run:
              name: Early return for build from a forked PR
              command: |
                echo "Cannot pass creds to forked PRs, so marking this step successful"
                circleci step halt
```

.circleci/config.yml
```diff
@@ -165,113 +165,73 @@ jobs:
...
+    - when:
+        condition:
+          equal: [ingestion-beam, << parameters.module >>]
+        steps:
+          - run:
+              name: Download resources needed for Beam tests
+              command: |
+                bin/download-schemas
+    - run:
+        name: Maven <<# parameters.integration >>Integration<</ parameters.integration >> Test
```
relud, line 211: This syntax of using boolean parameters for text blocks comes from https://circleci.com/docs/2.0/reusing-config/#boolean with `<<# ... >>` for true conditions and `<<^ ... >>` for false conditions.

relud (commented):
My primary motivation here is that spotless and checkstyle aren't being otherwise enforced on ingestion-core, but I also think it would be a good thing to require ingestion-core tests to pass independently

.circleci/config.yml
jklukas: It would be worth adding a code comment here to document this.

.circleci/config.yml
jklukas: Worth adding this link as a code comment.

.circleci/config.yml
```diff
@@ -165,113 +165,73 @@ jobs:
...
         environment: *edge_compose_env
 
-  ingestion-beam: &ingestion_beam
-    working_directory: /root/project/ingestion-beam
+  maven:
     parameters:
       executor:
         type: executor
+      integration:
+        type: boolean
```
jklukas, line 173: Would it be more elegant to have a `scope` parameter that's an enum of `integration` and `unit` instead of a boolean `integration` param?

jklukas (approved):
r+wc

Overall, I think this change is great for clarity and reducing duplication.

.circleci/config.yml
relud: sure

PR #1815, opened 2021-09-07
Bug 1729069 Handle static GCLB IP entry

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1729069#c8

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
@@ -14,6 +14,12 @@ import org.apache.beam.sdk.options.Validation;
  */
 public interface DecoderOptions extends SinkOptions, PipelineOptions {
 
+  @Description("Static IP associated with load balancer; will be ignored for IP lookups")
+  @Default.String("35.227.207.240")
+  String getProxyIpAddress();
+
+  void setProxyIpAddress(String value);
+
   @Description("Path (local or gs://) to GeoIP2-City.mmdb")
   @Validation.Required
   String getGeoCityDatabase();
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
@@ -189,7 +189,7 @@ public class GeoCityLookup
           // https://cloud.google.com/load-balancing/docs/https/#components
           //
           // In practice, many of the "first" addresses are bogus or internal,
-          // so we target the immediate sending client IP.
+          // so we target the immediate sending client IP by choosing the second-to-last entry.
           String[] ips = xff.split("\\s*,\\s*");
           ip = ips[Math.max(ips.length - 2, 0)];
           countIpForwarded.inc();
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
@@ -128,8 +128,13 @@ public abstract class Read extends PTransform<PBegin, PCollection<PubsubMessage>
             // We have to take care not to read additional bytes; see
             // https://github.com/mozilla/gcp-ingestion/issues/1266
             ByteBuffer byteBuffer = (ByteBuffer) record.get(FieldName.PAYLOAD);
-            byte[] payload = new byte[byteBuffer.limit()];
-            byteBuffer.get(payload);
+            final byte[] payload;
+            if (byteBuffer != null) {
+              payload = new byte[byteBuffer.limit()];
+              byteBuffer.get(payload);
+            } else {
+              payload = new byte[0];
+            }
 
             // We populate attributes for all simple string and timestamp fields, which is complete
             // for raw and error tables.
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
@@ -65,7 +65,7 @@ public class ParseProxyTest extends TestWithDeterministicJson {
             + "},\"payload\":\"notProxied++\"}",
         "{\"attributeMap\":" //
             + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
-            + ",\"x_forwarded_for\":\"4, 3, 6.7.8.9\"" //
+            + ",\"x_forwarded_for\":\"4,3,1\"" //
             + "},\"payload\":\"statroxied++\"}",
         "{\"attributeMap\":" //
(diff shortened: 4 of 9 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -14,6 +14,12 @@
  */
 public interface DecoderOptions extends SinkOptions, PipelineOptions {
 
+  @Description("Static IP associated with load balancer; will be ignored for IP lookups")
+  @Default.String("35.227.207.240")
+  String getProxyIpAddress();
+
+  void setProxyIpAddress(String value);
+
```
jklukas, line 22: For simplicity, I've chosen to have this new config option be a single-entry string, and have the default set to the expected prod value.

This could be turned into a comma-separated list in the future.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -189,7 +189,7 @@ public PubsubMessage apply(PubsubMessage message) {
           // https://cloud.google.com/load-balancing/docs/https/#components
           //
           // In practice, many of the "first" addresses are bogus or internal,
-          // so we target the immediate sending client IP.
+          // so we target the immediate sending client IP by choosing the second-to-last entry.
```
jklukas, line 192: Just making this documentation a bit more explicit.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
```diff
@@ -59,6 +63,10 @@ public void testOutput() {
             + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
             + ",\"x_forwarded_for\":\"3, 2, 1\"" //
             + "},\"payload\":\"notProxied++\"}",
+        "{\"attributeMap\":" //
+            + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
+            + ",\"x_forwarded_for\":\"4, 3, 6.7.8.9\"" //
```
jklukas, line 68: Note that the logic as written currently in this PR is attempting to follow exactly what's written in https://bugzilla.mozilla.org/show_bug.cgi?id=1729069#c8; itchecks for the `proxyIpAddress`, and if it exists strips the _trailing_ entry from the IP. This leads to the static IP still existing in X-Forwarded-For, although the later IP lookups only care about the second-to-last entry, so this works.

It would make the most sense to me to modify `ParseProxy` to parse X-Forwarded-For into an array of IPs, check each entry against `proxyIpAddress`. If we find a match, we remove it, and then `mkString` the array back into a comma-separated list. In that case, this test case would change to `4,3,1` rather than `4, 3, 6.7.8.9`.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
```diff
@@ -40,6 +40,10 @@ public void testOutput() {
             + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
             + ",\"x_forwarded_for\":\"3, 2, 1\"" //
             + "},\"payload\":\"notProxied++\"}",
+        "{\"attributeMap\":" //
+            + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
+            + ",\"x_forwarded_for\":\"4, 3, 6.7.8.9, 1\"" //
+            + "},\"payload\":\"statroxied++\"}",
```
jklukas, line 46: @relud I was getting errors when I tried to put a better name in the payload. I believe this is getting interpreted as a base64-encoded value, so it's kind of a crapshoot as to whether the nice name you put here is going to work or not. Is that true? Any advice on what to put in the payload here?

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
whd: I agree that parsing this into a list and removing `proxyIpAddress` if extant makes the most sense. I wrote the logic in the comment to be as similar to what we already had as possible, but this should achieve the same effect more clearly.

whd (commented):
This seems more comprehensible than what we had before. I'm going to run this in stage off of this branch to verify it has the expected behavior.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
relud: payload is definitely base64 encoded, the `+`s are being used for padding to make it valid base64, so in this case I'd use:
```suggestion
            + "},\"payload\":\"staticProxied+++\"}",
```

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseProxyTest.java
```diff
@@ -59,6 +63,10 @@ public void testOutput() {
             + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
             + ",\"x_forwarded_for\":\"3, 2, 1\"" //
             + "},\"payload\":\"notProxied++\"}",
+        "{\"attributeMap\":" //
+            + "{\"submission_timestamp\":\"2000-01-01T00:00:00.000000Z\"" //
+            + ",\"x_forwarded_for\":\"4,3,1\"" //
+            + "},\"payload\":\"statroxied++\"}",
```
relud, line 69: and here:
```suggestion
            + "},\"payload\":\"staticProxied+++\"}",
```

relud (approved):
r+ comments optional

whd (approved):
Ok re-running https://bugzilla.mozilla.org/show_bug.cgi?id=1729069#c10 from this branch and https://github.com/mozilla-services/cloudops-infra/compare/beam_newopt?expand=1 produced the correct result.

jklukas:
I want to run a sample backfill job using this code and verify the geo distribution before actually merging this.

ingestion-beam/src/main/java/com/mozilla/telemetry/io/Read.java
```diff
@@ -128,8 +128,13 @@ public BigQueryInput(String tableSpec, BigQueryReadMethod method, Source source,
...
             ByteBuffer byteBuffer = (ByteBuffer) record.get(FieldName.PAYLOAD);
-            byte[] payload = new byte[byteBuffer.limit()];
-            byteBuffer.get(payload);
+            final byte[] payload;
+            if (byteBuffer != null) {
+              payload = new byte[byteBuffer.limit()];
+              byteBuffer.get(payload);
+            } else {
+              payload = new byte[0];
+            }
```
jklukas, line 137: Added this code to handle null payload for backfill taking place in https://github.com/mozilla/bigquery-backfill/pull/15

whd:
I'm merging this and deploying today.

PR #1824, opened 2021-09-16
Publish docs on ContextualServicesReporter

jklukas:
This incorporates the content of [the internal document that Ben wrote up](https://docs.google.com/document/d/1RK4C-mc0TBKyEBLS0hotbCg2jEHBcZ0K4EdFj8rMaQ4/edit#) with a few modifications and makes it publicly available.


docs/ingestion-beam/contextual-services-job.md
```diff
@@ -1,8 +1,104 @@
 # Contextual Services Reporter Job
 
 The contextual services reporter forwards contextual-services click and impression events
-to an external partner in a privacy preserving way.
+to an external partner, preserving user privacy by minimizing the scope of contextual metadata.
```
wlach, line 4: I wonder if we should make this type of claim here, or if we should just state that we *do* minimize the scope of contextual metadata (and maybe briefly summarizing how). 

You can always link to the data blog post, which I assume has been reviewed as an official company statement: https://blog.mozilla.org/data/2021/09/15/data-and-firefox-suggest/

wlach (approved):
This makes sense to me. I have some minor wording suggestions.

docs/ingestion-beam/contextual-services-job.md
jklukas: I softened this statement and added a link to the blog post as you suggested.

PR #1826, opened 2021-09-16
Contextual services docs fix

jklukas:
Follow-up to #1825 and #1824. I actually ran `mkdocs` locally this time to verify the diagram is getting pulled in.


wlach (approved):
Part of me is tempted to add a netlify deploy preview so we can get an idea of what docs change look like. I guess we can see if this problem comes up again.

PR #1896, opened 2021-11-19
Bug 1742172 Add SanitizeAttributes for timestamp granularity

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1742172 for more on the
design here.


whd (commented):
Looks like tests aren't updated/passing and docs need an update but interface LGTM. I assume the granularity will be an enum in the MPS metaschema.

jklukas:
The failures I think are due to the `schemas.tar.gz` not having the correctly updated from this PR. Trying to understand that.

jklukas:
Tests should be passing now, and the mps change is up at https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/707

docs/architecture/overview.md
```diff
@@ -64,6 +64,7 @@ This document specifies the architecture for GCP Ingestion as a whole.
   1. Validate `payload` using a JSON Schema determined by attributes
   1. Parse `agent` attribute into `user_agent_*` attributes
   1. Produce `normalized_` variants of select attributes
+  1. Allow sanitization of attributes, configurable per document type via schema metadata
```
whd, line 67: Why is this `Allow sanitization` instead of `Sanitize` as in the other doc string?

whd (approved):
LGTM

docs/architecture/overview.md
jklukas, line 67:
```suggestion
  1. Sanitize attributes, configurable per document type via schema metadata
```

jklukas: Fixed.

PR #1898, opened 2021-11-24
Bug 1742172 Support override_attributes metadata

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1742172

See https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/709 for the coordinated change to schemas.

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/SanitizeAttributesTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/SanitizeAttributesTest.java
@@ -4,6 +4,7 @@ import com.google.common.collect.ImmutableMap;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
 import com.mozilla.telemetry.util.TestWithDeterministicJson;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.Map;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.testing.PAssert;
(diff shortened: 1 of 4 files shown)


jklukas:
cc @hackebrot since we discussed these tests recently, and this PR rolls in some improvements.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/SanitizeAttributesTest.java
```diff
@@ -17,42 +18,82 @@
...
 
   @Test
   public void testSubmissionTimestampGranularity() {
     // Depends on custom doctype injected in bin/download-schemas
-    Map<String, String> attributes = ImmutableMap.<String, String>builder()
+    Map<String, String> inputAttributes = ImmutableMap.<String, String>builder()
         .put(Attribute.DOCUMENT_NAMESPACE, "test") //
         .put(Attribute.DOCUMENT_TYPE, "seconds") //
         .put(Attribute.DOCUMENT_VERSION, "1") //
-        .put("submission_timestamp", "2021-01-01T12:00:00.123456Z") //
```
whd, line 48: I don't have enough familiarity with the code to understand why this wasn't an `Attribute` before but it looks uniform now.

whd (approved):
LGTM from the main logic and test, the JSON deserializing `Builder` stuff :shrug:

relud (approved):

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/SanitizeAttributesTest.java
jklukas: This is just cleaning up an oversight. That value of the constant is equal to `"submission_timestamp"`.

PR #1935, opened 2022-01-03
Update Snappy dep for Apple Silicon

jklukas:
The snappy-java library supports M1 processors as of 1.1.8.2; see
https://github.com/xerial/snappy-java/blob/master/Milestone.md#snappy-java-1182-2020-11-28

This change should be relatively safe, as snappy pulls in no transitive deps and we do not actually rely on snappy compression in production currently.

With this change, I was able to run `mvn clean test` successfully in this repo on an M1 Mac.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -62,6 +62,9 @@
         <jackson.version>2.12.4</jackson.version>
         <avro.version>1.8.2</avro.version>
 
+        <!-- The version pulled in by avro does not support Apple Silicon as of Jan 2022 -->
+        <snappy.version>1.1.8.3</snappy.version>
+
         <!-- Set empty default. -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -131,6 +134,11 @@
             <artifactId>commons-compress</artifactId>
             <version>1.21</version>
         </dependency>
+        <dependency>
+            <groupId>org.xerial.snappy</groupId>
+            <artifactId>snappy-java</artifactId>
+            <version>${snappy.version}</version>
```
relud, line 140: this should be in dependencyManagement because we don't use it directly

relud (approved):
r+wc

pom.xml
jklukas: Moved.

PR #1936, opened 2022-01-03
Disallow search terms ingestion via telemetry

jklukas:
Reverts #1840 which turned off the scrubbing originally introduced in #1708 and #1705

See https://bugzilla.mozilla.org/show_bug.cgi?id=1712850

We want to enable this scrubbing again during the week of Jan 11 per the decision in the doc
[Remove support for search_query in Suggest impression telemetry](https://docs.google.com/document/d/1uRI6xoetGroMPksyboPOL-lT7BOZdYsn1y7lr4H7BW0/edit). We will wait until Fx 96 has uptake above 80% before we merge this change.


scholtzan (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -290,8 +298,6 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       throw new AffectedByBugException("1733118");
     }
```
jklukas, line 300:
```suggestion
  }

```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -289,7 +297,6 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       // See also https://bugzilla.mozilla.org/show_bug.cgi?id=1733118
       throw new AffectedByBugException("1733118");
     }
-
   }
```
jklukas, line 300:
```suggestion

  }
```

PR #1953, opened 2022-01-19
Bug 1751034 Rename url -> url2 even if no url field exists

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1751034


relud (approved):

whd (approved):
For my reference, this code runs at ingestion-sink before final load into BQ?

jklukas:
Thank you for quick review here!

PR #1956, opened 2022-01-24
Bug 1751753 Sanitize search engine codes

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1751753 for context


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +346,36 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+        });
+  }
+
+  private static void processKeysForBug1751753(JsonNode inputNode) {
+    if (!inputNode.isObject()) {
+      // If this isn't an object, we have nothing to do here.
+      return;
+    }
+    ObjectNode searchNode = (ObjectNode) inputNode;
+    searchNode.fields().forEachRemaining(entry -> {
```
jklukas, line 366: I'll need to recast this as currently it raises ConcurrentModificationException in some cases.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +90,19 @@
       .put("/submit/sslreports/", "1585144") //
       .build();
 
+  // Static list prepared for bug 1751753, pulled from
+  // https://searchfox.org/mozilla-central/source/services/settings/dumps/main/search-telemetry-v2.json
+  // TODO: Pull list from RemoteSettings.
```
whd, line 95: This link 404s for me.

I think in future versions of this we'll want to make this an artifact or similar, as we don't have outbound access from the decoder and because it's generally safer to rely on e.g. gcs.

jklukas: I'll need to better understand details of how often this collection would be changing and whether we'd be at risk of a race condition where the pipeline might sanitize legitimate search codes due to only updating the list on pipeline restart.

jklukas: > This link 404s for me.

The change with this new file is not merged yet.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +90,19 @@
...
+  private static final Set<String> ALLOWED_SEARCH_CODES = ImmutableSet.of("MOZ2", "MOZ4", "MOZ5",
+      "MOZA", "MOZB", "MOZD", "MOZE", "MOZI", "MOZM", "MOZO", "MOZT", "MOZW", "MOZSL01", "MOZSL02",
+      "MOZSL03", "monline_dg", "monline_4_dg", "monline_7_dg", "firefox-a", "firefox-b",
+      "firefox-b-1", "firefox-b-ab", "firefox-b-1-ab", "firefox-b-d", "firefox-b-1-d",
+      "firefox-b-e", "firefox-b-1-e", "firefox-b-m", "firefox-b-1-m", "firefox-b-o",
+      "firefox-b-1-o", "firefox-b-lm", "firefox-b-1-lm", "firefox-b-lg", "firefox-b-huawei-h1611",
+      "firefox-b-is-oem1", "firefox-b-oem1", "firefox-b-oem2", "firefox-b-tinno", "firefox-b-pn-wt",
+      "firefox-b-pn-wt-us", "ubuntu", "ffab", "ffcm", "ffhp", "ffip", "ffit", "ffnt", "ffocus",
+      "ffos", "ffsb", "fpas", "fpsa", "ftas", "ftsa", "newext", "none");
```
jklukas, line 104: Need to update to include `monline_3_dg`.

jklukas: Also, we may want to explicitly add "other" to this list (although it would just be replaced with "other" anyway).

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +347,42 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
             .matches("0[.]([0-9]|1[0-6])[.].*"); // < 0.17
   }
 
+  // See bug 1751753 for explanation of context.
+  private static void processForBug1751753(ObjectNode json) {
+    // Sanitize keys in the SEARCH_COUNTS histogram.
+    JsonNode searchCounts = json.path("payload").path("keyedHistograms").path("SEARCH_COUNTS");
```
jklukas, line 353: Can we be guaranteed of the capitalization here? The BQ ingestion logic lowercases all field names, so we tolerate casing differences from that perspective. But the logic as written here would miss sanitizing the field if the histogram were recorded as "search_counts" in the JSON.

Maybe this is something @chutten can comment on.

chutten: I confirm that histogram names are SCREAMING_SNAKE_CASE when they leave Firefox Desktop. Scalars and Events are different, as are all other parts of other ping payloads.

jklukas: @chutten indicates that histogram names are reliably SCREAMING_SNAKE_CASE in what the browser sends.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +90,26 @@
       .put("/submit/sslreports/", "1585144") //
       .build();
 
+  private static final String REDACTED_SEARCH_CODE_VALUE = "other";
```
jklukas, line 93: There is ongoing discussion about whether we should use a different value here in order to allow distinguishing client-side sanitization (which uses "other") and pipeline-side sanitization.

jklukas:
Marking @sean-rose (who has recently implemented some JSON manipulation in this repo) and @ANich (who has search probe familiarity) for code review here.

Once we're ready to merge, I'd plan to let this go to stage and monitor live tables for a bit to validate we're seeing a reasonable frequency of "other.scrubbed" values before letting it go to prod.

I'll also ask for final validation of the code list from the wider team working on this before merging.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +90,26 @@
...
+      // https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/search-telemetry-v2/records
+      "MOZ2", "MOZ4", "MOZ5", "MOZA", "MOZB", "MOZD", "MOZE", "MOZI", "MOZM", "MOZO", "MOZT",
+      "MOZW", "MOZSL01", "MOZSL02", "MOZSL03", "monline_dg", "monline_3_dg", "monline_4_dg",
+      "monline_7_dg", "firefox-a", "firefox-b", "firefox-b-1", "firefox-b-ab", "firefox-b-1-ab",
+      "firefox-b-d", "firefox-b-1-d", "firefox-b-e", "firefox-b-1-e", "firefox-b-m",
+      "firefox-b-1-m", "firefox-b-o", "firefox-b-1-o", "firefox-b-lm", "firefox-b-1-lm",
+      "firefox-b-lg", "firefox-b-huawei-h1611", "firefox-b-is-oem1", "firefox-b-oem1",
+      "firefox-b-oem2", "firefox-b-tinno", "firefox-b-pn-wt", "firefox-b-pn-wt-us", "ubuntu",
+      "ffab", "ffcm", "ffhp", "ffip", "ffit", "ffnt", "ffocus", "ffos", "ffsb", "fpas", "fpsa",
+      "ftas", "ftsa", "newext"
```
jklukas, line 109: I recreated this list based on the content of https://phabricator.services.mozilla.com/D136768 about an hour ago. It also explicitly allows "none" and "other".

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
             .matches("0[.]([0-9]|1[0-6])[.].*"); // < 0.17
   }
 
+  // See bug 1751753 for explanation of context.
+  private static void processForBug1751753(ObjectNode json) {
+    // Sanitize keys in the SEARCH_COUNTS histogram.
+    JsonNode searchCounts = json.path("payload").path("keyedHistograms").path("SEARCH_COUNTS");
+    processKeysForBug1751753(searchCounts);
```
sean-rose, line 361: IMO it would be cleaner to check if this was an object before calling `processKeysForBug1751753()`, and then you could cast it as `ObjectNode` on the way in and avoid doing that check/cast in `processKeysForBug1751753()`.
```suggestion
    if (searchCounts.isObject()) {
      processKeysForBug1751753((ObjectNode) searchCounts);
    }
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+
+    // Sanitize keys in browser.search.content.* keyed scalars.
+    json.path("payload").path("processes").path("parent").path("keyedScalars") //
+        .fields().forEachRemaining(entry -> {
+          if (entry.getKey().startsWith("browser.search.content.")) {
+            processKeysForBug1751753(entry.getValue());
+          }
```
sean-rose, line 368:
```suggestion
          if (entry.getKey().startsWith("browser.search.content.") && entry.getValue().isObject()) {
            processKeysForBug1751753((ObjectNode) entry.getValue());
          }
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+        });
+  }
+
+  private static void processKeysForBug1751753(JsonNode inputNode) {
+    if (!inputNode.isObject()) {
+      // If this isn't an object, we have nothing to do here.
+      return;
+    }
+    ObjectNode searchNode = (ObjectNode) inputNode;
```
sean-rose, line 377:
```suggestion
  private static void processKeysForBug1751753(ObjectNode searchNode) {
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+    Lists.newArrayList(searchNode.fieldNames()).forEach(name -> {
```
sean-rose, line 378: It seems odd to me to rely on this ancillary `Lists` class from the Google API client.  I'd suggest importing `org.apache.commons.collections.IteratorUtils` and using that:
```suggestion
    IteratorUtils.toList(searchNode.fieldNames()).forEach(name -> {
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+      if (name.contains(":")) {
```
sean-rose, line 379: Even assuming the only current search keys that contain colons are the in-content ones, if any future keys are added containing colons then this will mangle those, so I'd recommend being more precise with this check.
```suggestion
      if (name.contains(".in-content:")) {
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+        // This is an in-content search with format:
+        // <provider>.in-content:[sap|sap-follow-on|organic]:[code|none]
```
sean-rose, line 381:
```suggestion
        // <provider>.in-content:[sap|sap-follow-on|organic]:[<code>|none]
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +354,45 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+        final String code = StringUtils.substringAfterLast(name, ":");
+        if (ALLOWED_SEARCH_CODES.contains(code)) {
+          // Search code is allowed; no modification needed for this key.
+        } else {
```
sean-rose, line 385:
```suggestion
        if (!ALLOWED_SEARCH_CODES.contains(code)) {
```

sean-rose (changes requested):

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +358,40 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+          }
+        });
+  }
+
+  private static void processKeysForBug1751753(ObjectNode searchNode) {
+    Lists.newArrayList(searchNode.fieldNames()).forEach(name -> {
+      if (name.contains(".in-content")) {
+        // This is an in-content search with format:
+        // <provider>.in-content:[sap|sap-follow-on|organic]:[<code>|none]
+        final String code = StringUtils.substringAfterLast(name, ":");
```
jklukas, line 383: I'm realizing this piece is a bit fragile. If the query param content contains a ":" character, then this could end up stripping only part of the user-provided value. I may need to use regexes and have different logic for the histogram and scalar cases.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +91,39 @@
...
 
+  private static final String REDACTED_SEARCH_CODE_VALUE = "other.scrubbed";
+
+  // The key format is <provider>.in-content:[sap|sap-follow-on|organic]:[<code>|none]
+  private static final Pattern DESKTOP_SEARCH_COUNTS_PATTERN = Pattern
+      .compile("(.+\\.in-content.[^:]+:)(.*)");
+
+  // The key format is <provider>:[tagged|tagged-follow-on|organic]:[<code>|none]
+  private static final Pattern DESKTOP_SEARCH_CONTENT_PATTERN = Pattern
+      .compile("([^:]+:[^:]+:)(.*)");
```
jklukas, line 102: I've switched to using two different regex expressions as the basis for matching and for parsing out which part is the `code` to be redacted. This should be more robust than the previous method of just stripping the part after the final `:` character.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -4,13 +4,17 @@
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.Lists;
```
sean-rose, line 7: Nit: this isn't in alphabetical order with the other imports.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -87,6 +91,39 @@
       .put("/submit/sslreports/", "1585144") //
       .build();
 
+  private static final String REDACTED_SEARCH_CODE_VALUE = "other.scrubbed";
+
+  // The key format is <provider>.in-content:[sap|sap-follow-on|organic]:[<code>|none]
+  private static final Pattern DESKTOP_SEARCH_COUNTS_PATTERN = Pattern
+      .compile("(.+\\.in-content.[^:]+:)(.*)");
```
sean-rose, line 98: This could be slightly more precise:
```suggestion
      .compile("([^.]+\\.in-content[:.][^:]+:)(.*)");
```

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -327,4 +368,41 @@ private static boolean bug1733118Affected(String namespace, String docType, Obje
...
+        });
+  }
+
+  private static void processKeysForBug1751753(ObjectNode searchNode, Pattern pattern) {
+    Lists.newArrayList(searchNode.fieldNames()).forEach(name -> {
+      Matcher match = pattern.matcher(name);
+      if (match.matches()) {
+        // Group 0 is the entire match;
+        // Group 1 is everything before the code, including trailing separator;
+        // Group 2 is the code
```
sean-rose, line 394: Using named capture groups could avoid the need for such comments.

sean-rose (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
jklukas: I still haven't gotten my IntelliJ settings back to what I want after laptop refresh. I used to be able to rely on it to do the right thing automatically.

jklukas:
Thanks for the thorough review, @sean-rose!

I'm going to let this propagate to stage and will validate in about an hour. I'll tag @whd to deploy to prod once that's done.

PR #1962, opened 2022-01-27
Bug 1751955 Fix edge cases for mobile search key sanitization

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1751955#c8

This is a fixup for #1961 which was merged and deployed to stage.
We need to get this change in before deploying mobile sanitization to prod.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -126,9 +127,17 @@
...
       // Yahoo
-      "monline_dg", "monline_3_dg", "monline_4_dg", "monline_7_dg"
-  // End of copied values.
-  );
+      "monline_dg", "monline_3_dg", "monline_4_dg", "monline_7_dg",
+      // End copied desktop codes.
+      // Additionally, these Baidu codes are relevant for mobile browsers.
+      "1000969a", "1000969b");
+
+  private static final Set<String> ALLOWED_SEARCH_CODES_LOWER = ALLOWED_SEARCH_CODES
```
sean-rose, line 135: - I don't think `ALLOWED_SEARCH_CODES_LOWER` quite accurately captures what's happening, since there's also the leading underscores mobile adds to codes starting with digits (it baffles me why they're doing that in the mobile client).
- If desktop ever adds a search code that starts with a digit I don't believe this arrangement would continue work, because AFAIK desktop isn't adding leading underscores like mobile does.

I'd suggest:
- Renaming `ALLOWED_SEARCH_CODES` to `ALLOWED_SEARCH_CODES_DESKTOP` and removing the mobile-only Baidu codes from that list.
- Renaming this `ALLOWED_SEARCH_CODES_MOBILE` and construct it as `ALLOWED_SEARCH_CODES_DESKTOP` + Baidu codes without leading underscores -> lowercase -> add leading underscores to codes starting with digits.
- Using the those two sets as appropriate in the `sanitizeDesktopSearchKeys` and `sanitizeMobileSearchKeys` methods rather than trying to use a single `isAllowedSearchCode` method.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -438,8 +451,8 @@ private static void sanitizeMobileSearchKeys(ObjectNode searchNode, Pattern patt
       if (match.matches()) {
         String prefix = match.group("prefix");
         String code = match.group("code");
-        String channel = match.group("channel");
-        boolean codeIsValid = ALLOWED_SEARCH_CODES.contains(code);
+        String channel = match.group("channel").toLowerCase();
```
sean-rose, line 454: I'm not entirely sure it's appropriate to lowercase the channel value as the values are lowercased on the client side, and I don't think we'd necessarily want to allow through `TS` instead of `ts`.

sean-rose (changes requested):

sean-rose (approved):

PR #1998, opened 2022-02-22
Switch exception class from AffectedByBug to UnwantedDataException

quiiver:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1749356

This change will prevent these error messages from showing
up in our monitoring dashboard.

https://mozilla.cloud.looker.com/dashboards/387?Submission+Date=90+day


jklukas (approved):
Looks great!

We should make sure to add a comment to https://bugzilla.mozilla.org/show_bug.cgi?id=1626020 to document that we're making this change. Enough time has passed that we're no longer interested in monitoring the rate or potentially backfilling these.

Feel free to merge. It will deploy to stage automatically, and I expect it to be safe to let this go to prod without additional testing.

PR #2000, opened 2022-02-23
Remove bad --upgrade flag in pip install command

relud:


quiiver (approved):

sean-rose (approved):

willkg (commented):
I poked around and I think `--upgrade` wasn't having the effect I thought it was having. But I think it's good to remove it because who knows what evil lurks in future pip.

PR #2001, opened 2022-02-24
Don't have pip resolve and install dependencies

sean-rose:
The dependencies are already precalculated by pip-compile and frozen in `requirements.txt`, and pip's own dependency resolution has issues.

- The [pip docs about repeatable installs](https://pip.pypa.io/en/stable/topics/repeatable-installs/#pinning-the-package-versions) specifically suggests using `--no-deps`.
- I successfully tested the build locally using `--no-deps`.
- I'm unsure whether pip will still [install dependencies before their dependents](https://pip.pypa.io/en/stable/cli/pip_install/#installation-order) when `--no-deps` is specified.

### DIFF
--- a/ingestion-edge/bin/build
+++ b/ingestion-edge/bin/build
@@ -4,4 +4,4 @@
 
 if ${VENV:-true}; then python3.8 -m venv "$VENV_DIR"; source "$VENV_ACTIVATE"; fi
 
-python3.8 -m pip install --no-cache-dir -r requirements.txt
+python3.8 -m pip install --no-cache-dir --no-deps -r requirements.txt


sean-rose:
> pip's own dependency resolution has issues.

As a specific example, today the build started [failing in CI](https://app.circleci.com/pipelines/github/mozilla/gcp-ingestion/3948/workflows/ab1f05cb-1e80-4500-97fa-b06df0984cdf/jobs/54072) because a new googleapis-common-protos version 1.55.0 was released and pip was trying to install that even though `requirements.txt` had `googleapis-common-protos[grpc]==1.54.0` (here's the associated [Slack thread](https://mozilla.slack.com/archives/GE83ZMSAW/p1645651475996789)).

The [googleapis-common-protos version was upgraded](https://github.com/mozilla/gcp-ingestion/pull/1999) to work around the problem, but the problem will likely recur when the next googleapis-common-protos 1.x version is released because:
* google-api-core 1.31.4 specifies `googleapis-common-protos<2.0dev,>=1.6.0` as a dependency.
* grpc-google-iam-v1 0.12.3 specifies `googleapis-common-protos[grpc]<2.0.0dev,>=1.5.2` as a dependency (note the `[grpc]` extra).
* pip appears to be resolving a package with extras specified separately from the same package without extras specified (even in the latest pip 22.0.3), so when resolving google-api-core's dependency `googleapis-common-protos<2.0dev,>=1.6.0` it doesn't take into consideration the `googleapis-common-protos[grpc]==...` version pin in `requirements.txt`.

sean-rose:
It turns out the problem we're running into with pip resolving packages with extras and without extras separately is a known issue (pypa/pip#9644) and using `--no-deps` is one of the suggested workarounds.

relud (approved):

PR #2024, opened 2022-03-16
[1754044] Add support for expiration_policy.collect_through_date

quiiver:
This is support work to be able to mark schemas as deprecated via the
`expiration_policy.collect_through_date` property in the `mozPipelineMetadata` object.

* Added support for expiration_policy in the PipelineMetadataStore object.
* Created a deprecation check in the ParsePayload pipeline.

side note: my java is pretty rusty and i tried to err on the side of saftey. so please let me know if you see something that would be more idiomatic.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -9,6 +9,7 @@ import com.mozilla.telemetry.decoder.MessageScrubber.MessageScrubberException;
 import com.mozilla.telemetry.decoder.MessageScrubber.MessageShouldBeDroppedException;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
 import com.mozilla.telemetry.ingestion.core.schema.JSONSchemaStore;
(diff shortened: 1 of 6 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -168,6 +176,18 @@ private ParsePayload(String schemasLocation) {
...
 
+      if (metadataStore == null) {
+        metadataStore = PipelineMetadataStore.of(schemasLocation, BeamFileInputStream::open);
+      }
+
+      try {
+        deprecationCheck(attributes);
+      } catch (DeprecatedMessageException e) {
+        PerDocTypeCounter.inc(attributes, "error_deprecated_message");
+        PerDocTypeCounter.inc(attributes, "error_submission_bytes", submissionBytes);
```
jklukas, line 187: Seeing this here makes me realize that we aren't properly setting these metrics for messages sent to error output via `MessageScrubber.scrub` either. But that problem is not in scope for this PR.

jklukas (approved):
This looks great! The choices you've made are reasonably idiomatic for this codebase. Feel free to merge; should be a no-op going to production since no doctypes have this metadata set yet.

PR #2031, opened 2022-03-22
Bug 1737861 Tag pings from BrowserStack ISP as automation

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1757216

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -11,6 +11,9 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Streams;
 import com.mozilla.telemetry.ingestion.core.Constant.Attribute;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


relud (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -295,6 +298,20 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       processForBug1751955(json);
     }
 
+    if ("BrowserStack".equals(attributes.get(Attribute.ISP_NAME))) {
+      String tags = Strings.nullToEmpty(attributes.get(Attribute.X_SOURCE_TAGS));
+      List<String> tagList = Arrays.stream(tags.trim().split("\\s*,\\s*"))
+          .filter(s -> s.length() > 0).collect(Collectors.toList());
+      if (!tagList.contains("automation")) {
```
sean-rose, line 305: I was at first thinking this might need to be a case insensitive match (e.g. an [example in the docs](https://github.com/mozilla/gcp-ingestion/blob/c0dabae9e73e6a0a9d4ae42a24aea516b352fd98/docs/architecture/edge_service_specification.md?plain=1#L54) has "Automation" capitalized), but it looks like the [copy deduplicate logic](https://github.com/mozilla/bigquery-etl/blob/79fc36a700d9fc0704bf35a8ec7d1b93e6705345/bigquery_etl/copy_deduplicate.py#L41-L43) this is intended to trigger uses a case sensitive comparison with "automation" in lowercase, so this is consistent with that.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
```diff
@@ -566,6 +566,30 @@ public void testRedactSearchCountsForBug1751955() throws Exception {
...
+    Map<String, String> twoTags = new HashMap<>(baseAttributes);
+    twoTags.put(Attribute.X_SOURCE_TAGS, "foo ,  bar");
+    MessageScrubber.scrub(twoTags, emptyNode);
+    assertEquals("foo, bar, automation", twoTags.get(Attribute.X_SOURCE_TAGS));
+
+    Map<String, String> existingAutomationTags = new HashMap<>(baseAttributes);
+    existingAutomationTags.put(Attribute.X_SOURCE_TAGS, "automation,  foo");
+    MessageScrubber.scrub(existingAutomationTags, emptyNode);
+    assertEquals("automation,  foo", existingAutomationTags.get(Attribute.X_SOURCE_TAGS));
+  }
```
sean-rose, line 591: Maybe also a test confirming an "automation" source tag isn't added if the ISP is something other than BrowserStack?

sean-rose (approved):
`r+wc`

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
jklukas: I've now updated that example as part of this PR.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
jklukas: Added a new test case for this.

PR #2032, opened 2022-03-23
Revert "Bug 1737861 Tag pings from BrowserStack ISP as automation"

jklukas:
Reverts mozilla/gcp-ingestion#2031

I announced this change in the related incident channel, and it turns out there's not yet consensus around dropping this data, so backing out for now.

### DIFF
--- a/docs/architecture/edge_service_specification.md
+++ b/docs/architecture/edge_service_specification.md
@@ -51,7 +51,7 @@ required group attributes {
   optional string x_debug_id           // example: "my_debug_session_1"
   optional string x_pipeline_proxy     // time that the AWS->GCP tee received the message, example: "2018-03-12T21:02:18.123456Z"
   optional string x_telemetry_agent    // example: "Glean/0.40.0 (Kotlin on Android)"
-  optional string x_source_tags        // example: "Automation, Other"
+  optional string x_source_tags        // example: "automation, other"
   optional string x_foxsec_ip_reputation // example: "95"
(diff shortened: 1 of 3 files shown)


docs/architecture/edge_service_specification.md
```diff
@@ -51,7 +51,7 @@ required group attributes {
   optional string x_debug_id           // example: "my_debug_session_1"
   optional string x_pipeline_proxy     // time that the AWS->GCP tee received the message, example: "2018-03-12T21:02:18.123456Z"
   optional string x_telemetry_agent    // example: "Glean/0.40.0 (Kotlin on Android)"
-  optional string x_source_tags        // example: "automation, other"
+  optional string x_source_tags        // example: "Automation, Other"
```
sean-rose, line 54: I'd keep this part.

jklukas, line 54:
```suggestion
  optional string x_source_tags        // example: "automation, other"
```

whd:
The code this reverted was never deployed to prod. I'm going to go ahead and redeploy prod directly to the latest version.

whd:
Actually, it's worth mentioning that because the build pipeline attempts to only build new beam images when the [ingestion-beam/ingestion-core code has changed](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/beam/Jenkinsfile.build#L35), the merging of this PR caused a build to trigger which noticed that the code matched exactly a previous build and re-used that. Since that was the latest version deployed to prod, no new deploy occurred.

PR #2043, opened 2022-04-04
CONSVC-1636 Temporarily filter to just contextual-services namespace

jklukas:


jklukas:
@quiiver will need to remove this filter as part of his PR for supporting Glean topsites-impression pings.

whd (approved):

PR #2049, opened 2022-04-06
[CONSVC-1636] Support glean based telemetry in ContextualServicesReporter

quiiver:
In order to support Glean Telemetry from mobile tiles I created a new
AutoValue class called SponsoredInteraction to contain all the values
necessary for reporting the URLS to our partners.

This required a refactoring of all downstream PTransforms to use this
new class in place of PubSubMessages. I believe that this will
ultimately make it easier to understand the state of the data through
the pipeline.

Test updates are coming in the next commit.


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,108 @@
...
+    public static final String FORM_PHONE = "phone";
+    public static final String FORM_TABLET = "tablet";
+    abstract String form();
+
+    // impression, click
+    public static final String INTERACTION_IMPRESSION = "impression";
+    public static final String INTERACTION_CLICK = "click";
+    abstract String interaction();
+
+    abstract String contextID();
```
quiiver, line 34: @jklukas can context_id be nullable?

jklukas: It's [marked as required in the desktop schemas](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/2f5f9d54926eee33b7de0889d8f2c853fed948bf/schemas/contextual-services/topsites-impression/topsites-impression.1.schema.json#L74) but can't be guaranteed at the schema level for the Glean case. I would expect it to always be present.

jklukas, line 34: We've consistently captured initialisms as CamelCase in this codebase, so this should be `contextId`. Likewise, we should have `reporterUrl` and `requestId`.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,108 @@
...
+    abstract String requestID();
+
+    @Nullable
+    abstract String submissionTimestamp();
+
+    public static Builder builder() {
+        return new AutoValue_SponsoredInteraction.Builder();
+    }
+
+    public String getDocumentType() {
```
jklukas, line 49: I think it would be reasonable to drop the `get` from this name considering other members are accessed without `get`.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,108 @@
...
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.io.Serializable;
+import java.util.Map;
+
+@AutoValue
+public abstract class SponsoredInteraction implements Serializable {
+    // topsites, suggest
+    public static final String SOURCE_TOPSITE = "topsites";
+    public static final String SOURCE_SUGGEST = "suggest";
```
jklukas, line 20: The document types for desktop suggest use `quicksuggest-` as a prefix rather than `suggest`, reflecting previous proposed branding of the feature. We probably want `quicksuggest` here.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteractionCoder.java
```diff
@@ -0,0 +1,41 @@
...
+
+import org.apache.beam.sdk.coders.*;
+import org.apache.beam.sdk.values.TypeDescriptor;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.Map;
+
+public class SponsoredInteractionCoder extends CustomCoder<SponsoredInteraction> {
```
jklukas, line 11: I am hoping that we don't need to maintain this coder and can instead use AutoValueSchema like in https://github.com/GoogleCloudPlatform/dataflow-sample-applications/blob/master/retail/retail-java-applications/data-engineering-dept/data-objects/src/main/java/com/google/dataflow/sample/retail/dataobjects/ClickStream.java#L141

See https://beam.apache.org/releases/javadoc/2.19.0/org/apache/beam/sdk/schemas/AutoValueSchema.html

Beam schemas implicitly register a coder.

jklukas (commented):
Leaving some initial feedback before logging off for the day.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
quiiver: that's the GO bleeding through 👍

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteractionCoder.java
quiiver: yeah i looked at that and I don't love the custom coder I need a clearer understanding of the difference between schema and coder and how to use the schema properly

jklukas: The Beam programming guide now has some pretty comprehensive docs about schemas. In particular, see https://beam.apache.org/documentation/programming-guide/#inferring-schemas

It's my understanding that when you annotate the class with `@DefaultSchema(AutoValueSchema.class)`, part of the annotation processing is that it generates a row-based coder and adds it to Beam's coder registry. 

I believe the relevant bit where we jump from schema to coder is:

https://github.com/apache/beam/blob/7eeffe05999bb0aea44148e76dedda8482744ea9/sdks/java/core/src/main/java/org/apache/beam/sdk/schemas/SchemaCoder.java#L113-L120

So you should never have to explicitly specify a coder for these PCollections.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,108 @@
...
+
+        public abstract Builder form(String form);
+
+        public abstract Builder interaction(String interaction);
+
+        public abstract Builder submissionTimestamp(String submissionTimestamp);
+
+        public abstract SponsoredInteraction build();
+
+        public Builder from(SponsoredInteraction interaction) {
```
jklukas, line 96: You should instead to able to provide an abstract `toBuilder` method on `SponsoredInteraction` and AutoValue will be able to fill it in.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,108 @@
...
+
+    // impression, click
+    public static final String INTERACTION_IMPRESSION = "impression";
+    public static final String INTERACTION_CLICK = "click";
+    abstract String interaction();
+
+    abstract String contextID();
+
+    @Nullable
+    abstract String reporterURL();
```
jklukas, line 37: I think this should be `reportingUrl` (with -ing) to be consistent with how it's named in the payload.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -146,15 +176,18 @@ private ParseReportingUrl(String urlAllowList) {
...
                 message.getAttribute(Attribute.GEO_DMA_CODE));
           }
 
           // We only add these dimensions for topsites clicks, not quicksuggest per
+          // We are also limiting this to desktop.
           // https://bugzilla.mozilla.org/show_bug.cgi?id=1738974
-          if (message.getAttribute(Attribute.DOCUMENT_TYPE).equals("topsites-click")) {
+          if (interaction.form().equals(SponsoredInteraction.FORM_DESKTOP)
+             && interaction.source().equals(SponsoredInteraction.SOURCE_TOPSITE)
+             && interaction.interaction().equals(SponsoredInteraction.INTERACTION_CLICK)) {
```
jklukas, line 190:
```suggestion
          if (SponsoredInteraction.FORM_DESKTOP.equals(interaction.form())
             && SponsoredInteraction.SOURCE_TOPSITE.equals(interaction.source())
             && SponsoredInteraction.INTERACTION_CLICK.equals(interaction.interaction())) {
```

To avoid potential NPE, it's safer to call `STRING_CONSTANT.equals(value)` than the other way around.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -87,40 +93,65 @@ private ParseReportingUrl(String urlAllowList) {
           }
 
           Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
+          String documentNS = message.getAttribute(Attribute.DOCUMENT_NAMESPACE);
+          String documentType = message.getAttribute(Attribute.DOCUMENT_TYPE);
```
jklukas, line 97: Style in this codebase has typically been to call these `namespace` and `docType`. If those might be ambiguous, then we should go fully formal with `documentNamespace` and `documentType`.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -34,13 +36,17 @@
...
+    PTransform<PCollection<PubsubMessage>, Result<PCollection<SponsoredInteraction>, PubsubMessage>> {
 
   private final String urlAllowList;
 
   private static transient Set<String> singletonAllowedImpressionUrls;
   private static transient Set<String> singletonAllowedClickUrls;
 
+
+  private static final String NS_DESKTOP = "contextual-services";
+  private static final String NS_FENIX = "org.mozilla.fenix";
```
jklukas, line 48: Note that the appIds use dots, but the corresponding namespaces use hyphens. So the expected namespaces here as they appear in PubSub attributes should be `org-mozilla-fenix`, `org-mozilla-firefox-beta`, and `org-mozilla-firefox`.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -113,11 +112,12 @@ public void process(@Element KV<String, PubsubMessage> element, @Timestamp Insta
...
-        PubsubMessage message = element.getValue();
-        Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
-        addClickStatusToReportingUrlAttribute(attributes);
+        SponsoredInteraction interaction = element.getValue();
+        String reportingURL = addClickStatusToReportingUrlAttribute(interaction.reporterURL());
         ghostClickCounter.inc();
-        out.output(KV.of(element.getKey(), new PubsubMessage(message.getPayload(), attributes)));
+        out.output(KV.of(
+                element.getKey(),
+                SponsoredInteraction.builder().from(interaction).reporterURL(reportingURL).build()));
```
jklukas, line 120: Per other comment, this can become `interaction.toBuilder.reporterURL(reportingURL).build()`

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -46,11 +46,11 @@
...
    * Composite transform that wraps {@code DetectClickSpikes} with keying by {@code context_id}.
    */
-  public static PTransform<PCollection<PubsubMessage>, PCollection<PubsubMessage>> perContextId(
+  public static PTransform<PCollection<SponsoredInteraction>, PCollection<SponsoredInteraction>> perContextId(
       Integer maxClicks, Duration windowDuration) {
     return PTransform.compose("DetectClickSpikesPerContextId", input -> input //
-        .apply(WithKeys.of((message) -> message.getAttribute(Attribute.CONTEXT_ID))) //
-        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of()))
+        .apply(WithKeys.of((interaction) -> interaction.contextID())) //
+        .setCoder(KvCoder.of(StringUtf8Coder.of(), SponsoredInteractionCoder.of()))
```
jklukas, line 53: Assuming we get AutoValueSchema working, then I think we'd look up the coder here using the registry per https://beam.apache.org/documentation/programming-guide/#default-coder-lookup

jklukas (commented):
Another set of comments after looking over this more thoroughly this morning.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
quiiver: yeah somehow i missed that in my first read through the autovalue docs 😬

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
quiiver: I am removing the FENIX namespace here as its not currently used.

checkstyle/checks.xml
```diff
@@ -50,7 +50,7 @@
 
     <module name="LineLength">
         <property name="fileExtensions" value="java"/>
-        <property name="max" value="100"/>
+        <property name="max" value="120"/>
```
relud, line 53: this change should be a separate PR, and it would probably be best to discuss it first in the weekly Data Platform Infrastructure Working Group meeting.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -176,34 +224,30 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
+              | ParsedReportingUrl.InvalidUrlException | RejectedMessageException e) {
+            return FailureMessage.of(ParseReportingUrl.class.getSimpleName(), ee.element(),
+                ee.exception());
+          }
+        }));
   }
 
   @VisibleForTesting
-  boolean isUrlValid(URL url, String doctype) {
+  boolean isUrlValid(URL url, String interaction) {
```
relud, line 242: this was a bit confusing to see `interaction` be a `String`, and based on how this is called I would expect this parameter to be `interactionType`

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -146,15 +190,19 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
+            urlParser.addQueryParam(ParsedReportingUrl.PARAM_FORM_FACTOR,
+                interaction.getFormFactor());
 
             urlParser.addQueryParam(ParsedReportingUrl.PARAM_DMA_CODE,
                 message.getAttribute(Attribute.GEO_DMA_CODE));
           }
 
           // We only add these dimensions for topsites clicks, not quicksuggest per
+          // We are also limiting this to desktop.
           // https://bugzilla.mozilla.org/show_bug.cgi?id=1738974
```
relud, line 202: additional comment should probably be added at the end, instead of in the middle:
```suggestion
          // We only add these dimensions for topsites clicks, not quicksuggest per
          // https://bugzilla.mozilla.org/show_bug.cgi?id=1738974
          // We are also limiting this to desktop.
```

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
```diff
@@ -1,45 +1,56 @@
...
 
+  private SponsoredInteraction.Builder getTestInteraction() {
+    return SponsoredInteraction.builder().setInteractionType("click").setSource("topsite")
+        .setFormFactor("phone").setContextId("1");
+  }
+
   @Rule
   public TestPipeline pipeline = TestPipeline.create();
 
+  private SchemaCoder<SponsoredInteraction> getCoder() {
```
relud, line 30: why not re-use the coder definition from `AggregateImpressions.java`?

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/ParseReportingUrlTest.java
```diff
@@ -34,11 +36,19 @@ public class ParseReportingUrlTest {
...
+      throw new RuntimeException(e.getMessage());
+    }
+  }
+
   @Test
   public void testAllowedUrlsLoadAndFilter() throws IOException {
     ParseReportingUrl parseReportingUrl = ParseReportingUrl.of(URL_ALLOW_LIST);
 
-    pipeline.run();
+    // pipeline.run();
```
relud, line 51:
```suggestion
    pipeline.run();
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -87,40 +94,78 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
+            JsonNode event = events.get(0);
+            String eventName = event.path("name").asText();
+            interactionBuilder.setInteractionType(
+                eventName.contains("click") ? SponsoredInteraction.INTERACTION_CLICK
+                    : SponsoredInteraction.INTERACTION_IMPRESSION);
+          }
+
+          interactionBuilder.setSource(docType.contains(SponsoredInteraction.SOURCE_TOPSITES)
+              ? SponsoredInteraction.SOURCE_TOPSITES
+              : SponsoredInteraction.SOURCE_SUGGEST);
```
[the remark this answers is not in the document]
quiiver: I'm curious about this. If you don't mind expanding on what you mean in this comment.

In a previous PTransform we filter messages to only include docTypes of `topsites-impression` `topsites-click` `quicksuggest-impression` and `quicksuggest-click` and namespaces of `contextual-services` (desktop) and `org-mozilla-fenix` (phone). There will likely be additional namespaces (ios) but we're not currently supporting that and the form factor there will still be `phone`

I can move the ternary expressions to explicit conditionals if you feel like that would be a safer option.

checkstyle/checks.xml
quiiver: Can do!

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -176,34 +224,30 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
 
-    if (doctype.endsWith("-click")) {
+    if (interaction.equals("click")) {
       allowedUrls = singletonAllowedClickUrls;
-    } else if (doctype.endsWith("-impression")) {
+    } else if (interaction.equals("impression")) {
       allowedUrls = singletonAllowedImpressionUrls;
     } else {
-      throw new IllegalArgumentException("Invalid doctype: " + doctype);
+      throw new IllegalArgumentException("Invalid doctype: " + interaction);
```
relud, line 250:
```suggestion
      throw new IllegalArgumentException("Invalid interaction: " + interaction);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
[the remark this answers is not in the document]
relud: yeah, to explain by example, I'd expect something like:
```suggestion
          String[] docTypeParts = docType.split("-", 2);

          if (docTypeParts.length != 2) {
            throw new IllegalArgumentException("Invalid docType: " + docType);
          }

          String source = docTypeParts[0];
          String interactionType = docTypeParts[1];

          if (!SponsoredInteraction.SOURCE_TOPSITES.equals(source)
              && !SponsoredInteraction.SOURCE_SUGGEST.equals(source) {
            // Unclear if this is necessary
            throw new IllegalArgumentException("Invalid source: " + source);
          }
          // interactionType is checked in isUrlValid

          interactionBuilder.setSource(source);
          interactionBuilder.setInteractionType(interactionType);

          if (NS_DESKTOP.equals(namespace)) {
            interactionBuilder.setFormFactor(SponsoredInteraction.FORM_DESKTOP);
          } else {
            // All other namespaces are for mobile
            interactionBuilder.setFormFactor(SponsoredInteraction.FORM_PHONE);

            // Unclear if this is necessary
            ArrayNode events = payload.withArray("events");
            if (events.size() != 1) {
              throw new RuntimeException("expect exactly 1 event in ping.");
            }
            JsonNode event = events.get(0);
            String eventName = event.path("name").asText();
            if (!eventName.endsWith("-" + interactionType)) {
              throw new IllegalArgumentException("Invalid event name '" + eventName
                  + "' for interaction type '" + interactionType + "'");
            }
          }
```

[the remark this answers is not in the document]
relud: > In a previous PTransform we filter messages to only include docTypes of `topsites-impression` `topsites-click` `quicksuggest-impression` and `quicksuggest-click` and namespaces of `contextual-services` (desktop) and `org-mozilla-fenix` (phone). There will likely be additional namespaces (ios) but we're not currently supporting that and the form factor there will still be `phone`

If i understand correctly those filters are in configuration, rather than in code. If configuration changed this should handle that gracefully. For source and interaction type, we should use the content of docType split on `-`, rather than assume source and interaction will never be a substring of the other. It sounds like the desktop or phone choice is fine as is, because even if configuration changes, everything else is FORM_PHONE, so it's an accurate assumption.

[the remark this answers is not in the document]
relud: > I can move the ternary expressions to explicit conditionals if you feel like that would be a safer option.

It's not that the ternary is unsafe, it's more that I think the code should probably account for more than two outcomes. If new doc types are added later, then behavior like `if a {} else if b {} else {throw exception}` will prevent hard to find issues where a new option for source or interaction type gets hidden.

If checking the payload event name isn't needed then a ternary operator for form factor would make sense.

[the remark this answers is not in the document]
quiiver: > If new doc types are added later, then behavior like if a {} else if b {} else {throw exception} will prevent hard to find issues where a new option for source or interaction type gets hidden.

I agree with that and will address it.

> If checking the payload event name isn't needed then a ternary operator for form factor would make sense.
Unfortunately we only get one docType for mobile tiles (`topsites-impression`) and the interaction type is encoded in the event name. I'll add some comments around that so its clearer for future me/us.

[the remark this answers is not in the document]
relud: > i don't think contains is strict enough here

For example, if we later add a source like `siteclick` and a interaction type `click-impression`, then with `docType = "siteclick-click-impression"` we would get incorrect behavior from using `contains("click")` or `endsWith("-impression")`. If we want to use contains, then we have to enforce that none of these values are substrings of eachother.

So this could use `startsWith` and `endsWith` and prefix/suffix the match with `-` and enforce a single `-` in `docType`, or it could use a regex or `String.split` to ensure we don't do any partial matching.

> Unfortunately we only get one docType for mobile tiles (topsites-impression) and the interaction type is encoded in the event name. I'll add some comments around that so its clearer for future me/us.

we should also throw an exception when docType doesn't match that expectation

checkstyle/checks.xml
quiiver: reverted

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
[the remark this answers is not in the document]
quiiver: I refactored this to use explicit `if {} else if {} else {}` and throwing an exception when the values don't match any expected value

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/LabelClickSpikesTest.java
quiiver: I moved the method to a static on `SponsoredInteraction`

quiiver:
I have resolved all codestyle errors except for the following:
```[ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/LabelClickSpikes.java:33: Line is longer than 100 characters (found 110). [LineLength]
[ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/LabelClickSpikes.java:42: Line is longer than 100 characters (found 110). [LineLength]
[ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/ParseReportingUrl.java:40: Line is longer than 100 characters (found 102). [LineLength]
[ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/SendRequest.java:26: Line is longer than 100 characters (found 109). [LineLength]
```

In each of these classes the type declaration of the `PTransform` is longer than 100 chars. Spotless will not allow any wrapping and checkstyle will not allow any lines longer than 100 chars.

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/ParseReportingUrlTest.java
```diff
@@ -165,7 +167,7 @@ public void testRequiredParamCheck() {
       return null;
     });
 
-    PAssert.that(result.output()).satisfies(messages -> {
+    PAssert.that(result.output().setCoder(getCoder())).satisfies(messages -> {
```
relud, line 170: why not use `SponsoredInteraction.getCoder()`?

relud:
> I have resolved all codestyle errors except for the following:
> 
> ```
> [ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/LabelClickSpikes.java:42: Line is longer than 100 characters (found 110). [LineLength]
> [ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/ParseReportingUrl.java:40: Line is longer than 100 characters (found 102). [LineLength]
> [ERROR] /root/project/ingestion-beam/src/main/java/com/********telemetry/contextualservices/SendRequest.java:26: Line is longer than 100 characters (found 109). [LineLength]
> ```
> 
> In each of these classes the type declaration of the `PTransform` is longer than 100 chars. Spotless will not allow any wrapping and checkstyle will not allow any lines longer than 100 chars.

individual violations can be suppressed with `@SuppressWarnings("checkstyle:methodlength")`, for example: https://github.com/mozilla/gcp-ingestion/blob/2aab6e773546ee9cd7d1eb7c7d502e2d6fb42cb5/ingestion-beam/src/test/java/com/mozilla/telemetry/util/TestWithDeterministicJson.java#L26-L27

ingestion-beam/src/test/java/com/mozilla/telemetry/contextualservices/ParseReportingUrlTest.java
quiiver: I thought I had found all the instances of that. 🤦  thank you!

relud (approved):
this looks good to me, but I'm not sure if we should wait for @jklukas to r+ before merging

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -0,0 +1,119 @@
+package com.mozilla.telemetry.contextualservices;
+
+import com.google.auto.value.AutoValue;
+import com.mozilla.telemetry.ingestion.core.schema.SchemaNotFoundException;
```
jklukas, line 4: We should use a different exception type here. If Beam has a built-in exception for this, we should consider that.

ingestion-beam/src/main/java/com/mozilla/telemetry/ContextualServicesReporter.java
```diff
@@ -60,14 +60,13 @@ public static PipelineResult run(ContextualServicesReporterOptions.Parsed option
...
+    PCollection<SponsoredInteraction> requests = pipeline //
         .apply(options.getInputType().read(options)) //
-        .apply(FilterByDocType.of(options.getAllowedDocTypes())) //
+        .apply(FilterByDocType.of(options.getAllowedDocTypes(), options.getAllowedNamespaces())) //
         .apply(VerifyMetadata.of()) //
         .failuresTo(errorCollections) //
         .apply(DecompressPayload.enabled(options.getDecompressInputPayloads())) //
-        .apply(ParseReportingUrl.of(options.getUrlAllowList())) //
-        .failuresTo(errorCollections) //
+        .apply(ParseReportingUrl.of(options.getUrlAllowList())).failuresTo(errorCollections) //
```
jklukas, line 69: Should probably revert this whitespace change.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/AggregateImpressions.java
```diff
@@ -45,28 +39,27 @@ public static AggregateImpressions of(String getAggregationWindowSize) {
...
 
   @Override
-  public PCollection<PubsubMessage> expand(PCollection<PubsubMessage> messages) {
-    return messages //
-        // Add reporting url as key, pubsub message as value
-        .apply(WithKeys.of((SerializableFunction<PubsubMessage, String>) //
-        AggregateImpressions::getAggregationKey)) //
-        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of())) //
+  public PCollection<SponsoredInteraction> expand(PCollection<SponsoredInteraction> messages) {
+    SerializableFunction<SponsoredInteraction, String> fn = AggregateImpressions::getAggregationKey;
```
jklukas, line 43: Improve name of fn here.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/EmitCounters.java
```diff
@@ -24,17 +23,19 @@ private EmitCounters() {
...
-        MapElements.into(TypeDescriptor.of(PubsubMessage.class)).via((PubsubMessage message) -> {
-          message = PubsubConstraints.ensureNonNull(message);
-          Map<String, String> attributes = message.getAttributeMap();
+  public PCollection<SponsoredInteraction> expand(PCollection<SponsoredInteraction> messages) {
+    return messages.apply(MapElements.into(TypeDescriptor.of(SponsoredInteraction.class))
+        .via((SponsoredInteraction interaction) -> {
+          // mock a doctype based on the sponsored interaction fields
+          Map<String, String> attributes = Map.of(Attribute.DOCUMENT_TYPE,
+              interaction.getDocumentType(), Attribute.DOCUMENT_NAMESPACE,
+              Optional.ofNullable(interaction.getOriginalNamespace()).orElse("unknown_namespace"));
```
jklukas, line 32: Let's avoid putting this constant here. Let's build the map incrementally.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -100,9 +93,9 @@ private class Fn extends DoFn<KV<String, PubsubMessage>, KV<String, PubsubMessag
     private final TimerSpec clickTimer = TimerSpecs.timer(TimeDomain.PROCESSING_TIME);
 
     @ProcessElement
-    public void process(@Element KV<String, PubsubMessage> element, @Timestamp Instant elementTs,
-        @StateId("click-state") ValueState<List<Long>> state, @TimerId("click-timer") Timer timer,
-        OutputReceiver<KV<String, PubsubMessage>> out) {
+    public void process(@Element KV<String, SponsoredInteraction> element,
+        @Timestamp Instant elementTs, @StateId("click-state") ValueState<List<Long>> state,
+        @TimerId("click-timer") Timer timer, OutputReceiver<KV<String, SponsoredInteraction>> out) {
```
jklukas, line 98: May want to use trailing `//` and split out each argument to its own line

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -87,40 +99,104 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
+          } else {
+            interactionBuilder.setFormFactor(SponsoredInteraction.FORM_PHONE);
+            // enforce that the only mobile docType is `topsites-impression`
+            if (!DT_TOPSITES_IMPRESSION.equals(docType)) {
+              throw new InvalidAttributeException("Unexpected docType for mobile ping: " + docType,
+                  docType);
+            }
+            ArrayNode events = payload.withArray("events");
+            if (events.size() != 1) {
+              throw new RuntimeException("expect exactly 1 event in ping.");
```
jklukas, line 135: `InvalidAttributeException` might be okay here. We should make sure we use something that is caught and sent to error output rather than letting this bubble up and take down the pipeline.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -176,34 +255,30 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
...
-              } catch (UncheckedIOException | IllegalArgumentException
-                  | ParsedReportingUrl.InvalidUrlException | RejectedMessageException e) {
-                return FailureMessage.of(ParseReportingUrl.class.getSimpleName(), ee.element(),
-                    ee.exception());
-              }
-            }));
+        .exceptionsVia((ExceptionElement<PubsubMessage> ee) -> {
+          try {
+            throw ee.exception();
+          } catch (UncheckedIOException | IllegalArgumentException
```
quiiver, line 264: Make sure we catch `InvalidAttributeException` here

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -276,6 +353,13 @@ private List<String[]> readPairsFromFile(String fileLocation, String paramName)
     }
   }
 
+  private String extractReportingUrl(ObjectNode payload) {
+    if (!payload.path(Attribute.REPORTING_URL).isMissingNode()) {
+      return payload.path(Attribute.REPORTING_URL).asText();
+    }
+    return payload.path("metrics").path("url2").path("top_sites_contile_reporting_url").asText();
```
jklukas, line 360:
```suggestion
    return payload.path("metrics").path("url").path("top_sites_contile_reporting_url").asText();
```

This should be `url` at this point in the pipeline. It gets translated to url2 at the BigQuery write step.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/FailureMessage.java
```diff
@@ -35,6 +36,15 @@ public static PubsubMessage of(Object caller, String payload, Throwable e) {
     return FailureMessage.of(caller, payload.getBytes(StandardCharsets.UTF_8), e);
   }
 
+  /**
+   * Return a PubsubMessage wrapping a SponsoredInteraction with attributes describing the error.
+   */
+  public static PubsubMessage of(Object caller, SponsoredInteraction interaction, Throwable e) {
```
jklukas, line 42: In terms of modules, it's not great that we're importing from `contextualservices` in this higher-level, but don't have a better suggestion for this.

jklukas: No action needed on this. Might be worth calling as a todo in the comments.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -1,19 +1,12 @@
...
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
-import org.apache.beam.sdk.coders.KvCoder;
-import org.apache.beam.sdk.coders.StringUtf8Coder;
-import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
-import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessageWithAttributesCoder;
 import org.apache.beam.sdk.metrics.Counter;
```
jklukas, line 10:
```suggestion
import org.apache.beam.sdk.coders.KvCoder;
import org.apache.beam.sdk.coders.StringUtf8Coder;
import org.apache.beam.sdk.metrics.Counter;
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/LabelClickSpikes.java
```diff
@@ -46,11 +40,10 @@ public class LabelClickSpikes extends
   /**
    * Composite transform that wraps {@code DetectClickSpikes} with keying by {@code context_id}.
    */
-  public static PTransform<PCollection<PubsubMessage>, PCollection<PubsubMessage>> perContextId(
+  public static PTransform<PCollection<SponsoredInteraction>, PCollection<SponsoredInteraction>> perContextId(
       Integer maxClicks, Duration windowDuration) {
     return PTransform.compose("DetectClickSpikesPerContextId", input -> input //
-        .apply(WithKeys.of((message) -> message.getAttribute(Attribute.CONTEXT_ID))) //
-        .setCoder(KvCoder.of(StringUtf8Coder.of(), PubsubMessageWithAttributesCoder.of()))
+        .apply(WithKeys.of((interaction) -> interaction.getContextId())) //
```
jklukas, line 46:
```suggestion
        .apply(WithKeys.of((interaction) -> interaction.getContextId())) //
        .setCoder(KvCoder.of(StringUtf8Coder.of(), SponsoredInteraction.getCoder())) //
```

When attempting to deploy to Dataflow, I got a coder error here.

jklukas (changes requested):
Once change needed to get this deployable.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/VerifyMetadata.java
```diff
@@ -69,7 +69,7 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
                 String.format("Invalid Firefox version: %s", version), "user_agent_version");
           }
 
-          // Verify release channel
+          // Verify release channel for specific doctype
```
jklukas, line 72: All Fenix pings are getting rejected in VerifyMetadata with "Invalid user agent: null" so we'll need more refactoring here to make sure the desktop-specific checks are only applied if document namespace is "contextual-services".

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/InvalidAttributeException.java
```diff
@@ -0,0 +1,15 @@
...
+
+public class InvalidAttributeException extends RuntimeException {
+
+  InvalidAttributeException(String message) {
+    super(message);
+  }
+
+  InvalidAttributeException(String message, String reason) {
+    super(message);
+    KeyedCounter.inc("invalid_attribute" + reason);
```
jklukas, line 13:
```suggestion
    KeyedCounter.inc("invalid_attribute_" + reason);
```

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -87,40 +99,104 @@ public Result<PCollection<PubsubMessage>, PubsubMessage> expand(
           }
 
           Map<String, String> attributes = new HashMap<>(message.getAttributeMap());
+          String namespace = Objects
+              .requireNonNull(message.getAttribute(Attribute.DOCUMENT_NAMESPACE));
```
jklukas, line 103: This will throw `NullPointerException` which isn't currently caught, so will cause the whole pipeline to fail. We may want to throw an `IllegalAttributeException` instead.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -276,6 +354,13 @@ private List<String[]> readPairsFromFile(String fileLocation, String paramName)
     }
   }
 
+  private String extractReportingUrl(ObjectNode payload) {
+    if (!payload.path(Attribute.REPORTING_URL).isMissingNode()) {
+      return payload.path(Attribute.REPORTING_URL).asText();
+    }
+    return payload.path("metrics").path("url").path("top_sites_contile_reporting_url").asText();
```
jklukas, line 361:
```suggestion
    return payload.path("metrics").path("url").path("top_sites.contile_reporting_url").asText();
```

The path in the JSON contains a `.` because these are constructed as `<metric_group>.<metric_name>`, but this gets normalized to an underscore for the BQ field name.

jklukas:
I've documented here all the issues I found when I deployed this to the contextual-services-dev project. Getting those fixes in place should be sufficient for this to be mergeable.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/SponsoredInteraction.java
```diff
@@ -66,13 +66,17 @@ public static Builder builder() {
    * it will build one based on `source` and `interactionType` fields.
    * @return String
    */
-  public String getDocumentType() {
-    if (getOriginalDocType() != null) {
+  public String getDocumentType(boolean derived) {
```
jklukas, line 69: The boolean here makes it very hard to understand intent when this method gets called. I'd prefer a separate method called `getDerivedDocumentType`. And maybe get rid of `getDocumentType()` and instead rely on explicitly calling `getOriginalDocumentType`

quiiver: yeah that's better. good call

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/FailureMessage.java
```diff
@@ -35,6 +36,16 @@ public static PubsubMessage of(Object caller, String payload, Throwable e) {
     return FailureMessage.of(caller, payload.getBytes(StandardCharsets.UTF_8), e);
   }
 
+  /**
+   * Return a PubsubMessage wrapping a SponsoredInteraction with attributes describing the error.
+   * TODO Investigate refactoring to not import SponsoredInteraction into this higher-level module.
+   */
+  public static PubsubMessage of(Object caller, SponsoredInteraction interaction, Throwable e) {
+    PubsubMessage message = new PubsubMessage(
+        interaction.toString().getBytes(StandardCharsets.UTF_8), Map.of());
```
jklukas, line 45: We send in an empty map of attributes here, thus almost all of the fields of the error message in BQ are null. In particular, `submission_timestamp` is null, which explains why we aren't seeing these messages appearing in BQ.

jklukas: We should at minimum hydrate the attributes map with submission_timestamp, document_type, and document_version. It would be good to include geo as well.

jklukas (commented):
Changes from the last few days all look good here. Fixing up the FailureMessage is the last blocking change needed here.

ingestion-beam/src/main/java/com/mozilla/telemetry/transforms/FailureMessage.java
```diff
@@ -35,6 +40,29 @@ public static PubsubMessage of(Object caller, String payload, Throwable e) {
...
+        Optional.ofNullable(interaction.getOriginalNamespace()).orElse("missing_namespace"));
+    attributes.put(Attribute.DOCUMENT_TYPE,
+        Optional.ofNullable(interaction.getOriginalDocType()).orElse("missing_doctype"));
+    attributes.put(Attribute.SUBMISSION_TIMESTAMP, Optional //
+        .ofNullable(interaction.getSubmissionTimestamp()) //
+        .orElseGet(() -> Time.epochMicrosToTimestamp(new Instant().getMillis() * 1000)));
+    Map<String, String> additionalAttributes = interaction.getAdditionalAttributes();
+    if (additionalAttributes != null) {
+      attributes.putAll(interaction.getAdditionalAttributes());
+    }
```
jklukas, line 59: I'd recommend putting additionalAttributes into the map first, so that the explicit fields of `SponsoredInteraction` will always take precedence if we happen to duplicate a field.

PR #2054, opened 2022-04-19
Bug 1765250 Avoid unroutable messages with missing document_id

jklukas:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1765250

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
@@ -49,6 +49,9 @@ public class ParseUriTest extends TestWithDeterministicJson {
             + "{\"uri\":\"/nonexistent_prefix/ce39b608-f595-4c69-b6a6-f7a436604648" //
             + "/main/Firefox/61.0a1/nightly/20180328030202\"" //
             + "},\"payload\":\"\"}",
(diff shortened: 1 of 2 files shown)


whd (approved):
This is... not how I expected the PR to look, but makes sense.

It would be interesting to do analysis on how often document ids include bad values; I in fact remember doing some analysis like this but I can't remember the context.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
```diff
@@ -49,6 +49,9 @@ public void testOutput() {
             + "{\"uri\":\"/nonexistent_prefix/ce39b608-f595-4c69-b6a6-f7a436604648" //
             + "/main/Firefox/61.0a1/nightly/20180328030202\"" //
             + "},\"payload\":\"\"}",
+        "{\"attributeMap\":{\"uri\":\"/submit/telemetry//update/Firefox/61.0a1" //
```
whd, line 52: This doesn't match the format I expect for a telemetry URI in that it doesn't appear to have a doc id. It should still fail though with the same error.

jklukas: That's exactly how the failing documents in bug 1765250 are structured. They are valid telemetry URIs except that the docId space is empty. The existing code would already reject items that have too many path elements or which don't have valid namespace/doctype/version to look up a schmea.

PR #2063, opened 2022-04-26
Bug 1766424 - Allow empty user agent string from Glean on Desktop

akkomar:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -234,10 +234,11 @@ public static void scrub(Map<String, String> attributes, ObjectNode json)
       throw new UnwantedDataException("1592010");
     }
 
-    // Glean enforces a particular user-agent string that a rogue fuzzer is not abiding by
+    // Up to the v0.13 Glean enforces a particular user-agent string that a rogue fuzzer is not abiding by
     // https://searchfox.org/mozilla-central/source/third_party/rust/glean-core/src/upload/request.rs#35,72-75
+    // Glean stopped submitting user-agent after v0.13, therefore here we also accept an empty one
```
badboy, line 239:
```suggestion
    // The Glean SDK stopped submitting an user-agent after v44.0.0, therefore here we also accept an empty one
```

badboy (approved):
small comment adjustement.

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/MessageScrubberTest.java
```diff
@@ -704,13 +704,12 @@ public void testUnwantedData1684980GleanUserAgent() {
     assertThrows(UnwantedDataException.class,
         () -> MessageScrubber.scrub(emptyAttributes, Json.createObjectNode()));
```
akkomar, line 705: We'll still discard pings with an empty user agent string

badboy: Ah, empty string. Not a missing string.

akkomar: Yeah, flagging this just in case, it looks to me that it should be discarded.

PR #2064, opened 2022-04-26
Bug 1766424 Allow new-style Glean header for desktop

jklukas:
Fixup to #2063; it turns out the newer Glean version pass on the browser user agent value rather than having that value be null.


jklukas:
cc @whd

relud (approved):

akkomar (approved):
LGTM, holding off merging until @badboy r+'s.

PR #2068, opened 2022-04-28
CONSVC-1764 Add heuristic to handle a specific fraud signature

jklukas:
See https://mozilla-hub.atlassian.net/browse/CONSVC-1764


jklukas:
cc @ncloudioj

ncloudioj (approved):
Thanks!

PR #2075, opened 2022-05-06
CONSVC-1777 Require "online" scenario for sending suggest pings

jklukas:
See https://mozilla-hub.atlassian.net/browse/CONSVC-1777


quiiver (approved):
lgtm

PR #2089, opened 2022-05-23
Fix doc string errors

relud:
also use `@throws` rather than `@exception` for more readable grammar and consistency (`Json` is currently using a mix of both)

Not sure why CI allows these when my local dev environment throws compiler errors like `invalid use of @throws`, but it seems to only crop up with Oracle Java and `<compilerArgument>-Xdoclint:all</compilerArgument>` and an annotation processor enabled (AutoValue).


ingestion-sink/src/main/java/com/mozilla/telemetry/ingestion/sink/transform/PubsubMessageToTemplatedString.java
```diff
@@ -20,7 +20,7 @@
  * <p>All message attributes, plus derived attributes provided by {@link DerivedAttributesMap},
  * are available for substitution.
  *
- * @throws IllegalArgumentException if template is not fully resolved for a given message.
+ * <p>Throws IllegalArgumentException if template is not fully resolved for a given message.
```
jklukas, line 23: Why the change here?

relud: because `@throws` isn't a valid tag on a class (even if that class inherits from `Function`)

PR #2100, opened 2022-06-09
Accept "top_site" metric category for iOS impressions

jklukas:
Currently, all iOS impressions are being rejected.


jklukas:
I've marked the test here as `@Ignore` since it currently fails due to user agent parsing (user agent metadata is all null for iOS). I'm going to merge this, but we'll still have to address the user agent issues before this starts flowing to partner.

PR #2128, opened 2022-07-15
Bug 1779801 - mark pings sent from rally-markup-fb-pixel-hunt as Unwa…

rhelmer:
…ntedData

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
@@ -89,6 +89,7 @@ public class MessageScrubber {
       .put("net-chinaedu-crystal", "1767076") //
       .put("com-you-browser", "1767077") //
       .put("org-giveasyoulive-donationreminder", "1767078") //
+      .put("rally-markup-fb-pixel-hunt", "1779801") //
       .build();
 
   private static final Map<String, String> IGNORED_TELEMETRY_DOCTYPES = ImmutableMap


quiiver (approved):
This looks good to me. Do we also need to remove it from the probe-scraper?

rhelmer:
> This looks good to me. Do we also need to remove it from the probe-scraper?

Yes I think we might as well. I should note that there is a possibility that we may re-launch this study, are these changes safely reversible if we do that?

quiiver:
> are these changes safely reversible if we do that?

Yes these changes here are reversible but we would want to make sure we revert before re-launching the study.

rhelmer:
> > are these changes safely reversible if we do that?
> 
> Yes the changes here are reversible but we would want to make sure we revert before re-launching the study.

OK thanks @quiiver ! Yes could we land this one and then also do the probe scraper? Thanks!

whd:
We implemented `collect_through_date` metadata in https://bugzilla.mozilla.org/show_bug.cgi?id=1748938 for this particular class of Unwanted Data. See also https://bugzilla.mozilla.org/show_bug.cgi?id=1777555 for a case where we leveraged that metadata, and then re-enabled data collection by modifying the metadata. 

Note that we've never done this with a glean app but we did implement support for it in https://bugzilla.mozilla.org/show_bug.cgi?id=1763467. I think this would be a good opportunity to test out custom metadata in a glean app via the newly developed facility.

> Do we also need to remove it from the probe-scraper?

I don't think we should do this unless we're deleting the data. There's a `deprecated: true` flag e.g. https://github.com/mozilla/probe-scraper/blob/main/repositories.yaml#L552 that we might consider setting, but if we remove the entry, I think that will cause the schema generation pipeline to fail. There are plans to change this in https://docs.google.com/document/d/1lwjCCY7mDri40tNMNMxXRaCaxaKIcOwSUJuNAKDCBmg/edit# but they haven't been fully implemented yet.

rhelmer:
> We implemented `collect_through_date` metadata in https://bugzilla.mozilla.org/show_bug.cgi?id=1748938 for this particular class of Unwanted Data. See also https://bugzilla.mozilla.org/show_bug.cgi?id=1777555 for a case where we leveraged that metadata, and then re-enabled data collection by modifying the metadata.
> 
> Note that we've never done this with a glean app but we did implement support for it in https://bugzilla.mozilla.org/show_bug.cgi?id=1763467. I think this would be a good opportunity to test out custom metadata in a glean app via the newly developed facility.

OK thanks, this does seem useful for Rally since our data collection usually (but not always) has an end date. @quiiver would you like me to do this instead or is that something you should handle?

rhelmer:
Opened https://github.com/mozilla/probe-scraper/pull/463

PR #2138, opened 2022-08-17
RS-430 Add Bing partner codes for ESR and iPad

akkomar:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -135,7 +135,7 @@ public class MessageScrubber {
       // https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/search-telemetry-v2/records
       // Bing
       "MOZ2", "MOZ4", "MOZ5", "MOZA", "MOZB", "MOZD", "MOZE", "MOZI", "MOZM", "MOZO", "MOZT",
-      "MOZW", "MOZSL01", "MOZSL02", "MOZSL03",
+      "MOZW", "MOZSL01", "MOZSL02", "MOZSL03", "MOZL", "MOZR",
```
sean-rose, line 138: It's best to keep these in sorted order within each group (easier to scan and see if a code is already there):
```suggestion
      "MOZ2", "MOZ4", "MOZ5", "MOZA", "MOZB", "MOZD", "MOZE", "MOZI", "MOZL", "MOZM", "MOZO",
      "MOZR", "MOZT", "MOZW", "MOZSL01", "MOZSL02", "MOZSL03",
```

Also, it looks like `MOZR` is a mobile-only code, so it would technically belong in `ALLOWED_MOBILE_SEARCH_CODES` below, but I don't think there's any real harm in having it here.

sean-rose (approved):
r+wc

PR #2180, opened 2022-10-11
Bug 1784911 - Drop firefox-desktop.background-update pings

akkomar:
Per discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1784911#c15, `firefox-desktop.background-update` pings are sent due to a bug in Glean SDK. This PR makes Decoder drop these pings so they're not showing up in monitoring queries. When this is deployed, a counter with a bug number label will start to be collected in the pipeline. As a follow up note should be added in the bug with a link to a dashboard to monitor this counter. This will allow for monitoring the effect of the future fix.


fbertsch (approved):
LGTM

PR #2184, opened 2022-10-20
Update flex template base image to the latest version

akkomar:
Confirmed that this runs without errors in a test project.


whd (approved):
Any way to keep this updated over time?

akkomar:
I think Dependabot should be able handle that with `package-ecosystem: "docker"`. I'll look into this when I'm back in Nov.

PR #2204, opened 2022-11-18
Upgrade Sanic for python 3.10 support

relud:
should fix ingestion-edge release

### DIFF
--- a/ingestion-edge/ingestion_edge/publish.py
+++ b/ingestion-edge/ingestion_edge/publish.py
@@ -8,7 +8,7 @@ from functools import partial
 from google.api_core.retry import Retry, if_exception_type
 from google.cloud.pubsub_v1 import PublisherClient
 from persistqueue import SQLiteAckQueue
-from typing import Callable, Dict, Tuple
+from typing import Callable, Dict, Optional, Tuple
 from .config import logger
 from .util import AsyncioBatch, HTTP_STATUS
 import google.api_core.exceptions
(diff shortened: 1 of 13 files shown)


ingestion-edge/tests/unit/publish/test_init_app.py
```diff
@@ -4,66 +4,12 @@
...
-        (
-            b"/submit/some/path/with=stuff",
-            "PUT",
-            {"topic": "ingestion", "suffix": "some/path/with=stuff"},
-        ),
-        (b"/submit/sslreports", "POST", {"topic": "tls_error_reports"}),
-        (b"/submit", "PUT", {"topic": "ingestion"}),
-    ],
-)
-async def test_endpoint(app, kwargs, method, mocker, uri_bytes):
```
relud, line 46: i removed this test because it was reaching into sanic internals that no longer exist.

ingestion-edge/ingestion_edge/publish.py
```diff
@@ -25,6 +25,37 @@
...
 
 
+def _legacy_encoding(value: Optional[str], request: Request) -> Optional[str]:
+    """Convert string from UTF-8 with surrogate escape characters to ISO-8859-1.
+
+    As of Sanic 21.3.0 in sanic.http.http1 headers are decoded with the default codec
+    (UTF-8) and errors="surrogateescape", but behavior prior to that was to decode
+    headers with ISO-8859-1, aka latin-1.
+
+    Unit tests hit decode headers in sanic.http.http1, but prod and some integration
```
fbertsch, line 35: If I'm reading this right, we're changing the behavior in prod because of the unit tests - is that right? Is it possible that we change the tests instead and leave prod as-is?

fbertsch (approved):
LGTM, one comment

relud:
waiting to merge this until next week, to avoid issues over the long weekend

PR #2224, opened 2022-12-12
Optimize schemas loading in the Decoder

akkomar:
We [load schemas](https://github.com/mozilla/gcp-ingestion/blob/fb8d34477ad100e2be2f144be83bee6ab10bce57/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/schema/SchemaStore.java#L151) in the [main processing method](https://github.com/mozilla/gcp-ingestion/blob/fb8d34477ad100e2be2f144be83bee6ab10bce57/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L108) of the `ParsePayload` transform. While testing Streaming Engine ([Bug 1792955](https://bugzilla.mozilla.org/show_bug.cgi?id=1792955#c6)) this approach turned out to be suboptimal. The reason was that Streaming Engine uses more worker threads than default Streaming Appliance, scales more agressively, and schema loading is CPU intensive which caused delays when starting new workers.

This PR optimizes loading of schemas in the Decoder by:
* using shared singleton instances of SchemaStores
* eagerly loading schemas in SchemaStores
* initializing SchemaStores in `DoFn.Setup` instead of in the processing method of `ParsePayload` transform. `Setup` is an idiomatic place to initialize `DoFn` resources in Beam

It does improve autoscaling behavior and resource utilization with Streaming Engine enabled. I expect it to cause slightly lower memory utilization with current production configuration.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
@@ -13,9 +13,9 @@ import com.mozilla.telemetry.ingestion.core.schema.PipelineMetadataStore;
 import com.mozilla.telemetry.ingestion.core.schema.SchemaNotFoundException;
 import com.mozilla.telemetry.ingestion.core.transform.NestedMetadata;
 import com.mozilla.telemetry.metrics.PerDocTypeCounter;
+import com.mozilla.telemetry.schema.SchemaStoreSingletonFactory;
 import com.mozilla.telemetry.transforms.FailureMessage;
 import com.mozilla.telemetry.transforms.PubsubConstraints;
-import com.mozilla.telemetry.util.BeamFileInputStream;
 import com.mozilla.telemetry.util.Json;
 import com.mozilla.telemetry.util.JsonValidator;
 import com.mozilla.telemetry.util.Time;
@@ -26,7 +26,6 @@ import java.time.LocalDate;
 import java.time.ZoneOffset;
 import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeParseException;
-import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
@@ -37,13 +36,15 @@ import java.util.zip.CRC32;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.metrics.Distribution;
 import org.apache.beam.sdk.metrics.Metrics;
-import org.apache.beam.sdk.transforms.FlatMapElements;
+import org.apache.beam.sdk.transforms.DoFn;
 import org.apache.beam.sdk.transforms.PTransform;
-import org.apache.beam.sdk.transforms.ProcessFunction;
+import org.apache.beam.sdk.transforms.ParDo;
 import org.apache.beam.sdk.transforms.WithFailures;
 import org.apache.beam.sdk.transforms.WithFailures.Result;
 import org.apache.beam.sdk.values.PCollection;
-import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.sdk.values.PCollectionTuple;
+import org.apache.beam.sdk.values.TupleTag;
+import org.apache.beam.sdk.values.TupleTagList;
 import org.everit.json.schema.Schema;
 import org.everit.json.schema.ValidationException;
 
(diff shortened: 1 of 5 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -75,132 +76,133 @@ public static ParsePayload of(String schemasLocation) {
...
-          } catch (IOException | SchemaNotFoundException | ValidationException
-              | MessageScrubberException | DeprecatedMessageException e) {
-            return FailureMessage.of(ParsePayload.class.getSimpleName(), //
-                ee.element(), //
-                ee.exception());
-          }
-        }));
+    PCollectionTuple parsed = messages
+        .apply(ParDo.of(new Fn()).withOutputTags(outputTag, TupleTagList.of(failureTag)));
+    return WithFailures.Result.of(parsed.get(outputTag), parsed.get(failureTag));
```
akkomar, line 93: Keeping `WithFailures.Result` here instead of plain `PCollectionTuple` as the former makes handling errors upstream easier.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -75,132 +76,133 @@ public static ParsePayload of(String schemasLocation) {
...
+      schemaStore = SchemaStoreSingletonFactory.getJsonSchemaStore(schemasLocation);
+      metadataStore = SchemaStoreSingletonFactory.getPipelineMetadataStore(schemasLocation);
+      validator = new JsonValidator();
+    }
 
-      if (schemaStore == null) {
-        schemaStore = JSONSchemaStore.of(schemasLocation, BeamFileInputStream::open);
-      }
+    @ProcessElement
+    public void processElement(@Element PubsubMessage message, MultiOutputReceiver out) {
```
akkomar, line 106: This is mostly the same code that was in `apply()` method before, the only differences are:
* schemastore and validator initializations are moved to `setup()`
* exception handling is moved here from `expand()` above since now we explicitly use `ParDo`
* instead of `return`ing results we write them to output receiver

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java
```diff
@@ -75,132 +76,133 @@ public static ParsePayload of(String schemasLocation) {
...
+        byte[] normalizedPayload = Json.asBytes(json);
 
-      PerDocTypeCounter.inc(attributes, "valid_submission");
-      PerDocTypeCounter.inc(attributes, "valid_submission_bytes", submissionBytes);
+        PerDocTypeCounter.inc(attributes, "valid_submission");
+        PerDocTypeCounter.inc(attributes, "valid_submission_bytes", submissionBytes);
 
-      return Collections.singletonList(new PubsubMessage(normalizedPayload, attributes));
+        out.get(outputTag).output(new PubsubMessage(normalizedPayload, attributes));
+      } catch (IOException | SchemaNotFoundException | ValidationException
```
akkomar, line 201: It might be worth checking later if using exceptions for control flow here doesn't negatively affect performance.

ingestion-beam/src/main/java/com/mozilla/telemetry/schema/SchemaStoreSingletonFactory.java
```diff
@@ -0,0 +1,55 @@
...
+  private static volatile PipelineMetadataStore pipelineMetadataStore;
+
+  private SchemaStoreSingletonFactory() {
+  }
+
+  /**
+   * Returns a singleton instance of {@link JSONSchemaStore}.
+   */
+  public static JSONSchemaStore getJsonSchemaStore(String schemasLocation) {
+    if (jsonSchemaStore == null) {
```
akkomar, line 30: Double-checked locking is used here because this will be called in `setup()` from multiple `DoFn` instances.

relud (approved):

whd:
FYSA this only landed in prod today (Jan. 11 UTC). I missed the operator input propagation step due to all the other circleci stuff going on, but this actually ended up saving us from propagating a schemas change that would have introduced another pipeline latency event due to `treeherder` weirdness discussed in slack and in https://bugzilla.mozilla.org/show_bug.cgi?id=1809496.

akkomar:
Thanks, I didn't rush it due to the CircleCI issue cleanup going on.
Compared to the previous job memory utilization seems to be ~28% lower now.

PR #2271, opened 2023-02-09
DSRE-1260 - Remove X-Pipeline-Proxy support from the ingestion pipeline

relud:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/DecoderOptions.java
```diff
@@ -21,6 +21,12 @@ public interface DecoderOptions extends SinkOptions, PipelineOptions {
 
   void setProxyIps(String value);
 
+  @Description("Path (local or gs://) to newline delimited text file of regexes matching ip "
+      + "addresses associated with load balancers and proxies that will be ignored for IP lookups.")
+  String getProxySubnetsPath();
```
relud, line 26: @whd I assume you don't actually want to provide subnets in regex format, could you give an example of the format you want for providing proxy subnets?

whd: Ideally I'd be providing the files as-is:

cloudfront: https://d7uri8nf7uskq.cloudfront.net/tools/list-cloudfront-ips

(from https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/LocationsOfEdgeServers.html)

And for GCP:
https://www.gstatic.com/ipranges/cloud.json
https://www.gstatic.com/ipranges/goog.json

From https://support.google.com/a/answer/10026322?hl=en / https://cloud.google.com/vpc/docs/configure-private-google-access

Unfortunately GCP doesn't appear to give CDN-specific allocations, and requires additional parsing to diff user-allocated and service-allocated ranges: https://cloud.google.com/vpc/docs/configure-private-google-access#ip-addr-defaults. I think trusting any GCP-provided IP that isn't allocated to customers would be acceptable but it's probably worth double checking that.

We've used the script they provided before https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-pioneer/tf/modules/study-project/compute_routes in some terraform config, but I would prefer if ops infra is responsible solely for versioning these artifacts as they are provided by upstream providers, and have it be the application's responsibility to parse them.

relud: > And for GCP:
> https://www.gstatic.com/ipranges/cloud.json
> https://www.gstatic.com/ipranges/goog.json

combining these in Java is significantly more complicated than it is in Python. I think it would be better to either re-use that script, or to have a copy of it in this repo that is used to prepare the list of non-client google blocks that is passed to the beam jobs.

whd: This appears to be published as https://github.com/GoogleCloudPlatform/networking-tools-python/tree/main/tools/cidr#how-to-run. I think it would be fine to use that in ops infra as the canonical way of generating the artifact, but it produces output that isn't really meant to be machine readable, though it looks like if you chop the first 3 lines it probably is.

Otherwise, we need to standardize on a format for the output of such a script, and we're back to https://github.com/mozilla/gcp-ingestion/pull/2271#discussion_r1101906259. I don't have an opinion beyond wanting to maintain the least amount of bespoke code as is reasonable. If we need to process into a format I suppose we could reuse the the format https://d7uri8nf7uskq.cloudfront.net/tools/list-cloudfront-ips uses.

relud: > it produces output that isn't really meant to be machine readable, though it looks like if you chop the first 3 lines it probably is

I think that will be good enough for me, i'll write this assuming i need to do the chopping myself.

relud: for google the version info is in the header of the computed file, for cloudfront version would you like me to use the filename, a separate cli argument, the file mtime, or some combination of those?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
```diff
@@ -43,22 +72,32 @@ public PubsubMessage apply(PubsubMessage message) {
...
+        final String[] xffArray = xff.split("\\s*,\\s*");
+        final List<String> filteredXffList = Arrays.stream(xffArray) //
+            .filter(StringUtils::isNotBlank) //
+            .filter(ip -> !proxyMatcher.apply(ip)) //
+            .collect(Collectors.toList());
+        if (xffArray.length != filteredXffList.size()) {
+          xff = String.join(",", filteredXffList);
           countProxyIpAddress.inc();
+          attributes.put("detected_proxy", "true");
+          attributes.put("proxy_list_versions", proxyMatcher.version);
```
relud, line 100: per https://bugzilla.mozilla.org/show_bug.cgi?id=1789992#c22:
> some amount of visibility beyond what was specified originally is desirable

@whd do these fields satisfy that for you?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
```diff
@@ -43,24 +82,30 @@ public PubsubMessage apply(PubsubMessage message) {
       // Prevent null pointer exception
       message = PubsubConstraints.ensureNonNull(message);
 
+      if (proxyMatcher == null) {
+        loadResourcesOnFirstMessage();
```
akkomar, line 86: I think it's generally better to do initializations in `@Setup` like in https://github.com/mozilla/gcp-ingestion/pull/2224/files#diff-6135a4da74da3d17c14be1739ed0c54b9643750f2ef0df5575cd3ca1bca74edfR99. Although I don't expect this to cause any trouble due to size of the lists.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseProxy.java
```diff
@@ -22,19 +28,52 @@
...
-    return new ParseProxy(proxyIpAddress);
+  public static ParseProxy of(String proxyIps, String cloudFrontSubnetList,
+      String googleSubnetList) {
+    return new ParseProxy(proxyIps, cloudFrontSubnetList, googleSubnetList);
   }
 
   /////////
 
-  private final String proxyIps;
+  private static transient ProxyMatcher singletonProxyMatcher;
```
akkomar, line 38: `transient` modifier is redundant since static fields aren't serialized.

akkomar (approved):

relud:
merging this will be handled by @whd

whd (commented):
This IP decoding logic is full of edge cases from previous configurations of the pipeline. One way to simplify this might be:

1. Remove `X-Pipeline-Proxy` special case logic (provides a way to spoof GeoIP that shouldn't be supported anymore)
2. Remove `proxyIps` support (rely on a uniform XFF string including the GCLB IP as the second to last entry)
3. Remove `ip_from_remote_addr` counter and IP lookup logic (we never use this path: https://github.com/mozilla/gcp-ingestion/blob/main/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java#L197)
4. Strip the last two entries from XFF (previously only the last entry was stripped), and change https://github.com/mozilla/gcp-ingestion/blob/main/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java#L192 to target the last entry
5. Apply the new logic from this PR on the XFF string after the stripping in (4) (but see inline note about counting proxy IPs differently based on their index)

There are simpler changes that could be made to this PR but in its current form all IP addresses become spoofable when `listGoogleNonCustomerIps` is set.

relud:
i believe i have made it so

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/GeoCityLookup.java
```diff
@@ -181,70 +179,60 @@ public PubsubMessage apply(PubsubMessage message) {
...
+          } catch (MalformedInputException ignore) {
+            malformedInput.inc();
+          } catch (UnknownHostException | GeoIp2Exception | IndexOutOfBoundsException ignore) {
+            // ignore these exceptions
+          }
         }
 
         // remove client ip from attributes
         attributes.remove(Attribute.X_FORWARDED_FOR);
-        attributes.remove(Attribute.REMOTE_ADDR);
```
relud, line 247: this is now removed in `ParseProxy`

whd:
@relud now that MDN has actually migrated to GCP, new information has come to light: the X-F-F string from GCP looks like this when using the trusted proxy mechanism:

```
clientip,    34.111.97.67, 34.127.193.132, 34.120.208.123, 35.191.10.222
^ client     ^ customer IP ^ trusted IP    ^ customer IP   ^ trusted IP
```

in this case, the first customer IP matches `developer.mozilla.org` and the second `incoming.telemetry.mozilla.org`. Therefore, it seems like we need to adjust the GCP logic to skip two entries instead of one, after which this should finally work.

relud:
> Therefore, it seems like we need to adjust the GCP logic to skip two entries instead of one, after which this should finally work.

done

whd:
I'm tentatively planning to roll this out on Monday (unless main ping stuff is ready to go). I'm running this in stage presently to make sure geo distributions still look ok with the simplification here.

Notably, it doesn't do the thing advertised in the PR title and branch name anymore, it just simplifies geoip decoding logic significantly.

PR #2301, opened 2023-03-03
DENG-547 - Use metadata to indicate source for sample_id 

relud:
- [x] requires https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/762 to ensure we continue sampling on impression id where appropriate

### DIFF
--- a/ingestion-beam/bin/download-schemas
+++ b/ingestion-beam/bin/download-schemas
@@ -78,7 +78,11 @@ function insert_testing_schemas() {
   "mozPipelineMetadata": {
     "bq_dataset_family": "test",
     "bq_metadata_format": "structured",
-    "bq_table": "test_v1"
+    "bq_table": "test_v1",
+    "sample_id_source_uuid_attribute": "client_id",
+    "sample_id_source_uuid_payload_path": [
+      "impression_id"
+    ]
   },
   "type" : "object",
   "title" : "default_schema",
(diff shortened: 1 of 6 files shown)


ingestion-beam/bin/download-schemas
```diff
@@ -78,7 +78,11 @@ function insert_testing_schemas() {
   "mozPipelineMetadata": {
     "bq_dataset_family": "test",
     "bq_metadata_format": "structured",
-    "bq_table": "test_v1"
+    "bq_table": "test_v1",
+    "sample_id_source_uuid_attribute": "client_id",
+    "sample_id_source_uuid_payload_path": [
+      "impression_id"
+    ]
```
sean-rose, line 85: I think it'd be better to have a more flexible approach of using one list of UUID attributes.  As it is you couldn't have more than two attributes, or if you have two attributes the first one can't be a nested field, which seems kind of arbitrary.
```suggestion
    "sample_id_source_uuid_attributes": [
      "client_id",
      "impression_id"
    ]
```

sean-rose: Or I guess with the current implementation `sample_id_source_uuid_attributes` would have to be a list of a list-of-strings to allow any nested fields to be specified as arrays of path parts.

How about supporting standard dotted field notation?  Or do we need to be concerned about field names containing dots?

relud: i don't know if we need to be concerned about field names containing dots, but we do have to be concerned about the difference between attributes and payload paths. both are available, but metadata is stripped from the payload before this step, so it's only available as attributes.

relud: and we have to support both because impression id is only in the payload

sean-rose: Is client ID not also in the payload?

sean-rose: Darn, never mind then.  Carry on. 🫡

sean-rose: One last idea:  If the set of metadata attributes is predefined you could check if the specified name is an attribute, or if not you could always first look for an attribute with the name and then fall back to looking in the payload (or go straight to the payload if it's a nested field).

scholtzan (approved):

PR #2384, opened 2023-06-13
Bug 1836760 - Drop com-freespoke-app as unwanted data

badboy:


lelilia:
already added with PR  #2385 since we seem to have tracked this in two different bugs

PR #2400, opened 2023-06-30
FXA-7925 - Add support for ingesting Cloud Logging messages

akkomar:
See https://mozilla-hub.atlassian.net/browse/FXA-7925

This will allow to ingest telemetry from server-side environments shipped to Pub/Sub via Cloud Logging.

ParseLogEntry transform is used to convert LogEntries to format compatible with structured ingestion. This transform was added back in the day for AET and seems like it was intended to be removed in https://github.com/mozilla/gcp-ingestion/pull/1599. I'm repurposing it since it is no longer used.


akkomar:
This needs to wait with deployment until [design of the server-side part is finalized](https://mozilla-hub.atlassian.net/browse/FXA-7922). However I have tested this end to end and don't expect any major changes to this PR.

relud (approved):

whd (approved):
> @whd I imagine this would read from a dedicated PubSub topic and output to structured-decoded and structured-error topics.

I agree this is the simplest topology. Other variations involving additional pipeline families would require [configuration](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/data-shared/tf/prod/envs/prod/bigquery-new/main.tf#L54-L58) to reason about provisioning pbe and pbd tables.

I think it makes sense for the Data Platform to provision a single topic e.g. `structured-cloudlogs` so that we maintain a single job and configuration for any additional logs. We will still need to coordinate permissions management for cloud logging on an individual basis but that should hopefully be less painful than a new topic for every server application.

akkomar:
https://github.com/mozilla/glean_parser/pull/595 has landed, no changes are required here so this PR is ready to be merged.

akkomar:
Failure of `ingestion-edge-verify-requirements` is a separate issue not related to changes introduced in this PR. It's going to be fixed after https://github.com/mozilla/gcp-ingestion/pull/2415 is merged.

PR #2409, opened 2023-07-13
DENG-1164 - Request resources for flush jobs

relud:
defaults match prod for ingestion edge, remove values from the config by setting environment values to empty strings


whd (approved):
I will test this by building stage from `latest` after this is merged and then updating the prod flush-manager deployment directly to use the built image. If that has the expected effect I'll redeploy all of prod.

PR #2437, opened 2023-08-04
DENG-1320 - Add support for decoding geo and user agent metadata in Cloud Logging-based telemetry

akkomar:
Goes together with https://github.com/mozilla/glean_parser/pull/612


ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
```diff
@@ -70,10 +70,14 @@ public static PipelineResult run(DecoderOptions.Parsed options) {
...
-        // Special case: parsing structured telemetry pings submitted from Cloud Logging
-        .map(p -> options.getLogIngestionEnabled()
-            ? p.apply(ParseLogEntry.of()).failuresTo(failureCollections)
-            : p)
+        // Special case: parsing structured telemetry pings submitted from Cloud Logging.
+        // We do GeoIP decoding here again because we need to extract an IP address from the
+        // message payload envelope in `ParseLogEntry`.
+        .map(p -> options.getLogIngestionEnabled() ? p //
+            .apply(ParseLogEntry.of())//
+            .failuresTo(failureCollections) //
```
akkomar, line 78: This might be at odds with the comment above in line 65. I think it was written back in the day when error tables weren't restricted.

@whd is it currently acceptable to route messages containing ip addresses to error tables?

relud (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
whd: I think we should still scrub IP addresses for data making its way to error tables for consistency with the rest of the platform. We plan to rely on upstream logs to recover from e.g. GeoIP decoding errors if it's necessary. If we want to put IP addresses in `pbe` we'll need to e.g. update https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/monitoring/payload_bytes_error_structured/view.sql#L5

 But it is true that direct `pbe` access is restricted. We could decide to keep IP addresses in PBE as long as we don't expose them for downstream analysis.

akkomar: > I think we should still scrub IP addresses for data making its way to error tables for consistency with the rest of the platform. We plan to rely on upstream logs to recover from e.g. GeoIP decoding errors if it's necessary. If we want to put IP addresses in `pbe` we'll need to e.g. update https://github.com/mozilla/bigquery-etl/blob/main/sql/moz-fx-data-shared-prod/monitoring/payload_bytes_error_structured/view.sql#L5

Got it, we can scrub IPs early in the transform to make sure they don't go to the errors collection.

One thing to note that I might not made clear is that in this case IPs are in the payload as Cloud Logging sink does not allow to attach custom attributes to Pub/Sub messages. So they'd already be excluded from the `pbe` view. @whd - does this change your assessment?

akkomar: Thinking about this more - it moght be worth to scrub for consistency, I've implemented this in https://github.com/mozilla/gcp-ingestion/pull/2437/commits/7167711312d97bb23b4574858c271985201a1f4c.

whd (approved):
I'm a little hazy on what the specific guarantees about log formatting are going to be present for server application logs. Stuff like whether we know which IP address related fields will have values and that they will have consistent meanings across applications (e.g. if there are levels of indirection in services handling and emitting logs, nginx proxies etc.).

I assume we can work around any consistencies in input on our end via [mozPipelineMetadata](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/schemas/metadata/metaschema/metaschema.1.schema.json#L6).

akkomar:
> I'm a little hazy on what the specific guarantees about log formatting are going to be present for server application logs. Stuff like whether we know which IP address related fields will have values and that they will have consistent meanings across applications (e.g. if there are levels of indirection in services handling and emitting logs, nginx proxies etc.).

We're aiming to ensure formatting and field meaning through clients using glean_parser-generated code for logging.

PR #2440, opened 2023-08-08
DENG-1267 - Handle geoip skip from metadata

relud:
blocks https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/777


whd (approved):
A bit unfortunate that this requires some code reshuffling.

At any rate, I wrote up the empirical test results in https://mozilla-hub.atlassian.net/browse/DENG-1267?focusedCommentId=720803. The only thing to note is that we'll likely want to set this to `2` for GCP-based proxies. I'm told that `bedrock` may still be in AWS and we'll want to check what the appropriate value is for this metadata on a case by case basis.

PR #2458, opened 2023-09-05
Updates to the position parameter.

quiiver:
Nan and I did some validating of the recent addition of the position param during out 1-1 and we found a couple things i missed in the code review. sorry!

* position is part of the message payload, not a top level key in the pubsub message.
* the extraction key is `position`
* position should only be set for topsite impressions.


ncloudioj (approved):
r=nanj
Looks good, thanks!

chelseatroy (approved):
C'est bon!

PR #2466, opened 2023-09-13
DENG-1352 - Add support for parsing FoG pings in contextual services reporter

relud:
for verification three batch jobs have been run against a sample of prod data for a full day with `reportingEnabaled=false` and `logReportingUrls=true`. The first was without this code change and only read contextual_services messages. The second included this code change and also only read contextual_services messages. The third included this code change, read contextual_services and firefox_desktop messages, and had `limitLegacyDesktopVersion=true`.

By comparing the first two jobs verified that this change will have no impact until the new doctypes are sent to it *and* we update the configuration. This means that this PR is safe to merge, and publishing the new doctypes to the pubsub topic is safe, and once both of those are done we can deploy the configuration changes to switch to the new logic.

By comparing the second and third jobs' number of SendRequest "errors" and extracting aggregate impression counts, I was able to validate the impact of this change aligns with our expectations of slight increase (~1.3%) due to better delivery guarantees in firefox_desktop than contextual_services.

Other common errors were within expected limits of ~1.3% or less change. For less common errors: `RejectedMessageException: Firefox version does not match doctype` went away as expected because that's the intended impact of `limitLegacyDesktopVersion=true`. `InvalidUrlException: Could not parse reporting URL` saw a 28% increase from ~9k to ~11.5k, which is big relative to eachother, but small overall.

One noteworthy change from validation is that `RejectedMessageException: Matches heuristic from CONSVC-1764` applies across `firefox_desktop.top_sites` where before it only applied to `contextual_services.topsites_click` before, and this resulted in a tenfold increase in occurence (from ~2k to ~20k). Inspecting this more closely I found that the number of impacted top sites clicks increased by 1.4% while the rest is attributed to newly rejecting top sites impressions.

The other noteworthy change from validation is that I saw significant performance degredation from restricting `user_agent_version` in `VerifyMetadata`, sometimes causing pipeline failures. As such, I moved this action into `FilterByDocType` where the messages can be dropped entirely instead of sent to the error output, which resolved the issue.


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -165,25 +206,28 @@ public Result<PCollection<SponsoredInteraction>, PubsubMessage> expand(
...
+
           // Set the source based on the value of the docType
-          if (DT_TOPSITES_CLICK.equals(docType) || DT_TOPSITES_IMPRESSION.equals(docType)) {
+          if (DT_TOPSITES.equals(docType) || DT_TOPSITES_CLICK.equals(docType)
+              || DT_TOPSITES_IMPRESSION.equals(docType)) {
             interactionBuilder.setSource(SponsoredInteraction.SOURCE_TOPSITES);
-          } else if (DT_QUICKSUGGEST_IMPRESSION.equals(docType)
+          } else if (DT_QUICKSUGGEST.equals(docType) || DT_QUICKSUGGEST_IMPRESSION.equals(docType)
               || DT_QUICKSUGGEST_CLICK.equals(docType)) {
             interactionBuilder.setSource(SponsoredInteraction.SOURCE_SUGGEST);
```
relud, line 217: the FoG pings have a `source` metric, but it doesn't seem to be the same as this

quiiver:
flagging @chelseatroy and @ncloudioj for additional 👀

quiiver (approved):
lgtm!

relud:
@whd fyi: merging this now because validation testing showed that it has no impact without config changes. i will be filing a dsre ticket to request the config changes before the change freeze, in sync with merging a related bigquery-etl pr and a communication to downstream consumers that internal changes are occurring.

whd:
> merging this now because validation testing showed that it has no impact without config changes

This has been deployed to prod in case you want to triple check it's a no-op without config changes.

Issue #2503, opened 2023-11-06
Update Beam SDK version


akkomar:
We are now running on Beam SDK version 2.42.0 which is [deprecated](https://cloud.google.com/dataflow/docs/support/sdk-version-support-status#apache-beam-sdks-2x).

Bumping sdk version breaks our build (see https://github.com/mozilla/gcp-ingestion/pull/2484 and previous PRs opened by Dependabot). Immediate reason of this breakage is removal of `google-http-client-jackson2` dependency which was transitively imported with `google-cloud-bigquery` and [removed in version 2.21](https://github.com/googleapis/java-bigquery/blob/main/CHANGELOG.md#features-13).

I tried two approaches for fixing the build with updated SDK so far:
* switch JacksonFactory to GsonFactory (as recommended in https://cloud.google.com/java/docs/reference/google-http-client/latest/com.google.api.client.json.jackson2.JacksonFactory)
* add `google-http-client-jackson2` dependency to ingestion-core

Both make the code compile but some tests are failing, probably caused by differences in how Jackson and Gson parse JSON.

PR #2518, opened 2023-11-27
Update Beam SDK to 2.51

akkomar:
Fixes https://github.com/mozilla/gcp-ingestion/issues/2503

@whd I went through the code and looks like changes to PubsubMessage in the SDK are related to new features we don't use so this PR seems to be low risk.

### DIFF
--- a/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParsePayloadTest.java
+++ b/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParsePayloadTest.java
@@ -9,6 +9,7 @@ import com.mozilla.telemetry.options.InputFileFormat;
 import com.mozilla.telemetry.options.OutputFileFormat;
 import com.mozilla.telemetry.schema.SchemaStoreSingletonFactory;
 import com.mozilla.telemetry.util.Json;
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParsePayloadTest.java
```diff
@@ -26,7 +27,7 @@
 import org.junit.Rule;
 import org.junit.Test;
 
-public class ParsePayloadTest {
+public class ParsePayloadTest extends TestWithDeterministicJson {
```
akkomar, line 30: This is is required to keep the test deterministic. Either bumping jackson version or excluding new fields from serialized JSON caused field order in test output to change.

Following changes in this file are reordering fields in alphabetical order.

relud (approved):

whd (approved):

akkomar:
I can reproduce this starting the runner locally so it doesn't seem like a template's fault.
I tried couple of different SDK versions and found that the job starts failing on 2.44.0. Looks like something changed between 2.43.0 and 2.44.0 that's causing this failure.

PR #2521, opened 2023-11-27
As of beam 2.44 google-cloud-storage is no longer a test-only dependency

relud:
fixes https://github.com/mozilla/gcp-ingestion/pull/2518#issuecomment-1828448944

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -75,7 +75,6 @@
         <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-storage</artifactId>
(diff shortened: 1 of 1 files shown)


whd (approved):

akkomar:
It might be better to import `beam-sdks-java-extensions-google-cloud-platform-core` as mentioned in [v2.44.0 release notes](https://github.com/apache/beam/releases/tag/v2.44.0).

relud:
i tried adding that and it didn't have any impact, i don't mind adding it just to be thorough

whd:
This appears to have done the trick, stage is running on 2.51.0 now.

PR #2531, opened 2023-12-19
Add support for new searchwith contextual services ping

mkaply:
We're adding a new contextual services ping and I think this is all that's required to support it.

Problem: We need to ingest an additional data source from Firefox and put it into the contextual service pipeline.

Solution: Add the new type (searchwith) to the existing lists and data based on the glean ping.

Acceptance criteria: New data is available to the beam,

It does appear there is a test here I didn't spot the first time. I'll see if I can update and run locally.


mkaply:
Firefox code changes are here:

https://phabricator.services.mozilla.com/D196490

chelseatroy (commented):
Looks like `topsites` and `quicksuggest` are represented in the unit tests for these two classes, but not `searchwith`. Could we please add that prior to merge? 🥺

mkaply:
So looking at these tests, they seem to be for the old mechanism (the explicit quicksuggest-impression versus the new quick-suggest doctype). 

Is there some straightforward way for me to run these tests if I modify them?

mkaply:
Test added.

quiiver (approved):
looks good

PR #2566, opened 2024-03-22
Fix namespace types for Mobile Suggest

cbguder:
Turns out the pipeline code also had these wrong. Fixed to support actual mobile namespaces (org-mozilla-fenix, org-mozilla-ios-fennec, etc.).

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
@@ -66,9 +66,7 @@ public class ParseReportingUrl extends
       ImmutableList.of("top_sites"), DT_QUICKSUGGEST, ImmutableList.of("quick_suggest"),
       DT_SEARCHWITH, ImmutableList.of("search_with"));
 
-  // namespaces and doctypes for Firefox Mobile glean telemetry
-  private static final String NS_IOS = "firefox-ios";
-  private static final String NS_ANDROID = "fenix";
+  // doctypes for Firefox Mobile glean telemetry
   private static final String DT_MOBILE_QUICKSUGGEST = "fx-suggest";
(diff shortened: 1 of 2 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -180,37 +178,45 @@ public Result<PCollection<SponsoredInteraction>, PubsubMessage> expand(
 
             // parse position for desktop.
             interactionBuilder.setPosition(parsePosition(payload).orElse("no_position"));
-          } else if (NS_IOS.equals(namespace) || NS_ANDROID.equals(namespace)) {
+          } else {
+            // Namespace is one of the many mobile namespaces:
```
dmueller, line 182: does this list of namespaces come from the registered glean apps? I'm wondering if we would want some check on this else branch for the namespaces or the doc type checks below will be sufficient regardless

dmueller (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
cbguder: Yeah, I'm told the source of truth for this is https://github.com/mozilla-services/mozilla-pipeline-schemas/tree/generated-schemas/schemas, but we filter what namespaces are allowed in this job here: https://github.com/mozilla-services/cloudops-infra/blob/0c194ac65fd67f5170854da5c58cc313b1bda7aa/projects/beam/tf/modules/dataflow-jobs/main.tf#L67

We did not have an additional namespace check here so far, and I certainly don't think it's a bad idea to add one; but I also think we should refactor this code further to get rid of the `if namespace == foo {} else if namespace == bar {}` type structure altogether.

Issue #2609, opened 2024-06-21
Unrouteable message due to normalization of keys


whd:
/CC @akkomar 

Here's a stripped down ping demonstrating the failure. Basically, it appears that this ping passes schema validation treating `wr.framebuild_time` and `wr_framebuild_time` as both acceptable but then chokes when reconciling the two at a later step. If you remove one or the other this ping gets routed correctly.

This should probably be treated as an error since a client should not be sending one or the other of these.

```json
{
    "ping_info": {
        "seq": 0,
        "start_time": "2024-06-20T11:50+03:00",
        "end_time": "2024-06-21T10:41+03:00",
        "reason": "overdue"
    },
    "client_info": {
        "telemetry_sdk_build": "52.7.0",
        "app_display_version": "115.12.0",
        "app_channel": "esr",
        "app_build": "20240603145132",
        "architecture": "x86",
        "os": "Windows",
        "os_version": "6.1",
        "first_run_date": "2024-06-18+03:00",
        "build_date": "1970-01-01T00:00:00+00:00",
        "client_id": "e7ee6fd7-ec98-451e-b75e-f8443b46297a",
        "windows_build_number": 7600
    },
    "metrics": {
        "timing_distribution": {
            "wr.framebuild_time": {
                "values": {
                    "0": 0
                },
                "sum": 0
            },
            "wr_framebuild_time": {
                "values": {
                    "0": 0
                },
                "sum": 0
            }
        }
    }
}
```

```
extendedStackTrace: "java.lang.ClassCastException: class com.fasterxml.jackson.databind.node.ArrayNode cannot be cast to class com.fasterxml.jackson.databind.node.ObjectNode (com.fasterxml.jackson.databind.node.ArrayNode and com.fasterxml.jackson.databind.node.ObjectNode are in unnamed module of loader 'app')
	at com.mozilla.telemetry.ingestion.core.transform.PubsubMessageToObjectNode$Payload.processField(PubsubMessageToObjectNode.java:439) ~[ingestion-sink-0.1-SNAPSHOT.jar:?]
	at com.mozilla.telemetry.ingestion.core.transform.PubsubMessageToObjectNode$Payload.transformForBqSchema(PubsubMessageToObjectNode.java:398) ~[ingestion-sink-0.1-SNAPSHOT.jar:?]
[...]

PR #2610, opened 2024-06-24
Fix handling of duplicate normalized field names in sink

akkomar:
Fixes the issue with map fields reported in https://github.com/mozilla/gcp-ingestion/issues/2609


ingestion-core/src/test/resources/testdata/payload-format-duplicate-normalized-field-names-input.ndjson
```diff
@@ -0,0 +1,10 @@
+{"attributeMap":{"document_namespace":"namespace_0","document_type":"bar","document_version":"1"},"payload":{"payload":{"test_int":-2, "test.int":-2}}}
```
BenWu, line 1: Is the order in the object deterministic when read?  e.g. would `{"test_int":-2, "test.int":-3}` give consistent results?  In your comment, you said `when the "dot version" of the field precedes the "underscore version"` but I'm not understanding why the order matters

akkomar: Order is deterministic when read. Yes, `{"test_int":-2, "test.int":-3}` give consistent results.

Order matters only for map fields because if the "dot" version comes first, we will transform its value with `PubsubMessageToObjectNode.Payload#expandMapType` and insert into the normalized ("underscore") version of the field in https://github.com/mozilla/gcp-ingestion/blob/e1b318c84f189d979d75768097c51e6aaf97139a/ingestion-core/src/main/java/com/mozilla/telemetry/ingestion/core/transform/PubsubMessageToObjectNode.java#L582. Then when in the subsequent iteration we try to read this expanded value we crash.

Also compare the last line in this file with a single line in `payload-format-duplicate-normalized-map-field-input.ndjson`. These two differ only in the order of "dot" and "underscore" fields. All of the samples from this file pass through the sink, the other one throws an exception.

BenWu (approved):

PR #2618, opened 2024-07-10
Bug 1907131 remove namespace com-goodtoolapps-zeus.baseline

edugfilho:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -118,6 +118,7 @@ public class MessageScrubber {
       .put("com-jio-web-ios", "1887756") //
       .put("com-pekobbrowser-unblocksites", "1896079") //
       .put("com-eusoft-trans-en", "1902673") //
+      .put("com-goodtoolapps-zeus.baseline", "1907131") //
```
BenWu, line 121: I don't think the ping name should be there
```suggestion
      .put("com-goodtoolapps-zeus", "1907131") //
```

edugfilho: My bad. I took it from the generated bug and I was under the impression that the ping name was part of the namespace, and even thought it was weird. Should've double checked.
Thank you

BenWu (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
BenWu: Yeah the bugs created by the "Missing document namespaces" and "Missing DocTypes/versions" links are different.  "Missing document namespaces" seems to be more correct

PR #2636, opened 2024-09-25
updated Stub URI parser to include new windows_ubr telemetry

erchen03:
per https://bugzilla.mozilla.org/show_bug.cgi?id=1859597
awaiting this PR to be merged https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/828

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
@@ -288,7 +288,8 @@ public class ParseUri {
         putBoolPerCode(ImmutableMap.of("set_default", 2)), //
         putString("download_ip"), putString("attribution"),
         putIntegerAsString("profile_cleanup_prompt"), putBool("profile_cleanup_requested"),
-        putString("distribution_id"), putString("distribution_version"));
+        putString("distribution_id"), putString("distribution_version"),
+        putInteger("windows_ubr"));
 
     private static BiConsumer<String, ObjectNode> ignore() {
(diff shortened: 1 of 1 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -288,7 +288,8 @@ private static class InvalidIntegerException extends InvalidUriException {
         putBoolPerCode(ImmutableMap.of("set_default", 2)), //
         putString("download_ip"), putString("attribution"),
         putIntegerAsString("profile_cleanup_prompt"), putBool("profile_cleanup_requested"),
-        putString("distribution_id"), putString("distribution_version"));
+        putString("distribution_id"), putString("distribution_version"),
+        putInteger("windows_ubr"));
```
BenWu, line 292: Important to note that order here matters so looking at https://phabricator.services.mozilla.com/D223765, it looks like this will be `ubr/build_id`. Small thing, but it would be easier for us merging this if this PR was combined with https://github.com/mozilla/gcp-ingestion/pull/2638

BenWu (commented):
Tests are failing from format check

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
akkomar: It would also be good to have these fields covered in `ParseUriTest`, but it's not a blocker for this PR since `distribution_id` and `distribution_version` haven't been added last time this file was touched.

PR #2658, opened 2024-10-21
Bump Beam SDK to 2.60.0

akkomar:
In version 2.52 Beam removed all Avro-dependent code from core package. I filed https://mozilla-hub.atlassian.net/browse/DENG-2029 last year to get rid of Avro in our codebase because of that. It turned out that BigQueryIO uses Avro, so we need to add `beam-sdks-java-extensions-avro` dependency anyway. Therefore I'm not removing Avro support in this PR.

Deployment caveat: [PubSubIO's read is update incompatible with previous versions so streaming jobs can't be updated](https://github.com/apache/beam/blob/master/CHANGES.md#breaking-changes-7).

I tested this on a synthetic structured-logging stream in https://console.cloud.google.com/dataflow/jobs/us-west1/2024-10-21_05_13_30-3138935860679087156;graphView=0?project=akomar-server-telemetry-poc&pageState=(%22dfTime%22:(%22l%22:%22dfJobMaxTime%22)). I haven't seen any issues, but it would probably be good to run this in stage for few hours first.

closes https://github.com/mozilla/gcp-ingestion/pull/2651

### DIFF
--- a/ingestion-beam/pom.xml
+++ b/ingestion-beam/pom.xml
@@ -61,6 +61,10 @@
         </dependency>
         <dependency>
             <groupId>org.apache.beam</groupId>
+            <artifactId>beam-sdks-java-extensions-avro</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.beam</groupId>
             <artifactId>beam-sdks-java-io-google-cloud-platform</artifactId>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


whd (approved):
> Deployment caveat: [PubSubIO's read is update incompatible with previous versions so streaming jobs can't be updated](https://github.com/apache/beam/blob/master/CHANGES.md#breaking-changes-7).

Good to know. We currently treat all code changes as non-updateable  for this reason.

I'll run this in stage and merge after a few hours. We no longer send a 1% sample of structured and telemetry data there by default but there's at least some data going to stub installer.

PR #2719, opened 2025-01-10
update ping version number to account for windows_ubr and stub build id

erchen03:


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseUri.java
```diff
@@ -245,10 +245,10 @@ private static class InvalidIntegerException extends InvalidUriException {
     public static final String FUNNELCAKE = "funnelcake";
 
     public static final Pattern PING_VERSION_PATTERN = Pattern
-        .compile(String.format("^v(?<%s>[6-9])(-(?<%s>\\d+))?$", VERSION, FUNNELCAKE));
+        .compile(String.format("^v(?<%s>[6-10])(-(?<%s>\\d+))?$", VERSION, FUNNELCAKE));
```
BenWu, line 248: I'm not sure why the error message in CI is cryptic but the issue is this regex is invalid.  Something like this should work to get numbers [6, 19]:
```suggestion
        .compile(String.format("^v(?<%s>[6-9]|1[0-9])(-(?<%s>\\d+))?$", VERSION, FUNNELCAKE));
```
I would also recommend adding a v10 test case to https://github.com/mozilla/gcp-ingestion/blob/fb7e9ed9e891e3e2320d85e05d7c1a1aedb57780/ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java#L321
This might be tricky to do if you don't have the environment setup so I can help with this if you need

erchen03:
Could you help me with the v10 test case? It looks the v9 test case is also missing

BenWu:
I just updated the tests so I'll get review from someone else

ingestion-beam/src/test/java/com/mozilla/telemetry/decoder/ParseUriTest.java
```diff
@@ -317,36 +327,62 @@ private static Map<String, String> joinCases(List<Map<String, String>> cases) {
...
         Arrays.asList(//
             "v6/" + String.join("/", Collections.nCopies(35, " ")), //
             "v7/" + String.join("/", Collections.nCopies(36, " ")), //
             "v8/" + String.join("/", Collections.nCopies(38, " ")), //
-            "", "v6", "v7", "v8")))
-        .map(v -> "{\"attributeMap\":{\"message_id\":\"00000000001\",\"uri\":\"/stub/" + v
+            "v9/" + String.join("/", Collections.nCopies(40, " ")), //
+            "v10/" + String.join("/", Collections.nCopies(42, " ")), //
+            "v6", "v7", "v8", "v9", "v10", // UnexpectedPathElementsException
+            "", "v1", "v61", "v21" // UnknownPingVersionException
```
akkomar, line 355: Thanks for adding comments, it's not immediately obvious what's going on here ☠️

akkomar (approved):
LGTM, test seems to be covering v9 and v10.

PR #2721, opened 2025-01-10
Increase dates in streaming integration tests

BenWu:
Tests just started [failing](https://app.circleci.com/pipelines/github/mozilla/gcp-ingestion/7109/workflows/8f027d24-5e0c-42d9-84ad-3077b173099c/jobs/82130) with 
```
Value 1578862938123456 for partition column submission_timestamp of the destination table ******************************:gcloud_test_dataset_temp_77037075_c866_42bb_8f84_39534d531e0b.my_test_table is outside the allowed bounds. 
You can only stream to partitions of date range within 1825 days in the past and 366 days in the future relative to the current date.
``` 
Setting this to 2026-01-10 should give us another 6 years before it breaks again


scholtzan (approved):

ingestion-beam/src/test/java/com/mozilla/telemetry/integration/BigQueryIntegrationTest.java
```diff
@@ -307,6 +307,10 @@ public void canSetStrictSchemaMode() throws Exception {
         matchesInAnyOrder(Lists.newArrayList("{\"type\":\"main\"}", null, "{\"type\":\"main\"}")));
   }
 
+  /**
+   * This test will fail if the submission_timestamp is over six years from the current date.
```
sean-rose, line 311: 1825 days = 5 years
```suggestion
   * This test will fail if the submission_timestamp is over five years from the current date.
```

PR #2744, opened 2025-02-27
Stream support for unnesting events in ParseAmplitudeEvents

scholtzan:
We were running into the following exception when unnesting events while running this pipeline in streaming mode:

```
java.util.ConcurrentModificationException
	at java.base/java.util.ArrayList$ArrayListSpliterator.tryAdvance(ArrayList.java:1634)
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127)
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:230)
	at java.base/java.util.stream.MatchOps$MatchOp.evaluateSequential(MatchOps.java:196)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.anyMatch(ReferencePipeline.java:528)
	at com.mozilla.telemetry.amplitude.ParseAmplitudeEvents.lambda$extractEvents$11(ParseAmplitudeEvents.java:240)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at com.mozilla.telemetry.amplitude.ParseAmplitudeEvents.extractEvents(ParseAmplitudeEvents.java:228)
	at com.mozilla.telemetry.amplitude.ParseAmplitudeEvents.lambda$expand$1fb25989$1(ParseAmplitudeEvents.java:114)
	at org.apache.beam.sdk.transforms.FlatMapElements$FlatMapWithFailures$2.processElement(FlatMapElements.java:362)
```

This PR refactors the unnesting of events to be compatible when streaming

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/amplitude/ParseAmplitudeEvents.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/amplitude/ParseAmplitudeEvents.java
@@ -23,6 +23,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
+import java.util.stream.StreamSupport;
 import org.apache.beam.sdk.io.gcp.pubsub.PubsubMessage;
 import org.apache.beam.sdk.transforms.FlatMapElements;
 import org.apache.beam.sdk.transforms.PTransform;
@@ -111,7 +112,7 @@ public class ParseAmplitudeEvents extends
 
           // each event from the payload is mapped to a separate Amplitude event
           try {
-            final ArrayList<ObjectNode> events = extractEvents(payload, namespace, docType);
+            final List<ObjectNode> events = extractEvents(payload, namespace, docType);
 
             return events.stream().map((ObjectNode event) -> {
               AmplitudeEvent.Builder amplitudeEventBuilder = AmplitudeEvent.builder();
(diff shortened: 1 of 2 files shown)


BenWu (approved):
Is `singletonAllowedEvents` the one that's being modified or `events`?  I'm not sure where the concurrency is coming from and how this fixes it but the code looks fine

scholtzan:
`events` was the one being modified. It was basically just a list with all the events from the event ping. `singletonAllowedEvents` contains the eventAllowlist read from the CSV file in GCS

akkomar (commented):
> events was the one being modified

But they weren't traversed at this time.

I think in this case `singletonAllowedEvents` modification was causing this error. It's a singleton, but initialized in `readAllowedEventsFromFile()` which is called for each message. Beam runs stages in multiple threads in parallel, so this initialization will run in parallel too.

I guess this PR helped because now traversal through `singletonAllowedEvents` is faster.

I think the right way to fix this would be to initialize singleton with `synchronized` block like we do in https://github.com/mozilla/gcp-ingestion/blob/main/ingestion-beam/src/main/java/com/mozilla/telemetry/schema/SchemaStoreSingletonFactory.java#L37-L44 to ensure it is thread-safe. We could also put initialization in a `@Setup` method like in https://github.com/mozilla/gcp-ingestion/blob/main/ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParsePayload.java#L107-L108, but just for readability (it can run multiple times).

PR #2745, opened 2025-02-28
singletonAllowedEvents initialize in synchronized block

scholtzan:
Follow-up to https://github.com/mozilla/gcp-ingestion/pull/2744#pullrequestreview-2651560882

During initialization `ConcurrentModificationException` when initializing the event allow list. This doesn't block the pipeline from running, but putting it into a `synchronized` block should fix the issue

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/amplitude/ParseAmplitudeEvents.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/amplitude/ParseAmplitudeEvents.java
@@ -40,7 +40,7 @@ public class ParseAmplitudeEvents extends
 
   private final String eventsAllowListPath;
 
-  private static transient List<String[]> singletonAllowedEvents;
+  private static transient volatile List<String[]> singletonAllowedEvents;
 
   public static ParseAmplitudeEvents of(String eventsAllowListPath) {
     return new ParseAmplitudeEvents(eventsAllowListPath);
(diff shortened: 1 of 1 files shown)


BenWu (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/amplitude/ParseAmplitudeEvents.java
```diff
@@ -192,30 +192,32 @@ List<String[]> readAllowedEventsFromFile() throws IOException {
       throw new IllegalArgumentException("File location must be defined");
     }
 
-    try (InputStream inputStream = BeamFileInputStream.open(eventsAllowListPath);
-        InputStreamReader inputStreamReader = new InputStreamReader(inputStream);
-        BufferedReader reader = new BufferedReader(inputStreamReader)) {
-      singletonAllowedEvents = new ArrayList<String[]>();
+    synchronized (ParseAmplitudeEvents.class) {
```
akkomar, line 195: When multiple threads arrive here at the same time at startup, they'll proceed to enter the synchronized block in turn. As a result they'll read the file multiple times.
I think it's safe from concurrency point of view as other threads will see the list fully initialized (because of the synchronized block), but if we wanted to make sure that this file is read only once we'd need to wrap this in another null check:
```suggestion
    synchronized (ParseAmplitudeEvents.class) {
       if (singletonAllowedEvents == null) {
         ...
```
and make singletonAllowedEvents `volatile`.

akkomar (approved):
I think this should work, but added a suggestion if we want to be super correct.

PR #2761, opened 2025-03-27
Filter desktop top-sites pings to drop pings from version greater than 136

gleonard-m:
[DSRE-1938](https://mozilla-hub.atlassian.net/browse/DSRE-1938)

With the rollout of Ads Unified API to Firefox desktop clients the Ads Unified API is now used for impression and click reporting.

Firefox will continue to submit `top-sites` pings but with the `reportingUrl1` metric not set.  This missing metric is resulting in the increase in error detections and BQ streaming inserts.

As of Firefox 137  (release date 2025-04-01) Firefox will be updated to use the Ads Unified API exclusively.  We can now filter the `top-sites` by version and prevent the excessive error reporting.  

This PR updates the filtering to drop any Firefox desktop `top-sites` above version 136.

[DSRE-1938]: https://mozilla-hub.atlassian.net/browse/DSRE-1938?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/FilterByDocType.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/FilterByDocType.java
@@ -80,6 +80,7 @@ public class FilterByDocType
       if ("contextual-services".equals(namespace) || "firefox-desktop".equals(namespace)) {
         // Verify Firefox version here so rejected messages don't go to error output
         final int minVersion;
+        int maxVersion = Integer.MAX_VALUE;
         final boolean isLegacyDesktop;
         if (doctype.startsWith("topsites-")) {
           minVersion = 87;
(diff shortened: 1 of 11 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/FilterByDocType.java
```diff
@@ -100,8 +105,8 @@ public void processElement(@Element PubsubMessage message, OutputReceiver<Pubsub
         final String version = message.getAttribute(Attribute.USER_AGENT_VERSION);
         try {
           if (version == null || minVersion > Integer.parseInt(version)
-              || (limitLegacyDesktopVersion && isLegacyDesktop
-                  && 116 <= Integer.parseInt(version))) {
+              || maxVersion < Integer.parseInt(version) || (limitLegacyDesktopVersion
+                  && isLegacyDesktop && 116 <= Integer.parseInt(version))) {
             PerDocTypeCounter.inc(message.getAttributeMap(), "version_filter_rejected");
```
gleonard-m, line 110: I'm not sure if this should still inc the counter.

cbguder: I think it's fine. As far as I can tell this metric has been 0 for a while now, so we would get some visibility into how many messages we're dropping without it necessarily getting mixed up with messages dropped due to minVersion.

cbguder (approved):
Thank you!

PR #2777, opened 2025-05-01
Add country code and form factor to Sponsored Suggest reporting URLs

cbguder:
This is per the [integration guidelines](https://docs.google.com/document/d/165T1slU78f6ZqFhvFoTtlpUtwsMdEzwICi8_F297jUg/edit?usp=sharing) for the new AMP Suggest API.

In the following weeks, we will also:
1. Add the Firefox version to click URLs
2. Aggregate impressions similar to what we're doing for Sponsored Tiles today

I originally considered doing this during ingestion (see https://github.com/mozilla-services/mars/pull/573), but ultimately decided against it as:

1. The 2 changes mentioned above will require us to update the Contextual Services pipeline anyway, and it makes sense to have it all in one place
2. There will be some portion of users with country codes that don't match the suggest set (e.g. QA users in Romania testing UK suggest). I think it makes sense to send accurate country information in these cases.

### DIFF
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
@@ -270,7 +270,11 @@ public class ParseReportingUrl extends
           // https://mozilla-hub.atlassian.net/browse/DENG-392
           if (SponsoredInteraction.FORM_DESKTOP.equals(interaction.getFormFactor())
               && SponsoredInteraction.SOURCE_SUGGEST.equals(interaction.getSource())) {
-            addAdditionalDimensionsForSuggest(builtUrl, interaction);
+            addCustomDataForDesktopSuggest(builtUrl, interaction);
+          }
+
+          if (SponsoredInteraction.SOURCE_SUGGEST.equals(interaction.getSource())) {
+            addAdditionalDimensionsForInternationalSuggest(builtUrl, interaction, payload);
           }
 
           reportingUrl = builtUrl.toString();
@@ -289,7 +293,7 @@ public class ParseReportingUrl extends
         }));
   }
 
-  private static void addAdditionalDimensionsForSuggest(BuildReportingUrl builtUrl,
+  private static void addCustomDataForDesktopSuggest(BuildReportingUrl builtUrl,
       SponsoredInteraction interaction) {
     Stream<Optional<String>> customDataElements = Stream.of(
         Optional.ofNullable(interaction.getScenario()),
@@ -299,6 +303,24 @@ public class ParseReportingUrl extends
     String customDataParam = customDataElements.flatMap(Optional::stream).reduce(originalValue,
         (output, param) -> String.format("%s_%s", output, param));
     builtUrl.addQueryParam(BuildReportingUrl.PARAM_CUSTOM_DATA, customDataParam);
+  }
+
+  private static void addAdditionalDimensionsForInternationalSuggest(BuildReportingUrl builtUrl,
+      SponsoredInteraction interaction, ObjectNode payload) {
+    // Do not add additional parameters for legacy suggest URLs
+    String baseUrl = builtUrl.getBaseUrl();
+    if (baseUrl.contains("ampxdirect.com") || baseUrl.startsWith("https://imp.mt48.net/static")) {
+      return;
+    }
+
+    if (!payload.hasNonNull(Attribute.NORMALIZED_COUNTRY_CODE)) {
+      throw new RejectedMessageException(
+          "Missing required payload value " + Attribute.NORMALIZED_COUNTRY_CODE, "country");
+    }
+
+    builtUrl.addQueryParam(BuildReportingUrl.PARAM_COUNTRY_CODE,
+        payload.get(Attribute.NORMALIZED_COUNTRY_CODE).asText());
+    builtUrl.addQueryParam(BuildReportingUrl.PARAM_FORM_FACTOR, interaction.getFormFactor());
   }
 
   private static void addAdditionalDimensionsForTopSitesClicks(BuildReportingUrl builtUrl,
--- a/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
+++ b/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
@@ -265,15 +265,14 @@ public class ParseReportingUrl extends
             addAdditionalDimensionsForTopSitesClicks(builtUrl, attributes);
           }
 
-          // If we're on desktop and quicksuggest then add the attribution source (data sharing
-          // preference) to the `custom-data` query param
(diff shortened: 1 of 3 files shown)


ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -301,6 +305,24 @@ private static void addAdditionalDimensionsForSuggest(BuildReportingUrl builtUrl
     builtUrl.addQueryParam(BuildReportingUrl.PARAM_CUSTOM_DATA, customDataParam);
   }
 
+  private static void addAdditionalDimensionsForInternationalSuggest(BuildReportingUrl builtUrl,
+      SponsoredInteraction interaction, ObjectNode payload) {
+    // Do not add additional parameters for legacy suggest URLs
+    String baseUrl = builtUrl.getBaseUrl();
+    if (baseUrl.contains("ampxdirect.com") || baseUrl.startsWith("https://imp.mt48.net/static")) {
+      return;
+    }
```
cbguder, line 314: This is the downside of doing this in the Contextual Services pipeline. The Glean events do not indicate how the suggest data was sourced (from SFTP or the new AMP API), so we resort to looking at the reporting URL.

AMP uses `{mozillacla,firefoxmobilecla}.ampxdirect.com` for click URLs and `https://imp.mt48.net/static` for impression URLs for suggestions sourced from the SFTP server (i.e. US suggestions currently).

They use `bridge.*.admarketplace.net` for click URLs and and `https://imp.mt48.net/imp` for impression URLs for suggestions sourced from the Suggest API.

ingestion-beam/src/test/resources/contextualServices/urlAllowlist.csv
```diff
@@ -3,3 +3,6 @@ impression.com,impression
 click2.com,click
 test.com,click
 test.com,impression
+imp.mt48.net,impression
+ampxdirect.com,click
+admarketplace.net,click
```
cbguder, line 8: Adding these real domains as we now branch on the reporting URL

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
cbguder: Received confirmation from AMP that this is OK.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -270,7 +270,11 @@ public Result<PCollection<SponsoredInteraction>, PubsubMessage> expand(
           // https://mozilla-hub.atlassian.net/browse/DENG-392
           if (SponsoredInteraction.FORM_DESKTOP.equals(interaction.getFormFactor())
               && SponsoredInteraction.SOURCE_SUGGEST.equals(interaction.getSource())) {
-            addAdditionalDimensionsForSuggest(builtUrl, interaction);
+            addCustomDataForDesktopSuggest(builtUrl, interaction);
+          }
+
+          if (SponsoredInteraction.SOURCE_SUGGEST.equals(interaction.getSource())) {
```
efixler, line 276: I'm assuming that there's something about `addAdditionalDimensionsForInternationalSuggest` that expects side effects from either `addAdditionalDimensionsForSuggest` or  `addCustomDataForDesktopSuggest`, otherwise you could combine this conditional with the one above, with a nested condition for desktop.

efixler (approved):
The code looks straightforward enough. I'm less familiar with the syntax in the unit tests, but hey, you wrote tests!

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
```diff
@@ -301,6 +305,24 @@ private static void addAdditionalDimensionsForSuggest(BuildReportingUrl builtUrl
...
+    // Do not add additional parameters for legacy suggest URLs
+    String baseUrl = builtUrl.getBaseUrl();
+    if (baseUrl.contains("ampxdirect.com") || baseUrl.startsWith("https://imp.mt48.net/static")) {
+      return;
+    }
+
+    if (!payload.hasNonNull(Attribute.NORMALIZED_COUNTRY_CODE)) {
+      throw new RejectedMessageException(
+          "Missing required payload value " + Attribute.NORMALIZED_COUNTRY_CODE, "country");
+    }
```
dmueller, line 319: Do we know if this case happens today / is valid? I wonder if these ones should continue on as if this branch didn't exist rather than getting rejected -- at this point the ad was seen & interacted with, right?

dmueller: Separate thought that I'll dump in the same thread:
NORMALIZED_COUNTRY_CODE -- is this determined on the server?

I ask because I'm wondering if the country that the client determines it is in order to pick which suggestions to use could differ from what the server determined the country was; would there be any issue if they don't match? The result of this would be some amount of suggestions for country X showing up in country Y/Z/etc.

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
dmueller: > AMP uses {mozillacla,firefoxmobilecla}.ampxdirect.com for click URLs and https://imp.mt48.net/static for impression URLs for suggestions sourced from the SFTP server (i.e. US suggestions currently).
>
> They use bridge.*.admarketplace.net for click URLs and and https://imp.mt48.net/imp for impression URLs for suggestions sourced from the Suggest API.

might be worth capturing some of this in the comment here to de-fuzz what "legacy suggest url" means. I could go either way, but I'm wondering if someone is looking at this code in some months if it will be confusing and they'd need to track down this PR

dmueller (approved):

ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java
cbguder: I'm pretty sure this is determined by Glean. I copy/pasted the check for `NORMALIZED_COUNTRY_CODE` from `addAdditionalDimensionsForTopSites` below. I haven't checked the logs, but one would expect this to happen occasionally (we see it in MARS as well). Good point about letting it through though (perhaps without the `country-code`) because the interaction did indeed happen. I'll make that change.

It is also true that there will be cases where the country doesn't match the country we fetched suggestions for. I called this out in the PR description; I think it makes sense to send the actual country in these cases, but perhaps this is worth a conversation with AMP.

cbguder: Posed this question to AMP on Slack. Assuming they'll prefer to get the actual country at click/impression time, but if they prefer otherwise we can pivot to https://github.com/mozilla-services/mars/pull/573.

PR #2782, opened 2025-05-12
[DENG-8480] Use CODEOWNERS for tagging reviewers for dependabot

scholtzan:
https://mozilla-hub.atlassian.net/browse/DENG-8480

Dependabot reviewers configuration is being removed, and instead they are recommending to rely on CODEOWNERS files to assign reviewers

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,8 +17,6 @@ updates:
     versions: [ ">=2.2.0" ]
   schedule:
     interval: daily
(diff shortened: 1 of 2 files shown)


BenWu (approved):
It said the codeowners was invalid because @mozilla/data-platform-infra-wg didn't have write access so I gave the team the writer role.  The team is a subset of the telemetry team so it shouldn't matter

PR #2857, opened 2025-11-06
DENG-9533 - Support ingesting Glean server pings submitted directly to Pub/Sub

akkomar:
## Description

Adds a third ingestion path to the Decoder job: messages published directly to Pub/Sub by server-side telemetry clients, bypassing Cloud Logging / fluentbit. Enabled at deploy time via a new `--directPubsubEnabled=true` flag (parallel to `--logIngestionEnabled`).

I intend to land this first, then prepare IAC config, land https://github.com/mozilla/glean_parser/pull/821, and then test this in stage with MARS.

I have tested this end to end using template generated from https://github.com/mozilla/glean_parser/pull/821 in a sandbox project: https://github.com/akkomar/glean-server-examples/tree/pubsub/test-go-publisher.

- [DENG-9533](https://mozilla-hub.atlassian.net/browse/DENG-9533)

deployed manually.

See the full [code deployment policy](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27922303/Ingestion+Beam#Prod-Code-Deployment-Policy)
for details.

[DENG-9533]: https://mozilla-hub.atlassian.net/browse/DENG-9533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ExtractIpFromDirectPubsub.java
```diff
@@ -0,0 +1,87 @@
...
+          countIpExtracted.inc();
+
+          byte[] sanitizedPayload = payloadJson.toString().getBytes(StandardCharsets.UTF_8);
+          return new PubsubMessage(sanitizedPayload, attributes);
+        }
+        return message;
+
+      } catch (IOException e) {
+        throw new ParseDirectPubsub.InvalidDirectPubsubException(
+            "Message has no submission_timestamp but could not be parsed as Direct Pub/Sub JSON");
```
BenWu, line 75: This doesn't look like the correct log message for this

ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
```diff
@@ -84,6 +86,17 @@ public static PipelineResult run(DecoderOptions.Parsed options) {
             .apply(ParseLogEntry.of())//
             .failuresTo(failureCollections) : p)
 
+        // Special case: parsing structured telemetry pings published directly to Pub/Sub.
+        .map(p -> options.getDirectPubsubEnabled() ? p //
+            // We first extract IP address from the payload and remove it for consistency
+            // with the standard flow.
+            .apply(ExtractIpFromDirectPubsub.of()) //
```
BenWu, line 93: Why is ip extraction done here instead of earlier and then using the same geo lookup transforms as normal messages?

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseDirectPubsub.java
```diff
@@ -0,0 +1,127 @@
...
+
+        ObjectNode messagePayloadJson;
+        try {
+          messagePayloadJson = Json.readObjectNode(message.getPayload());
+        } catch (IOException e) {
+          throw new InvalidDirectPubsubException("Message payload could not be parsed as JSON");
+        }
+
+        // Extract required fields from ping envelope
+        String documentId = messagePayloadJson.path("document_id").textValue();
```
BenWu, line 91: Could the publisher put these fields be in the message attributes instead so that they don't need to be extracted?  I'm not sure that's better in any way, just curious if there's a reason for doing it this way.  Is it easier to package it like this for glean?

BenWu (dismissed):

ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
akkomar: I'm keeping this consistent with logentry path. IIRC this was written this way to keep that path's operations together, but now when I look at this it would maybe be clearer to do what you suggest. I'm not sure though if reorganizing this wouldn't change some exception propagation behavior. I'll revisit this later or in a separate PR.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseDirectPubsub.java
akkomar: Good point. The current shape mirrors the LogEntry envelope, but by using attributes we could skip parsing JSON to get them. I'll investigate.

ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
akkomar: Took a stab at simplifying this in https://github.com/mozilla/gcp-ingestion/pull/2936. It's more readable IMO, but I'd rather test this in stage first.

ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/ParseDirectPubsub.java
akkomar: I moved ping metadata to message attributes and it turned out to be a better approach - results in less code in Decoder.

ingestion-beam/src/main/java/com/mozilla/telemetry/Decoder.java
akkomar: This part got simpler after moving ping metadata to message attributes so I'll resolve this thread.

BenWu (approved):

PR #2920, opened 2026-04-14
fix: split build-and-push into per-project workflows with path filters

mikaeld:
## Summary

Split the single `build-and-push.yaml` workflow into two separate workflows with path filters so that only the relevant image is rebuilt when code changes:

- **`build-and-push-edge.yaml`**: triggers only on `ingestion-edge/**` changes
- **`build-and-push-sink.yaml`**: triggers on `ingestion-sink/**`, `ingestion-core/**`, `pom.xml`, and `checkstyle/**` changes

Previously, any commit to main triggered image builds for both projects, causing unnecessary rebuilds and rolling restarts in both ingestion-edge and ingestion-sink.

## Path filters

| Workflow | Triggers on |
|---|---|
| edge | `ingestion-edge/**`, workflow file |
| sink | `ingestion-sink/**`, `ingestion-core/**`, `pom.xml`, `checkstyle/**`, workflow file |

The sink paths include `ingestion-core/` and root `pom.xml` because the sink Dockerfile uses the repo root as build context and compiles against `ingestion-core`.

Fixes [SVCSE-4419](https://mozilla-hub.atlassian.net/browse/SVCSE-4419)

[SVCSE-4419]: https://mozilla-hub.atlassian.net/browse/SVCSE-4419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-and-push-sink.yaml
+++ b/.github/workflows/build-and-push-sink.yaml
@@ -5,6 +5,7 @@ on:
     branches:
       - main
     paths:
+      - "checkstyle/**"
       - "ingestion-sink/**"
       - "ingestion-core/**"
       - "pom.xml"
--- a/.github/workflows/build-and-push-sink.yaml
+++ b/.github/workflows/build-and-push-sink.yaml
@@ -1,0 +1,25 @@
+name: Build and Push ingestion-sink
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "ingestion-sink/**"
+      - "ingestion-core/**"
+      - "pom.xml"
+      - ".github/workflows/build-and-push-sink.yaml"
+
+jobs:
+  ingestion-sink:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-push-sink.yaml
```diff
@@ -0,0 +1,25 @@
+name: Build and Push ingestion-sink
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - "ingestion-sink/**"
```
akkomar, line 9:
```suggestion
      - "checkstyle/**"
      - "ingestion-sink/**"
```

akkomar (approved):
r+wc

PR #2961, opened 2026-07-27
feat(ingestion-edge): allow flush-manager to run under restricted PSS and inherit node scheduling

mikaeld:
## Description

Adds opt-in flags to `flush_manager.py` so drain Jobs work correctly on GKE shared clusters that enforce PodSecurity `restricted:latest` and use taints to isolate tenant workloads:

- `--enable-restricted-security-context`: adds a pod- and container-level `securityContext` that satisfies PSS `restricted`.
- `--node-selector` (JSON dict): sets `nodeSelector` on the drain Job's PodSpec.
- `--tolerations` (JSON list of snake_case toleration dicts): sets `tolerations` on the drain Job's PodSpec.

### securityContext applied when `--enable-restricted-security-context` is set

```
container:
  allowPrivilegeEscalation: false
  capabilities:
    drop: [ALL]
  seccompProfile:
    type: RuntimeDefault
pod:
  runAsNonRoot: true
  runAsUser: 10001
  runAsGroup: 10001
  fsGroup: 10001
  seccompProfile:
    type: RuntimeDefault
```

## Motivation

On GKE shared clusters that enforce PSS `restricted:latest` (e.g. `dataservices-high` used by the SVCSE-4142 ingestion-edge v2 deployment), the current Job template is rejected by the admission controller with:

```
Error creating: pods "flush-pvc-..." is forbidden: violates PodSecurity "restricted:latest": allowPrivilegeEscalation != false, unrestricted capabilities, runAsNonRoot != true, seccompProfile
```

Drain never runs, causing Released PVs / orphan PVCs to accumulate. Observed during Phase 3 of the SVCSE-4142 cutover with ~300 orphan PVCs.

Once PSS admission is satisfied, Jobs still need to schedule onto the same tainted node pool as the StatefulSet they drain. On `dataservices-high`, the `ingestion-edge-1` node pool carries `tenant=ingestion-edge:NoExecute`. Without matching `tolerations` (and a matching `nodeSelector` so the scheduler doesn't just pick any node), the Jobs would fall through to the untainted default pool: it would appear to work (StorageClass is `WaitForFirstConsumer` zonal `pd-balanced`, and the default pool spans the same zones), but silently cross the tenant isolation boundary. If the default pool ever loses a zone or gains its own taint, Jobs go Pending and the SVCSE-4142 symptom returns with no alarm.

## Behavior on v1 (defaults)

All three flags default to off / empty:
- `--enable-restricted-security-context` defaults to `False`
- `--node-selector` defaults to `{}` -> PodSpec.nodeSelector is `None`
- `--tolerations` defaults to `[]` -> PodSpec.tolerations is `None`

v1's cloudops-infra cluster doesn't enforce PSS, runs containers as root, and has no scheduling constraints on the ingestion-edge node pool. Not opting into any of the flags keeps existing v1 behavior identical.

## Behavior on v2 (opt in)

The v2 chart's flush-manager container will pass all three flags in its args (follow-up PR in `dataservices-infra`), sourcing `--node-selector` and `--tolerations` from the same YAML anchors the StatefulSet uses so they stay in sync by construction:

```yaml
--enable-restricted-security-context
--node-selector={"tenant": "ingestion-edge"}
--tolerations=[{"key": "tenant", "operator": "Equal", "value": "ingestion-edge", "effect": "NoExecute"}]
```

Drain Jobs then start under `restricted:latest`, schedule onto `ingestion-edge-1`, and drain telemetry data on the same isolated node pool as the workload that queued it.

- SVCSE-4142
- SVCSE-3814 (related: alerting for this exact class of failure)
- SVCSE-4440 (related: namespace-scoped PVC-scan loop; would provide reconciliation for orphans that predate this fix)

### DIFF
--- a/ingestion-edge/ingestion_edge/flush_manager.py
+++ b/ingestion-edge/ingestion_edge/flush_manager.py
@@ -6,7 +6,7 @@ from datetime import datetime, timedelta
 from functools import partial
 from multiprocessing.pool import ThreadPool
 from time import sleep
-from typing import Callable, Dict, List
+from typing import Callable, Dict, List, Optional
 import json
 import os
 
@@ -129,7 +129,7 @@ parser.add_argument(
     default=[],
     type=json.loads,
     help="Tolerations for flush job pods as a JSON list of toleration objects with "
-    "snake_case keys (e.g. '[{\"key\": \"tenant\", \"operator\": \"Equal\", "
+    'snake_case keys (e.g. \'[{"key": "tenant", "operator": "Equal", '
     '"value": "ingestion-edge", "effect": "NoExecute"}]\'). Defaults to none. Set '
     "to match the workload's tolerations so drain Jobs are not evicted from tainted "
     "node pools.",
@@ -212,8 +212,8 @@ def _create_flush_job(
     namespace: str,
     service_account_name: str,
     restricted_security_context: bool = False,
-    node_selector: Dict[str, str] = None,
-    tolerations: List[dict] = None,
+    node_selector: Optional[Dict[str, str]] = None,
+    tolerations: Optional[List[dict]] = None,
 ) -> V1Job:
     logger.info(f"creating job: {name}")
     container_security_context = None
@@ -308,8 +308,8 @@ def flush_released_pvs(
     namespace: str,
     service_account_name: str,
     restricted_security_context: bool = False,
-    node_selector: Dict[str, str] = None,
-    tolerations: List[dict] = None,
+    node_selector: Optional[Dict[str, str]] = None,
+    tolerations: Optional[List[dict]] = None,
 ):
     """
     Flush persistent volumes.
@@ -421,8 +421,8 @@ def flush_released_pvs_and_delete_complete_jobs(
     namespace: str,
     service_account_name: str,
     restricted_security_context: bool = False,
-    node_selector: Dict[str, str] = None,
-    tolerations: List[dict] = None,
+    node_selector: Optional[Dict[str, str]] = None,
+    tolerations: Optional[List[dict]] = None,
 ):
     """Flush released persistent volumes then delete complete jobs.
 
--- a/ingestion-edge/ingestion_edge/flush_manager.py
+++ b/ingestion-edge/ingestion_edge/flush_manager.py
@@ -49,6 +49,12 @@ ALREADY_EXISTS = "AlreadyExists"
 CONFLICT = "Conflict"
 NOT_FOUND = "Not Found"
 
+# V1Toleration's __init__ accepts snake_case names (e.g. toleration_seconds),
+# but Kubernetes YAML/JSON conventionally uses camelCase (tolerationSeconds).
+# Map the wire-format names to Python attribute names so --tolerations JSON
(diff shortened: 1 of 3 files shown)


whd:
Per slack discussion this likely needs additions for taints and tolerations. Generally I'm expecting we will need to test/verify this works in v2 stage and this might require a couple PRs to get right, but neither v1 nor v2 prod auto deploys so such PR-based iteration is safe.

ingestion-edge/ingestion_edge/flush_manager.py
```diff
@@ -180,8 +211,27 @@ def _create_flush_job(
...
+            seccomp_profile=V1SeccompProfile(type="RuntimeDefault"),
+        )
+        pod_security_context = V1PodSecurityContext(
+            run_as_non_root=True,
+            run_as_user=10001,
+            run_as_group=10001,
+            fs_group=10001,
+            seccomp_profile=V1SeccompProfile(type="RuntimeDefault"),
+        )
+    toleration_objs = [V1Toleration(**t) for t in tolerations or ()] or None
```
whd: 🤖  saith
```
  New defect: V1Toleration(**t) accepts only snake_case, and fails in a hot loop

  flush_manager.py:234:

  toleration_objs = [V1Toleration(**t) for t in tolerations or ()] or None

  V1Toleration has exactly one field whose Python name differs from its Kubernetes name:

  attribute_map: {'effect': 'effect', 'key': 'key', 'operator': 'operator',
                  'toleration_seconds': 'tolerationSeconds', 'value': 'value'}

  Verified against the installed client:

  current anchor -> NoExecute OK
  camelCase tolerationSeconds -> TypeError: V1Toleration.__init__() got an
    unexpected keyword argument 'tolerationSeconds'. Did you mean 'toleration_seconds'?
```

ingestion-edge/ingestion_edge/flush_manager.py
```diff
@@ -180,8 +211,27 @@ def _create_flush_job(
...
+    pod_security_context = None
+    if restricted_security_context:
+        container_security_context = V1SecurityContext(
+            allow_privilege_escalation=False,
+            capabilities=V1Capabilities(drop=["ALL"]),
+            seccomp_profile=V1SeccompProfile(type="RuntimeDefault"),
+        )
+        pod_security_context = V1PodSecurityContext(
+            run_as_non_root=True,
+            run_as_user=10001,
```
whd, line 239: I have a mild preference for either making this a full input object or otherwise documenting that this magic value comes from https://github.com/mozilla/helm-charts/

ingestion-edge/tests/unit/flush_manager/test_create_job.py
```diff
@@ -195,6 +195,155 @@ def create_job(namespace, body):
     batch_api.list_namespaced_job.called_once_with("namespace")
 
 
+def test_flush_released_pvs_with_restricted_security_context(
```
whd, line 198: This seems like it could be significantly DRYer than it is

whd (approved):
r+ but per previous discussion this might take some iteration in stage to actually verify.

PR #2976, opened 2026-09-13
[SPONS-323] Skip contextualservices reporting for Moz-supplied, non-AMP interactions 

mashalifshin:
## Description

Mozilla Ads is expanding beyond AMP for supplying sponsored suggestions and will soon have Mozilla-supplied sponsored suggestions in addition to AMP ones, coming from new ad providers or from our own direct sold sponsored suggestions. Pings for interactions with these non-AMP suggestions will soon be flowing through the contextualservices reporter job, however we do not want to report those to AMP, or call any reporting URL for those at all.

The goal of the PR is to update the existing contextualservices job to skip over pings that are not AMP-specific, while retaining the existing AMP workflow entirely unchanged so it can continue to call the AMP-provided reporting_url for interactions with AMP's sponsored suggestions. 

Non-AMP pings are identified by mozilla-host reporting urls. The urls aren't called, but they are important to send in the ping with the `suggestion_id` parameter, to record in GBQ for our own telemetry.  

For non-AMP sponsored suggestions, we don't need a reporting url as we will be able to directly use the pings within our own GBQ data warehouse, so here we just skip over this for now. The eventual plan is to move suggest interactions to MARS callbacks as we have done for Tiles and Native ads, so that the MARS amp-report job would centralize the functionality and this job can be sunset.

## Related Tickets & Documents

- [SPONS-323](https://mozilla-hub.atlassian.net/browse/SPONS-323)
- [Product Brief and Eng Design Doc](https://docs.google.com/document/d/1mHO71QHHxaFVJkGcL4baIeEZ2mT0ztbHjwmXvbeYNMM/edit?tab=t.0)

### Changes affecting ingestion-edge

Code updates are always safe to merge since production code updates are always
deployed manually.

### Changes affecting ingestion-sink

Code updates are always safe to merge since production code updates are always
deployed manually. See [these instructions](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27921000/Ingestion+Sink#IngestionSink-Toupdatethecodeversion)
for updating the code version.

### Changes affecting ingestion-beam (including ingestion-core)

- Only merge changes to `main` that you want to propagate to production automatically
- Check [pipeline latency](https://yardstick.mozilla.org/d/bZHv1mUMk/pipeline-latency?orgId=1&from=now-6h&to=now) before merging, particularly if:
  - The merge will occur within 2 hours of the UTC date change, or
  - You are merging multiple PRs in quick suggestion

[...]

PR #2978, opened 2026-09-15
[SPONS-331] Skip contextualservices reporting for suggest-specific placeholder urls only

mashalifshin:
## Description

Follow-up to #2976.

Neither change alters what gets reported. `ads.mozilla.org` is not in the URL allow list,
so both `/v1/st` and `/v1/t` were already being dropped before any request was made — this
corrects *which* mechanism drops them, and what the counters say about it.

 1. Narrow the placeholder match from host-only to host and path. Only the MARS suggest
  placeholder at `/v1/st` is filtered out. The host-only match was also catching the legacy
  topsites callback at `/v1/t`, which is a real reporting URL and now goes back to the normal
  path through the URL allow list.
 2. Move the filter ahead of `EmitCounters`. `valid_impression_url`, `valid_click_url`,
  `valid_submission`, and `valid_submission_merino` mean "will be reported". Inbound volume is still recoverable by adding `quick_suggest_moz_ads_filtered_urls`  back in.
  3. Renames for consistency, no behavior change: `FilterMozAdsInteractions` →
  `FilterMozAdsSuggestPlaceholders`, `isMozAdsReportingUrl` →
  `isMozAdsSuggestPlaceholderToFilter`, and the two constants to match. The counter name
  `quick_suggest_moz_ads_filtered_urls` is unchanged, though its Beam namespace follows the class rename, but that's probably okay since this is a 1day old counter that no one is consuming yet.
  
## Expected impact

- `quick_suggest_moz_ads_filtered_urls` drops to ~0 — it was only ever counting `/v1/t`, and
  there is currently no `/v1/st` traffic in any environment.
- `firefox_desktop/top_sites_v1/rejected_nonnull_url` returns at roughly its pre-#2976 rate
  (~102 per 11 days), which is `/v1/t` hitting the allow list again and the signal that the
  narrowing worked.
- The reorder is numerically inert today, since there is no `/v1/st` traffic to exclude from
  the `valid_*` counters.
  
 
[SPONS-331](https://mozilla-hub.atlassian.net/browse/SPONS-331)
[Product Brief and Eng Design Doc](https://docs.google.com/document/d/1mHO71QHHxaFVJkGcL4baIeEZ2mT0ztbHjwmXvbeYNMM/edit?tab=t.0)

deployed manually.

See the full [code deployment policy](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27922303/Ingestion+Beam#Prod-Code-Deployment-Policy) for details.

[SPONS-331]: https://mozilla-hub.atlassian.net/browse/SPONS-331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gleonard-m:
If the urls are being filtered by existing URL checks [ParseReportingUrl.java#L423-L441 ](https://github.com/mozilla/gcp-ingestion/blob/0e400deb06c8362b860aa718e5471ea0b9c3249f/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java#L423-L441 ) is this additional check needed?

mashalifshin:
> If the urls are being filtered by existing URL checks [ParseReportingUrl.java#L423-L441 ](https://github.com/mozilla/gcp-ingestion/blob/0e400deb06c8362b860aa718e5471ea0b9c3249f/ingestion-beam/src/main/java/com/mozilla/telemetry/contextualservices/ParseReportingUrl.java#L423-L441) is this additional check needed?

Good question! I think right now we have redundant checks in prod, and this PR is what makes it not redundant, and makes the new counter actually count only suggest pings. Lmk if this makes sense:

1. Before I made any changes, pings with `reporting_url`s like https://ads.mozilla.org/v1/t...` were getting filtered by `isUrlValid` in `ParseReportingUrl`, because the host is not in the allowlist csv. `rejected_nonnull_url` gets incremented, error row is written. 
2. First change, in prod now, added a short circuit before `isUrlValid` to filter out just non-AMP suggest interactions by host, and `FilterMozAdsInteractions` transform downstream that drops those elements and counts them. This is redundant right now, we're filtering the same exact same interactions, just earlier in the flow and via a different mechanism. And we're filtering all interactions with moz hosts, not just suggest ones. So currently, interactions for tiles stopped producing error rows as above, and started incrementing `quick_suggest_moz_ads_filtered_urls` instead. 
3. This PR. Adds host and path filtering to the ``isMozAdsReportingUrl` in `ParseReportingUrl`. That's one distinction, this has the capability to filter by path. So now I think this is functioning as I originally intended before I knew we had pings with moz hosts still flowing through ctxsvc reporter. Now suggest interactions are short circuited (not considered invalid wrt allow list, but just bypassed earlier in the flow),  `quick_suggest_moz_ads_filtered_urls` is incremented only for our special, suggest specific urls that have path `v1/st`, and not for everything with a moz host. And Tiles interactions go back to  behavior of step # 1 before I touched the code. 

All that said, there could totally be a better way to structure bypassing/no-op-ing/short-circuiting these suggest-specific, non-AMP interactions. definitely lmk if you have ideas there. But I didn't want to conflate this filtering with the allow list filtering, even though in practice for all of the states listed above, nothing gets sent to amp except amp reports.

luc-lisi (approved):
This looks good and thank you @mashalifshin !

+1 to there probably being good discussion to be had around a long-term solution for this job, but I think that's better saved for when deadlines are over. So agree'd this seems the safest way to handle it for now! 🎉

mashalifshin:
Having written out the answer to @gleonard-m 's excellent question, I did think of a way to make this a bit clearer, the naming is now misleading, give me a moment to push up another version

mashalifshin:
Okay, pushed a rename update that I think clarifies things a bit and reflects what I learned since the previous PR. My big explanatory comment is now stale, but the PR description describes the current changes accurately.

@gleonard-m I do see your point also that we're doing the same filtering twice:

```
// ParseReportingUrl.expand()
if (isMozAdsSuggestPlaceholderToFilter(reportingUrl)) {
  PerDocTypeCounter.inc(attributes, "valid_url");
  return interaction.toBuilder().setReportingUrl(reportingUrl).build();
}
 
 ...
 
// FilterMozAdsSuggestPlaceholders.processElement()
if (ParseReportingUrl.isMozAdsSuggestPlaceholderToFilter(interaction.getReportingUrl())) {
  filteredUrlsCounter.inc();
  return;
}
```

So we make the same check twice, first to attach the reporting url, then later downstream we process the same element with the same check, and they aren't structurally connected, they just use the same underlying helper. 

Claude's got a good idea for what to do here, but it's a larger refactor that I don't want to add to this fix. @gleonard-m if you think this makes sense, we could try in a follow-up PR?

> **The cleaner shape** is a single check: have `ParseReportingUrl` emit zero elements for a placeholder and own the counter, then delete `FilterMozAdsSuggestPlaceholders` entirely. One evaluation, one place, and the guarantee gets *stronger* — the element never enters the pipeline at all, rather than being removed after.
> 
> What blocks it: `expand()` is built on `MapElements`, which is one-in-one-out. Dropping requires `FlatMapElements`, which touches every path through that method, not just the placeholder one.

mashalifshin:
> This looks good and thank you @mashalifshin !
> 
> +1 to there probably being good discussion to be had around a long-term solution for this job, but I think that's better saved for when deadlines are over. So agree'd this seems the safest way to handle it for now! 🎉

@luc-lisi please have one more look, I think this at least clearer now in terms of naming. And in my last comment I have another possible refactor to clarify this.

I think we should plan to get suggest onto MARS callbacks, and then this job can be the reporter only for older clients that don't use callbacks ...

PR #2979, opened 2026-09-17
[Bug 2073281] Scrub crash ping Java exception messages

scholtzan:
## Description

See https://mozilla.slack.com/archives/CKFKC7Y1Y/p1789674235829369

Uncaught Java exception messages contain arbitrary app state and must not be persisted. Two fields in crash pings from the fenix, focus_android, and klar_android namespaces carry them:

- `metrics.object["crash.java_exception"]`: removes `$.messages` (structure v0) and `$.throwables[].message` (v1).
- `metrics.object["meta.annotations"]`: the raw annotation blob, shipped since Firefox 155 as `{"source": "<serialized JSON>"}` where JavaException is itself a serialized JSON string. Both levels are parsed, scrubbed, re-serialized.

`crash.java_exception` is only populated when there was an uncaught Java exception, so dropping on its presence would discard every Android Java-crash report. Scrubbing gets the same privacy outcome and keeps the stacks.

This scrubber rewrites a payload field. Existing redactions only delete keys, so this is new.

Error output isn't covered. ParsePayload's failure path builds FailureMessage from the original bytes, so a ping scrubbed here that later fails validation reaches error tables unscrubbed. This may be ok for this?

- Bug 2049744

deployed manually.

See the full [code deployment policy](https://mozilla-hub.atlassian.net/wiki/spaces/SRE/pages/27922303/Ingestion+Beam#Prod-Code-Deployment-Policy) for details.


ingestion-beam/src/main/java/com/mozilla/telemetry/decoder/MessageScrubber.java
```diff
@@ -268,6 +270,21 @@ public class MessageScrubber {
...
+      // fenix
+      "org-mozilla-firefox", "org-mozilla-firefox-beta", "org-mozilla-fenix",
+      "org-mozilla-fenix-nightly", "org-mozilla-fennec-aurora",
+      // focus_android
+      "org-mozilla-focus", "org-mozilla-focus-beta", "org-mozilla-focus-nightly",
+      // klar_android
+      "org-mozilla-klar");
+
+  // Bug tracking removal of Java exception messages from the crash ping.
+  private static final String JAVA_EXCEPTION_BUG = "2049744";
```
scholtzan: This bug is related to the initial change the introduced this issue. Maybe it's better to add a separate one to track the scrubbing...

BenWu (approved):
LGTM but I'm not sure about how we want to deal with the potential error output

scholtzan:
I'll merge this for now. It'd be good to get an idea how many land in errors and handle that as a follow-up.

